PR #14, opened 2025-11-12
style: remove invalid example label

ThomasDsantos:
Go's time package [ParseDuration](https://pkg.go.dev/time#ParseDuration) function doesn't support days.

Note that it would be useful to support days when setting custom silence times in the future.


SoulKyu:
Totally agree, the under func should allow custom silence with days.
In my todo-list.

I merge it so example is more accurate with actual behavior.

PR #29, opened 2026-05-07
feat: allow for objects in env

cwrau:
That way the user can use stuff like `valueFrom.secretKeyRef` and such


SoulKyu:
Thanks for your contribution !

Legit, merging

PR #37, opened 2026-07-24
fix(dashboard): SSE-pushed alerts bypass every hidden-alert check in the client filter mirror

### LINKED ISSUE #34: fix(dashboard): SSE-pushed alerts bypass every hidden-alert check in the client filter mirror
## Problem

SSE pushes **unfiltered** alerts to every browser (`internal/webui/services/alert_cache.go:241-249` broadcasts `NewAlerts`/`UpdatedAlerts` straight from the Alertmanager diff, with no per-user filtering). The client is therefore the only gate: `applyIncrementalUpdate()` calls `alertMatchesFilters()` before merging (`internal/webui/templates/scripts/dashboard_data.templ:316`, `:343`).

That client-side mirror of `applyDashboardFilters()` (`internal/webui/handlers/dashboard_handlers.go:372-455`) is broken for **every hidden-alert mechanism**:

1. **Filter-preset hidden alerts — comparison against the wrong type.**
   `dashboard_data.templ:498-502` does `this.filterHiddenAlerts.includes(alert.fingerprint)`, but `filterHiddenAlerts` is an array of **objects**, not strings — it is populated from the API as `data.hidden_alerts` (`dashboard_filter_presets.templ:196`, `:222`) and pushed as `{fingerprint, alert_name, instance, reason}` (`dashboard_filter_presets.templ:314-319`, matching `internal/webui/models/filter_preset.go:54-59`). Every other call site correctly uses `.some(a => a.fingerprint === fp)` (`dashboard_filter_presets.templ:308`, `dashboard_actions.templ:166`). `Array.includes(string)` over objects is **always `false`** → the check never fires.

2. **Filter-preset hidden rules — non-existent field.**
   `alertMatchesHiddenRule()` (`dashboard_data.templ:525-538`) bails on `if (!rule.matchers) return false;`. Preset rules have no `matchers` field: they are `{name, label_key, label_value, is_regex, is_enabled}` (`internal/webui/models/filter_preset.go:62-70`), which is exactly what the server matches on (`internal/webui/services/hidden_alerts_service.go:191-215`). The function returns `false` for every rule → the check never fires.

3. **Globally hidden alerts — not checked at all.**
   The server drops them via `hiddenAlertsService.IsAlertHidden(sessionID, alert)` (`dashboard_handlers.go:384`). `alertMatchesFilters()` has no equivalent. The list is available client-side (`window.currentSettingsModal.hiddenAlerts`) and the notification path already uses it (`notification_service.templ:343-347`); the dashboard merge path does not.

Symmetric consequence: in `displayMode === 'hidden'` the server returns **only** hidden alerts (`dashboard_handlers.go:396-400`), but the client adds any SSE alert to that view since it never evaluates hidden-ness.

### Failure path

1. User hides a noisy alert (globally, or via a saved filter preset) — the server-rendered dashboard correctly stops showing it.
2. The alert flaps: it disappears from Alertmanager, the cache deletes it (`alert_cache.go:223`), then it re-fires and is re-added as new (`alert_cache.go:175-178`) and broadcast over SSE.
3. `alertMatchesFilters()` returns `true` (all three hidden checks are inert) → the alert is pushed into `this.alerts` (`dashboard_data.templ:343-349`) and rendered.
4. It stays visible until a full page reload re-applies the server-side filter.

## Impact

Hiding an alert is the main tool users have against known-noisy alerts, and flapping alerts are exactly what hits this path. On a live dashboard (SSE is the default transport) hidden alerts silently come back and stay back — the feature appears to work at load time and then quietly stops. Two of the three checks were clearly written with the intent to filter (added in `a94e616`), so this is an unnoticed data-shape regression, not a design choice.

## Proposed approach

Fix the mirror in `internal/webui/templates/scripts/dashboard_data.templ` (edit the `.templ`, regenerate with `make webui-templates`, never touch `*_templ.go`):

- `alertMatchesFilters()`: use `this.filterHiddenAlerts.some(h => h.fingerprint === alert.fingerprint)`.
- `alertMatchesHiddenRule(alert, rule)`: rewrite against the real shape — skip when `!rule.is_enabled`, read `alert.labels?.[rule.label_key]` (skip when absent), then `rule.is_regex ? new RegExp(rule.label_value).test(v) : (rule.label_value === '' || rule.label_value === v)`, mirroring `hidden_alerts_service.go:191-215`. Wrap the `RegExp` construction in `try/catch` so one bad user rule cannot break the whole merge.
- Add the global hidden-alert check (reuse `isAlertHidden()` from `dashboard_actions.templ:792-807`, or the `window.currentSettingsModal?.hiddenAlerts` lookup already used in `notification_service.templ:343-347`).
- Handle `displayMode === 'hidden'` as the inverse case: keep only alerts that are hidden.

## Acceptance criteria

- [ ] With a filter preset hiding fingerprint `F`, an SSE `newAlerts` payload containing `F` does not add it to the table.
- [ ] With a preset rule `label_key=team, label_value=noisy, is_enabled=true`, an SSE alert carrying `team=noisy` is not added; a non-regex rule with an empty `label_value` hides every alert carrying that label key (same semantics as the server).
- [ ] An alert hidden globally from the settings modal is not added to the table when it re-fires over SSE.
- [ ] In `displayMode === 'hidden'`, a newly firing non-hidden alert is **not** added to the view.
- [ ] After an SSE merge, a manual full reload of the dashboard shows the same alert set (client and server filters agree).
- [ ] An invalid regex in a hidden rule does not throw inside `applyIncrementalUpdate()`.

SoulKyu:
## Problem

SSE broadcasts unfiltered alerts to every browser, so `alertMatchesFilters()` in `dashboard_data.templ` is the only gate before merging them into the table. All three hidden-alert mechanisms in that mirror were inert:

1. **Filter-preset hidden alerts** — `this.filterHiddenAlerts.includes(alert.fingerprint)` compared a string against an array of `{fingerprint, alert_name, ...}` objects, so it was always `false`.
2. **Filter-preset hidden rules** — `alertMatchesHiddenRule()` bailed on a non-existent `rule.matchers` field; preset rules are `{label_key, label_value, is_regex, is_enabled}`.
3. **Globally hidden alerts** — not checked at all on the dashboard merge path (the notification path already checks `window.currentSettingsModal.hiddenAlerts`).

Net effect: a hidden alert that flaps comes back over SSE and stays visible until a full page reload.

## Changes

All in `internal/webui/templates/scripts/dashboard_data.templ` (generated `*_templ.go` regenerated via `make webui-templates`):

- `alertMatchesFilters()` now computes hidden-ness the same way as the server's `applyDashboardFilters`: global (`window.currentSettingsModal?.hiddenAlerts` by fingerprint) OR filter-preset (fingerprint list via `.some(h => h.fingerprint === ...)`, plus rules).
- `displayMode === 'hidden'` is handled as the inverse case: only hidden alerts are kept, matching the server, so newly firing non-hidden alerts no longer leak into the hidden view.
- `alertMatchesHiddenRule()` rewritten against the real rule shape, mirroring `HiddenAlertsService.IsAlertHiddenByFilter`: skips disabled rules, skips alerts missing the label key, empty non-regex `label_value` matches any alert carrying the key, and `new RegExp` is wrapped in try/catch so one invalid user regex cannot break the whole SSE merge.

## Validation

- `go build ./...` passes after `make webui-templates`.
- Node self-check of the extracted mixin against every acceptance criterion (preset fingerprint hide, exact/empty/regex/disabled rules, global hide, hidden display mode inverse, invalid regex no-throw, classic-mode resolved drop): all pass.

Closes #34


internal/webui/templates/scripts/dashboard_data.templ
```diff
@@ -494,20 +494,22 @@ templ DashboardData() {
...
-					if (this.filterHiddenAlerts.includes(alert.fingerprint)) {
+				// Check hidden-ness (global + filter-preset), mirroring the server's
+				// applyDashboardFilters: hidden mode shows only hidden alerts, every
+				// other mode drops them
+				const isGlobalHidden = (window.currentSettingsModal?.hiddenAlerts || [])
+					.some(hidden => hidden.fingerprint === alert.fingerprint);
+				const isFilterHidden =
+					(this.filterHiddenAlerts || []).some(hidden => hidden.fingerprint === alert.fingerprint) ||
+					(this.filterHiddenRules || []).some(rule => this.alertMatchesHiddenRule(alert, rule));
+				const isHidden = isGlobalHidden || isFilterHidden;
```
SoulKyu: **Non-blocking: globally hidden *rules* are still bypassed on the SSE path.**

`isGlobalHidden` only mirrors half of the server's global check. `HiddenAlertsService.IsAlertHidden` (`internal/webui/services/hidden_alerts_service.go:119-168`) matches both the fingerprint list **and** the user's global label rules (`userHiddenRules`), while this mirror only checks `window.currentSettingsModal?.hiddenAlerts` by fingerprint. So an alert hidden by a global rule (settings modal → hidden rules) that flaps and comes back over SSE reappears in the table until reload — the exact failure path this PR fixes for the other three mechanisms. Symmetrically, in `displayMode === 'hidden'` a rule-hidden alert updated over SSE is dropped from the hidden view because the client computes `isHidden = false` for it.

The data is already available client-side: `loadHiddenRules()` runs in the settings modal's `init()` (`dashboard_settings.templ:69`), so `window.currentSettingsModal?.hiddenRules` is populated at page load. One caveat if you extend the check: global rules serialize with different field names than preset rules — `labelKey`/`labelValue`/`isRegex`/`enabled` (see `internal/backend/models/user_hidden_alerts.go:25-36` and the existing usage `rule.enabled` at `dashboard_settings.templ:587`) — so `alertMatchesHiddenRule` can't be reused as-is; either normalize the shape or add a small camelCase-variant matcher:

```js
const isGlobalHidden =
	(window.currentSettingsModal?.hiddenAlerts || []).some(hidden => hidden.fingerprint === alert.fingerprint) ||
	(window.currentSettingsModal?.hiddenRules || []).some(rule => this.alertMatchesGlobalHiddenRule(alert, rule));
```

Since issue #34 only enumerated the fingerprint-based global check, this is fine as a fast-follow rather than a blocker for this PR.


internal/webui/templates/scripts/dashboard_data.templ
```diff
@@ -521,20 +523,24 @@ templ DashboardData() {
...
-						? new RegExp(matcher.value).test(labelValue)
-						: labelValue === matcher.value;
+				const labelValue = alert.labels?.[rule.label_key];
+				if (labelValue === undefined) return false;
 
-					if (matcher.isEqual && !matches) return false;
-					if (!matcher.isEqual && matches) return false;
+				if (rule.is_regex) {
+					try {
+						return new RegExp(rule.label_value).test(labelValue);
```
SoulKyu, line 547: **Non-blocking: empty `label_value` on a regex rule diverges from the server — `new RegExp('')` matches everything.**

On the server, `CompileFilterRules` (`internal/webui/services/hidden_alerts_service.go:225+`) only compiles a regex when `rule.IsRegex && rule.LabelValue != ""`, and `IsAlertHiddenByFilter` treats a missing compiled regex as no-match — so an enabled regex rule with an empty `label_value` never hides anything server-side. Here, `new RegExp('')` produces a regex that matches every string, so the same rule hides **every** alert carrying the label key on the client. Such a rule is creatable: `addFilterHiddenRule` (`dashboard_filter_presets.templ:334`) only requires `label_key` and defaults `label_value` to `''`.

Consequence: with an `is_regex: true, label_value: ''` rule, alerts carrying that label key vanish from the live SSE-updated view but reappear on full reload (and the inverse mismatch in the hidden display mode) — a small instance of the client/server disagreement this PR set out to remove.

Suggested change, mirroring the server's compile guard:

```js
if (rule.is_regex) {
	if (rule.label_value === '') return false;
	try {
		return new RegExp(rule.label_value).test(labelValue);
	} catch (e) {
		// Invalid user-supplied regex must not break the SSE merge
		return false;
	}
}
```

(Remember to re-run `make webui-templates` after editing the `.templ`.)


SoulKyu (commented):
Solid fix — I verified the new client mirror line-by-line against `applyDashboardFilters` and `HiddenAlertsService.IsAlertHiddenByFilter`, confirmed the preset hidden-alert/rule shapes (`{fingerprint,...}` objects, `{label_key, label_value, is_regex, is_enabled}` rules) match what `dashboard_filter_presets.templ` actually stores, checked the generated `dashboard_data_templ.go` is in sync, and `go build ./internal/webui/...` passes. All of issue #34's enumerated acceptance criteria are addressed.

Two non-blocking client/server divergences remain in the mirror, detailed inline: globally hidden *rules* are still not checked on the SSE path, and an empty `label_value` on a regex rule behaves opposite to the server. Neither is a regression introduced here, so this is safe to merge as-is.



internal/webui/templates/scripts/dashboard_data.templ
SoulKyu: @SoulKyu This has been fixed in cfffd92.

isGlobalHidden in dashboard_data.templ now also matches window.currentSettingsModal?.hiddenRules, normalizing the camelCase global-rule fields (labelKey/labelValue/isRegex/enabled) into the existing alertMatchesHiddenRule matcher, so rule-hidden alerts stay hidden (and stay visible in hidden display mode) across SSE updates.




internal/webui/templates/scripts/dashboard_data.templ
SoulKyu: @SoulKyu This has been fixed in cfffd92.

alertMatchesHiddenRule now returns false when is_regex is set and label_value is empty, mirroring the server's CompileFilterRules guard so an empty regex rule no longer hides every alert carrying the label key on the client.




SoulKyu:
**Looper fixer round complete** — cfffd92

- ✅ Review comment on `internal/webui/templates/scripts/dashboard_data.templ` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/37#discussion_r3644218691)
  - isGlobalHidden in dashboard_data.templ now also matches window.currentSettingsModal?.hiddenRules, normalizing the camelCase global-rule fields (labelKey/labelValue/isRegex/enabled) into the existing alertMatchesHiddenRule matcher, so rule-hidden alerts stay hidden (and stay visible in hidden display mode) across SSE updates.
- ✅ Review comment on `internal/webui/templates/scripts/dashboard_data.templ:547` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/37#discussion_r3644218698)
  - alertMatchesHiddenRule now returns false when is_regex is set and label_value is empty, mirroring the server's CompileFilterRules guard so an empty regex rule no longer hides every alert carrying the label key on the client.

PR #43, opened 2026-07-24
fix(dashboard): reset to page 1 when filters, sort, grouping or display mode change

### LINKED ISSUE #39: fix(dashboard): filtering, sorting or switching display mode keeps the stale page number and shows an empty table
## Problem

The dashboard keeps `currentPage` across every state change that shrinks the result set. `setItemsPerPage()` is the **only** mutator that resets it (`internal/webui/templates/scripts/dashboard_utilities.templ:498-502`):

```js
setItemsPerPage(value) {
    this.itemsPerPage = parseInt(value);
    this.currentPage = 1; // Reset to first page
    this.loadDashboardData();
},
```

None of the other entry points do:

- `applyFilters()` — `dashboard_utilities.templ:80-83` (called by the search box, `pages/NewDashboard.templ:372-373`, and by every filter checkbox/dropdown: `NewDashboard.templ:421, 482, 548, 607, 678, 748`)
- `clearAllFilters()` — `dashboard_utilities.templ:85-96`
- `clearFilter()` / `removeFilter()` — `dashboard_utilities.templ:98-109`
- `sortBy()` — `dashboard_utilities.templ:171-179`
- `setGroupByLabel()` — `dashboard_utilities.templ:526-529`
- `setDisplayMode()` — `scripts/dashboard_core.templ:424-448`
- `setViewMode()` — `scripts/dashboard_core.templ:450-459`
- `loadFilterPreset()` — `scripts/dashboard_filter_presets.templ:201`

`loadDashboardData()` then sends the stale page (`dashboard_data.templ:30`) and the server returns an empty slice for any out-of-range offset (`internal/webui/handlers/dashboard_handlers.go:494-501`):

```go
offset := (pagination.Page - 1) * pagination.Limit
if offset >= len(alerts) {
    return []*webuimodels.DashboardAlert{}
}
```

### Failure path

1. 300 firing alerts, 50 per page. The operator pages to **page 3**.
2. They type `postgres` in the search box, or tick a single team in the Team dropdown — 12 alerts match.
3. Request goes out as `page=3&limit=50` → offset 100 ≥ 12 → server returns `alerts: []` with `metadata.totalCount = 12`.
4. The table renders **empty** while the header claims there are results, and the pagination footer reads
   `Showing 101 to 12 of 12 results` — `getPaginationStartIndex()` returns `(3-1)*50+1 = 101` and `getPaginationEndIndex()` clamps to `totalItems` (`dashboard_utilities.templ:451-459`).
5. `getTotalPages()` is now `1`, so **Next is disabled** (`NewDashboard.templ:904`) and `getPageNumbers()` renders only `[1]` — with *no* page highlighted, since `currentPage === page` is false for every rendered button (`NewDashboard.templ:895`). The only way out is the Previous arrow or clicking `1`.

Same for switching display mode: on page 4 of *Classic* with 200 alerts, clicking **Acknowledged** (2 alerts) shows an empty Acknowledged tab. The natural reading is "no acknowledged alerts", which is wrong.

## Impact

Filtering and searching are the primary way operators narrow a busy dashboard, and paging past page 1 is normal on any real Alertmanager. The result is a blank table that looks like "nothing matches" — the operator concludes the filter is broken or that the alert they were hunting is gone, and either abandons the filter or reloads the page. Nothing in the UI points at the cause; the pagination footer actively contradicts itself (`101 to 12 of 12`).

## Proposed approach

Reset `currentPage = 1` on every state change that alters the result set, in the `.templ` sources (regenerate with `make webui-templates`; never edit `*_templ.go`):

- `dashboard_utilities.templ`: `applyFilters()`, `clearAllFilters()`, `sortBy()` (sorting reorders the whole set — page 3 of a re-sorted list is meaningless), `setGroupByLabel()`. `clearFilter()`/`removeFilter()` already route through `applyFilters()`.
- `dashboard_core.templ`: `setDisplayMode()`, `setViewMode()`.
- `dashboard_filter_presets.templ`: when a preset is loaded/applied.

The lazy version is to put the reset inside `loadDashboardData()`'s callers via a single `resetPageAndLoad()` helper and use it everywhere except the four genuine pagination actions (`goToPage`, `nextPage`, `previousPage`, and the initial load).

Belt-and-braces (optional, cheap): in `loadDashboardData()`, if the response comes back with `alerts.length === 0 && totalItems > 0 && currentPage > getTotalPages()`, clamp `currentPage` to `getTotalPages()` and re-fetch once.

## Acceptance criteria

- [ ] On page 3, typing a search term that matches fewer alerts than one page shows the matching alerts, not an empty table.
- [ ] On page 3, ticking a severity/team/status/alertname filter lands on page 1 of the filtered set.
- [ ] "Clear all filters" from a deep page returns to page 1.
- [ ] Changing the sort field/direction returns to page 1.
- [ ] Switching display mode (classic → acknowledged/hidden/full) returns to page 1.
- [ ] Switching list ↔ group view returns to page 1.
- [ ] Loading a saved filter preset returns to page 1.
- [ ] The pagination footer can never read a start index greater than the total (`Showing 101 to 12 of 12` is not reachable).
- [ ] Explicit pagination (page numbers, prev/next) is unaffected.

SoulKyu:
## Problem

The dashboard kept `currentPage` across every state change that shrinks or reorders the result set — only `setItemsPerPage()` reset it. On page 3 with 50/page, typing a search that matches 12 alerts sent `page=3&limit=50`; the server returned an empty slice for the out-of-range offset and the table rendered blank while the footer read `Showing 101 to 12 of 12 results`, with no page button highlighted and Next disabled.

## Changes

Reset `currentPage = 1` in every mutator that changes the result set, in the `.templ` sources (regenerated with `make webui-templates`):

- `internal/webui/templates/scripts/dashboard_utilities.templ`: `applyFilters()` (covers search box, all filter dropdowns, and `clearFilter()`/`removeFilter()` which route through it), `clearAllFilters()`, `sortBy()`, `setGroupByLabel()`
- `internal/webui/templates/scripts/dashboard_core.templ`: `setDisplayMode()`, `setViewMode()`
- `internal/webui/templates/scripts/dashboard_filter_presets.templ`: `applyFilterPreset()` (covers `loadPreset()`)

Explicit pagination (`goToPage`, `nextPage`, `previousPage`) and the initial load are untouched.

The optional response-side clamp in `loadDashboardData()` was skipped: with every UI mutator resetting the page, the out-of-range state is no longer reachable through the UI.

## Validation

- `make webui-templates` regenerated cleanly; generated `*_templ.go` files carry the resets
- `go build ./...` passes

Closes #39


internal/webui/templates/scripts/dashboard_utilities.templ
```diff
@@ -79,6 +79,7 @@ templ DashboardUtilities() {
 			// Filtering utilities
 			applyFilters() {
 				// Debounced filter application happens automatically via data loading
+				this.currentPage = 1; // Result set changes, stale page would be out of range
 				this.loadDashboardData();
```
SoulKyu: **Non-blocking (pre-existing, but this PR touches this function):** `applyFilters()` does not return the promise from `loadDashboardData()`, yet the search box chains onto its return value in `pages/NewDashboard.templ:375-376`: `@input.debounce.300ms="isSearching = true; applyFilters().finally(() => { isSearching = false; })"`. Because `applyFilters()` returns `undefined`, `.finally()` throws a `TypeError` on every debounced keystroke and Enter press, so `isSearching` is never reset — the search spinner (the `animate-spin` icon shown at `NewDashboard.templ:369` while `isSearching` is true) sticks on permanently after the first search, and the console fills with errors.

The page-reset added here still works because `loadDashboardData()` is invoked before the caller's `.finally()` blows up, so this doesn't block the fix. But since this function is being edited anyway, the one-line repair is to return the promise:

```suggestion
				return this.loadDashboardData();
```

(then regenerate with `make webui-templates`). `loadDashboardData` is `async` (`dashboard_data.templ:6`), so the returned promise makes the `.finally()` chain behave as intended, and no existing caller depends on the current `undefined` return.


SoulKyu (commented):
Reviewed the pagination-reset change across the three `.templ` sources and their regenerated `*_templ.go` files. The fix is well targeted: every UI mutator that shrinks or reorders the result set (`applyFilters`, `clearAllFilters`, `sortBy`, `setGroupByLabel`, `setDisplayMode`, `setViewMode`, `applyFilterPreset`) now resets `currentPage` to 1 before reloading, explicit pagination is untouched, and I verified the generated files match the sources (same reset counts in each `_templ.go`) and that the resolved view's separate `resolvedPagination` offset is unaffected by the reset in `setDisplayMode`.

One non-blocking, pre-existing issue adjacent to the touched code is noted inline; it does not affect the correctness of this fix.



SoulKyu:
🔀 Rebase (conflicts resolved): rebased onto `origin/main` (`5f4019a` + merge `80858f1`), force-pushed as `44677a3`.

**Conflict**: 1 file pair — `internal/webui/templates/scripts/dashboard_utilities.templ` (+ its generated `_templ.go`), in `applyFilters()`.

Both sides edited the exact same two lines:
- **main** (#42, `5f4019a`): `return this.loadDashboardData();` — returns the promise so the search box can chain `.finally()` and stop its spinner.
- **this PR**: `this.currentPage = 1;` before `this.loadDashboardData();` — resets pagination so a stale out-of-range offset no longer renders an empty table.

The two changes are orthogonal, so both were kept:

```js
applyFilters() {
    this.currentPage = 1; // Result set changes, stale page would be out of range
    // Return the promise so callers like the search box can chain .finally()
    return this.loadDashboardData();
},
```

Order matters: the page reset must happen *before* the fetch, and the return value still propagates to the caller — both behaviours are intact.

The other five hunks of this PR (`dashboard_core.templ`, `dashboard_filter_presets.templ`, and the remaining `dashboard_utilities.templ` mutators) applied cleanly.

**Generated files**: `*_templ.go` were **not** hand-edited — only the `.templ` source was resolved, then regenerated with `make webui-templates`.

**Verification**:
- `go build ./...` → passes.
- `go vet ./...` → output is byte-for-byte identical to `origin/main` (diffed against a clean worktree of main). The single remaining finding, `internal/webui/router.go:52:3: log.Fatalf does not support error-wrapping directive %w`, is **pre-existing on main** and unrelated to this PR, so it was left untouched rather than smuggled into a rebase commit. Worth a separate one-line fix (`%w` → `%v`).

internal/webui/templates/scripts/dashboard_utilities.templ
SoulKyu: @SoulKyu I'm not making a code change for this thread.

Already implemented on this branch: applyFilters() at internal/webui/templates/scripts/dashboard_utilities.templ:83 already contains `return this.loadDashboardData();` (commit 5f4019a from PR #42, in this branch's history), and the regenerated dashboard_utilities_templ.go carries it. No change needed.

PR #52, opened 2026-07-24
perf(dashboard): fetch colors only for changed alerts on SSE updates

### LINKED ISSUE #46: perf(dashboard): every SSE update makes each browser refetch colors for the entire filtered alert set
## Problem

The SSE broadcast built in `AlertCache.refreshAlerts` never carries colors — colors are per-user, the stream is a shared fan-out:

```go
// internal/webui/services/alert_cache.go:241-249
update := &webuimodels.DashboardIncrementalUpdate{
    NewAlerts: newAlertsForSSE, UpdatedAlerts: updatedAlertsForSSE,
    RemovedAlerts: removedFingerprints, LastUpdateTime: time.Now().Unix(),
}   // no Colors
ac.notifySubscribers(update)
```

So the client always falls into the "fetch them" branch:

```js
// internal/webui/templates/scripts/dashboard_data.templ:366-375
} else if (this.sseConnection && (update.newAlerts?.length > 0 || update.updatedAlerts?.length > 0)) {
    this._colorLoadTimeout = setTimeout(() => { this.loadAlertColors(true); }, 500);
}
```

`loadAlertColors(true)` hits `GET /api/v1/dashboard/alert-colors`, and that handler **ignores pagination entirely**: it rebuilds the full alert set for the display mode (including a backend round-trip for resolved alerts in `full`/`resolved` mode), filters it, and computes + serialises a color entry for *every* filtered alert:

```go
// internal/webui/handlers/dashboard_handlers.go:1729-1763
switch filters.DisplayMode { ... allAlerts = alertCache.GetResolvedAlertsWithLimit(...) ... }
filteredAlerts := applyDashboardFilters(allAlerts, filters, sessionID)
finalResults := computeAlertColorsMap(filteredAlerts, sessionID)   // not paginatedAlerts
```

The alert cache polls Alertmanager every 10 s (`config/config.go:191`) and broadcasts whenever *any* alert is new or changed (`hasAlertChanged`, `alert_cache.go:339-361`). In a production environment where something is always firing/silencing/resolving, that is one full color refetch per browser per poll cycle — for a payload the client mostly already has, of which it renders 50 rows (`limit` default, `dashboard_handlers.go:276`).

## Impact

Per connected user, with a few thousand alerts: a `~100 KB–1 MB` JSON response every ~10 s, plus the server-side filter pass and MD5 fingerprinting of every alert on each request, plus one `ValidateSession` gRPC round-trip. On the client, `this.alertColors = result.data.colors` replaces the whole reactive object, so Alpine re-evaluates color bindings for the entire table. Scale that by the number of open dashboards — a NOC wall with a dozen tabs turns a 10 s poll into a steady multi-MB/s of pure overhead. Users see it as a dashboard that gets sluggish exactly when alert churn is highest.

## Proposed approach

The alerts that changed are already in hand on the client, and a suitable endpoint already exists: `POST /api/v1/dashboard/alerts/bulk-colors` (`router.go:257` → `GetBulkAlertColors`, `dashboard_handlers.go:1792`) takes `{alerts: [{fingerprint, labels}]}` and returns colors for exactly those alerts.

- In `applyIncrementalUpdate` (`internal/webui/templates/scripts/dashboard_data.templ:366-375`), replace the `loadAlertColors(true)` fallback with a bulk-colors POST for `update.newAlerts ∪ update.updatedAlerts`, merging the result into `this.alertColors` the same way the non-SSE branch does.
- While in `GetAlertColors`: compute colors for the paginated slice rather than the full filtered set, so the remaining callers stop paying for off-screen alerts.

Templ workflow applies: edit the `.templ` sources and regenerate with `make webui-templates`; never edit `*_templ.go`.

## Acceptance criteria

- With SSE connected and a handful of alerts changing, the browser issues **no** `GET /api/v1/dashboard/alert-colors` request; it issues one `POST /alerts/bulk-colors` whose body contains only the new/updated fingerprints.
- Newly arrived alerts are rendered with their correct user colors (no grey flash, no color lag) — the current behaviour is preserved.
- Response size for the color refresh scales with the number of changed alerts, not with the total number of alerts.
- A full `loadDashboardData()` still returns embedded colors for the rendered page (unchanged behaviour, `dashboard_handlers.go:211`).

SoulKyu:
## Problem

The SSE broadcast is a shared fan-out and never carries per-user colors, so every incremental update pushed each connected browser into `loadAlertColors(true)` → `GET /api/v1/dashboard/alert-colors`. That handler rebuilt and filtered the full alert set and computed a color entry for **every** filtered alert — a payload scaling with total alert count (100 KB–1 MB with a few thousand alerts), refetched by every open dashboard on every ~10 s poll cycle under alert churn.

## Changes

**Client (`internal/webui/templates/scripts/dashboard_data.templ`)**
- `applyIncrementalUpdate` no longer falls back to a full `loadAlertColors(true)`. It accumulates the changed alerts (`newAlerts ∪ updatedAlerts`, keyed by fingerprint) into a pending map and, after the existing 500 ms debounce, sends one `POST /api/v1/dashboard/alerts/bulk-colors` with only those `{fingerprint, labels}` pairs.
- The response is merged into `this.alertColors` (spread merge, same as the embedded-colors path), so unchanged rows keep their color objects and Alpine doesn't re-evaluate bindings for the whole table.
- Pending alerts accumulate across debounced updates, so bursts don't drop any fingerprints. Past the endpoint's 1000-alert cap it falls back to the old full refresh.
- `*_templ.go` regenerated with the go.mod-pinned templ v0.3.906.

**Server (`internal/webui/handlers/dashboard_handlers.go`)**
- `GetAlertColors` now applies the same sorting + pagination as `GetDashboardData` before computing colors, so the remaining callers (initial-load fallback, preference refresh — which already send `page`/`limit`/`sortField` params) pay for the rendered page only, not every off-screen alert.

`GetDashboardData` still embeds colors for the rendered page — the first paint is unchanged.

## Validation

- `go build ./...` and `go test ./internal/webui/...` (36 tests, 11 packages) pass.
- With SSE connected and alerts changing: no `GET /alert-colors` request; one debounced `POST /alerts/bulk-colors` containing only the changed fingerprints, response size proportional to the number of changed alerts.

Closes #46


internal/webui/templates/scripts/dashboard_data.templ
```diff
@@ -250,6 +250,49 @@ templ DashboardData() {
 				}
 			},
 
+			// Fetch colors for only the pending changed alerts (SSE path) via the
+			// bulk-colors endpoint, merging results into the existing color map.
+			// Payload scales with changed alerts, not the full filtered set.
+			async loadBulkAlertColors() {
+				const pending = this._pendingColorAlerts || {};
+				this._pendingColorAlerts = {};
```
SoulKyu, line 258: **Failed bulk fetches permanently drop the pending fingerprints (no self-heal).**

`_pendingColorAlerts` is cleared here, *before* the fetch. If the request then fails — the `catch` block, the `!result.success` branch, or the `handleAuthError` early return — the method only logs, and those fingerprints are never retried. The old code path self-healed: any later SSE churn triggered a full `loadAlertColors(true)` that re-covered every alert. With this change, an alert whose one bulk fetch failed keeps a stale or missing color entry until the user does a full reload; only alerts that happen to churn again get re-fetched.

Suggested fix: on failure, merge the batch back into the pending map without clobbering entries added in the meantime, so the next debounced flush retries it:

```js
} catch (error) {
	console.error('Error loading bulk alert colors:', error);
	this._pendingColorAlerts = { ...pending, ...this._pendingColorAlerts };
}
```

and the same re-merge in the `!result.success` branch. (The `handleAuthError` return can stay as-is — that path redirects to login anyway.)


internal/webui/templates/scripts/dashboard_data.templ
```diff
@@ -250,6 +250,49 @@ templ DashboardData() {
...
+					});
+
+					if (window.dashboardInstance && window.dashboardInstance.handleAuthError(response)) {
+						return;
+					}
+
+					const result = await response.json();
+
+					if (result.success && result.data.colors) {
+						this.alertColors = { ...this.alertColors, ...result.data.colors };
```
SoulKyu, line 285: **`alertColors` now grows without bound over long-lived SSE sessions.**

Before this PR, the SSE path periodically ran `loadAlertColors(true)`, which *replaced* the whole map (`this.alertColors = result.data.colors || {}`) and so implicitly pruned entries for alerts that no longer exist. Now the SSE path only ever spread-merges into the map, and nothing removes entries when alerts are resolved/removed. On an always-open dashboard (the typical SRE wall-screen case) with alert churn, every new fingerprint adds an entry that is never evicted — and each merge spread-copies the whole map, so the per-update cost also creeps up with the accumulated size.

Suggested fix: prune in `applyIncrementalUpdate`'s `removedAlerts` handling, e.g.:

```js
update.removedAlerts.forEach(fp => delete this.alertColors[fp]);
```

(then reassign `this.alertColors = { ...this.alertColors }` if you need the reactivity nudge). That keeps the map bounded by the set of alerts the client actually knows about.


SoulKyu (commented):
Reviewed the SSE bulk-color path end to end: the client accumulation/debounce logic, the new `POST /alerts/bulk-colors` usage, and the sort+pagination added to `GetAlertColors`. The server-side pagination matches `GetDashboardData` exactly (flat sort → paginate, then group), the only `/alert-colors` caller already sends `sortField`/`page`/`limit`, fingerprint remapping in the bulk endpoint is labels-derived on both sides, and color matching only uses labels, so the bulk path loses no fidelity. `go build` passes and the regenerated `*_templ.go` contains the new code.

Two non-blocking robustness points on the new client code, inline below.



SoulKyu:
## 🧪 QA Report

Env: fresh `make test` images (webui image rebuilt from this head, binary contains `loadBulkAlertColors`), run as an isolated compose project on ports 18081/19093 to avoid clobbering a pre-existing stack on this host. Real Chromium (Playwright) driving login → dashboard → live SSE churn, with a user colour rule `severity=critical → #FF00FF` so colours are observable.

|---|---|---|
| SSE connected + a handful of alerts change → **no** `GET /alert-colors`, one `POST /alerts/bulk-colors` with only new/updated fingerprints | ✅ | Injected 6 critical alerts into the fake Alertmanager, watched 45 s of SSE churn. Requests after the marker were **exclusively** `POST /api/v1/dashboard/alerts/bulk-colors` (5 calls, one per SSE batch); `GET /alert-colors` count = **0**. Bodies carried only the changed fingerprints (2, 8, 10, 14…) out of 181 total alerts, each with non-empty `labels`. Console: `Loaded colors for 14 changed alerts via bulk endpoint`. |
| Newly arrived alerts render with correct user colours (no grey flash, no colour lag) | ✅ | The 6 SSE-delivered `QASse*` alerts (`severity=critical`) each resolved to `{backgroundColor:"#FF00FF19", textColor:"#b200b2", borderColor:"#FF00FF", colorSource:"user"}` — the user rule, not the grey/severity fallback. `hasColor: true` for every one; unchanged rows kept their existing colour objects (spread merge, `colorsBefore 45 → colorsAfter 53`). |
| Colour-refresh response size scales with changed alerts, not total | ✅ | Measured on the same dataset: 1 changed → 231 B, 3 → 549 B, 14 → 2281 B, vs the full filtered set (80 alerts) at 12 795 B. Server side, `GET /alert-colors?page=1&limit=50` now returns 50 colours / 7.9 KB instead of all 181. |
| Full `loadDashboardData()` still embeds colours for the rendered page | ✅ | `GET /api/v1/dashboard/data` returns `colors` whose key set is **exactly** the page's fingerprints, on every page: p1 50/50, p2 50/50, p3 20/20 (total 120). First paint unchanged. |

**Exploratory:** No adjacent breakage. `GET /alert-colors` pagination stays consistent with what's rendered — page 1/2/3 colour sets match their own alerts and page 1 vs 2 are disjoint (group view also renders `paginatedAlerts`, so the new pagination can't starve it). Page 2 navigation: 43/43 rows coloured from embedded colours, no extra colour request. Preference change still does the full `GET /alert-colors` refresh and recolours the table (`#FF00FF → #00FF00`). Group view: 50/50 alerts coloured. Regenerating templates with the go.mod-pinned templ v0.3.906 reproduces the committed `dashboard_data_templ.go` byte-for-byte. Only console errors were pre-auth 401s on `/api/v1/profile/timezone` and `/api/impersonate/status` — unrelated to this change.

Minor, non-blocking: on the SSE path `this.alertColors` is now only ever merged into, never pruned, so a long-lived tab accumulates colour entries for alerts that have since left the view. Bounded by distinct fingerprints seen, so not a regression against the reported problem — just worth knowing before someone leaves a NOC tab open for a week.

**Verdict: PASS**

internal/webui/templates/scripts/dashboard_data.templ
SoulKyu: @SoulKyu This has been fixed in c97825d.

In loadBulkAlertColors, both the catch block and the !result.success branch now re-merge the failed batch into _pendingColorAlerts via { ...pending, ...this._pendingColorAlerts }, preserving entries added in the meantime, so the next debounced flush retries the failed fingerprints.




internal/webui/templates/scripts/dashboard_data.templ
SoulKyu: @SoulKyu This has been fixed in c97825d.

applyIncrementalUpdate's removedAlerts handling now deletes each removed fingerprint from alertColors (and from _pendingColorAlerts to skip fetching colors for gone alerts), keeping the maps bounded by the alerts the client knows about.




SoulKyu:
**Looper fixer round complete** — c97825d

- ✅ Review comment on `internal/webui/templates/scripts/dashboard_data.templ:258` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/52#discussion_r3645234863)
  - In loadBulkAlertColors, both the catch block and the !result.success branch now re-merge the failed batch into _pendingColorAlerts via { ...pending, ...this._pendingColorAlerts }, preserving entries added in the meantime, so the next debounced flush retries the failed fingerprints.
- ✅ Review comment on `internal/webui/templates/scripts/dashboard_data.templ:285` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/52#discussion_r3645234872)
  - applyIncrementalUpdate's removedAlerts handling now deletes each removed fingerprint from alertColors (and from _pendingColorAlerts to skip fetching colors for gone alerts), keeping the maps bounded by the alerts the client knows about.

PR #66, opened 2026-07-24
perf(backend): bound GetAllAcknowledgedAlerts to live alert keys

### LINKED ISSUE #47: perf(backend): GetAllAcknowledgedAlerts self-joins the whole acknowledgments table every 10s and ships every row to the WebUI
## Problem

Every alert-cache refresh cycle (default 10 s, `config/config.go:191`) unconditionally asks the backend for **every acknowledgment ever created**:

```go
// internal/webui/services/alert_cache.go:407-438  (called from loadBackendData on every refreshAlerts)
go ac.loadAcknowledgmentsEfficiently()
...
acknowledgedAlerts, err := ac.backendClient.GetAllAcknowledgedAlerts()   // :414 — no arguments
for fingerprint, acknowledgment := range acknowledgedAlerts {
    if alert, exists := ac.alerts[fingerprint]; exists { ... }           // :423-424 — the rest is discarded
}
```

The backing query is a full scan plus a self-join over a `GROUP BY` subquery, with no time bound and no key filter:

```go
// internal/backend/database/gorm_db.go:477-496
gdb.db.Table("acknowledgments").
    Select("acknowledgments.*, users.username").
    Joins("JOIN users ON users.id = acknowledgments.user_id").
    Joins("JOIN (SELECT alert_key, MAX(created_at) as max_created FROM acknowledgments GROUP BY alert_key) latest ON ...").
    Find(&acks)
```

And `acknowledgments` is never pruned. Rows are deleted only on an explicit user un-ack (`gorm_db.go:466`) or when the same user re-acks the same key (`:425`). The archival path (`storeResolvedAlertInBackend` → `CreateResolvedAlert`) copies acks into the `resolved_alerts` JSON blob but leaves the source rows in place, and the only retention job in the codebase is `CleanupExpiredResolvedAlerts` (`internal/backend/server.go:321`). The table therefore grows monotonically for the life of the deployment, while the query over it runs 8 640 times a day per WebUI instance and its **entire** result — `alert_key`, `reason` text, `username` for every alert ever acknowledged — is marshalled to protobuf and shipped over gRPC, only for the WebUI to keep the handful of entries matching currently-firing fingerprints.

The index situation makes it worse: `Acknowledgment.AlertKey` is indexed on its own (`internal/backend/models/models.go:89`), but the subquery groups the whole table and the outer join matches on `(alert_key, created_at)`, so neither side is covered.

Note the fix pattern already exists 100 lines above, for exactly the same problem on comments:

```go
// internal/backend/database/gorm_db.go:379-411
// GetCommentCountsBatch retrieves comment counts for multiple alert keys in a single query.
// This solves the N+1 query problem when loading comment counts for many alerts.
```

and the WebUI already collects the fingerprint list it needs for it (`alert_cache.go:445-459`).

## Impact

Cost grows with the deployment's entire history rather than with the number of alerts on screen: a team acknowledging ~100 alerts/day accumulates ~36 k rows a year, and every one of them is scanned, joined, serialised and sent over the wire six times a minute forever. On the SQLite deployments the project supports this is the query most likely to become the WebUI's steady-state DB load, and it degrades monotonically — the longer the install has been useful, the slower it gets, with no knob for an operator to reclaim it.

## Proposed approach

Mirror `GetCommentCountsBatch`:

- Add an `alert_keys` field to `GetAllAcknowledgedAlertsRequest` in `proto/alert.proto` (or a new `GetAcknowledgmentsBatch` RPC) and regenerate with `make proto` — never hand-edit `*.pb.go`.
- `GormDB.GetAllAcknowledgedAlerts` takes the key list and adds `WHERE alert_key IN ?` to both the outer query and the `latest` subquery, so the scan is bounded by the number of live alerts.
- `loadAcknowledgmentsEfficiently` collects fingerprints under `RLock` first — the same three-step shape `loadCommentCountsEfficiently` already uses (`alert_cache.go:441-488`) — and passes them in.
- Add a composite index on `(alert_key, created_at)` for `Acknowledgment` so the "latest per key" lookup is index-served.

Keeping an unfiltered variant is fine if another caller needs it, but the 10 s hot path must not use it.

## Acceptance criteria

- A refresh cycle with N cached alerts sends at most N `alert_key` values and receives at most N acknowledgments; acknowledgments for alerts that are not currently cached never cross the gRPC boundary.
- Acknowledgment badges on the dashboard behave exactly as today (ack shown, un-ack clears it, ack by another user visible after one cycle).
- The SQL executed on the hot path is bounded by `alert_key IN (...)` on both the outer query and the `latest` subquery.
- A composite `(alert_key, created_at)` index exists on `acknowledgments` after migration.
- The empty-cache case makes no DB round-trip at all (same early return as `GetCommentCountsBatch`, `gorm_db.go:385-387`).

SoulKyu:
## Problem

Every WebUI alert-cache refresh cycle (default 10 s) called `GetAllAcknowledgedAlerts` with no arguments, which ran a full scan of the `acknowledgments` table plus a self-join over an unbounded `GROUP BY` subquery, then shipped **every acknowledgment ever created** over gRPC — only for the WebUI to keep the handful matching currently-cached fingerprints. The table is never pruned, so cost grew with deployment history rather than with alerts on screen.

## Changes

Mirrors the existing `GetCommentCountsBatch` pattern:

- **`proto/alert.proto`**: added `repeated string alert_keys = 1` to `GetAllAcknowledgedAlertsRequest`; regenerated with `make proto` (no hand edits to `*.pb.go`).
- **`internal/backend/database/gorm_db.go`**: `GetAllAcknowledgedAlerts` now takes the key list and applies `alert_key IN (...)` to both the outer query and the `latest` (max `created_at` per key) subquery, so the scan is bounded by the number of live alerts. Empty input returns immediately with no DB round-trip.
- **`internal/backend/models/models.go`**: composite index `idx_acknowledgments_alert_key_created_at` on `(alert_key, created_at)` so the latest-per-key lookup is index-served; created by the existing `AutoMigrate` path.
- **`internal/webui/services/alert_cache.go`**: `loadAcknowledgmentsEfficiently` collects cached fingerprints under `RLock` first (same three-step shape as `loadCommentCountsEfficiently`) and passes them to the backend; skips the gRPC call entirely when the cache is empty.
- **`internal/backend/services/*.go`, `internal/webui/client/backend_client.go`**: plumb the key list through the RPC service and client.

## Validation

- `go build ./...` and `go test ./...` pass (pre-existing `go vet` warning in `internal/webui/router.go:52` is untouched by this change).
- New `internal/backend/database/gorm_db_test.go` covers: latest-per-key selection with username join, filtering to requested keys only, empty-input early return, and presence of the composite index after migration.

## Acceptance criteria

- A refresh with N cached alerts sends at most N keys and receives at most N acknowledgments; acks for uncached alerts never cross the gRPC boundary.
- Badge behaviour unchanged: ack shown, un-ack clears, another user's ack visible after one cycle (same latest-per-key + username query, just bounded).
- Hot-path SQL bounded by `alert_key IN (...)` on both the outer query and the subquery.
- Composite `(alert_key, created_at)` index exists after migration.
- Empty cache makes no DB round-trip.

Closes #47

### DIFF
--- a/internal/backend/models/models.go
+++ b/internal/backend/models/models.go
@@ -91,10 +91,10 @@ func (c *Comment) BeforeCreate(tx *gorm.DB) error {
 
 type Acknowledgment struct {
 	ID        string    `gorm:"primaryKey;type:varchar(32)" json:"id"`
-	AlertKey  string    `gorm:"not null;size:500;index" json:"alert_key"`
+	AlertKey  string    `gorm:"not null;size:500;index;index:idx_acknowledgments_alert_key_created_at,priority:1" json:"alert_key"`
 	UserID    string    `gorm:"not null;size:32" json:"user_id"`
 	Reason    string    `gorm:"not null;type:text" json:"reason"`
-	CreatedAt time.Time `json:"created_at"`
+	CreatedAt time.Time `gorm:"index:idx_acknowledgments_alert_key_created_at,priority:2" json:"created_at"`
 	UpdatedAt time.Time `json:"updated_at"`
 
 	User User `gorm:"foreignKey:UserID" json:"user,omitempty"`
--- a/internal/backend/database/gorm_db.go
+++ b/internal/backend/database/gorm_db.go
@@ -474,20 +474,34 @@ func (gdb *GormDB) DeleteAcknowledgment(alertKey, userID string) error {
 	return nil
 }
 
-func (gdb *GormDB) GetAllAcknowledgedAlerts() (map[string]models.AcknowledgmentWithUser, error) {
+// GetAllAcknowledgedAlerts retrieves the latest acknowledgment for each of the
+// given alert keys in a single query, mirroring GetCommentCountsBatch so the
+// scan is bounded by the number of live alerts instead of the whole table.
+func (gdb *GormDB) GetAllAcknowledgedAlerts(alertKeys []string) (map[string]models.AcknowledgmentWithUser, error) {
+	result := make(map[string]models.AcknowledgmentWithUser)
+
+	if len(alertKeys) == 0 {
+		return result, nil
+	}
+
 	var acks []models.AcknowledgmentWithUser
+
+	latest := gdb.db.Table("acknowledgments").
+		Select("alert_key, MAX(created_at) as max_created").
+		Where("alert_key IN ?", alertKeys).
+		Group("alert_key")
 
 	err := gdb.db.Table("acknowledgments").
 		Select("acknowledgments.*, users.username").
 		Joins("JOIN users ON users.id = acknowledgments.user_id").
-		Joins("JOIN (SELECT alert_key, MAX(created_at) as max_created FROM acknowledgments GROUP BY alert_key) latest ON acknowledgments.alert_key = latest.alert_key AND acknowledgments.created_at = latest.max_created").
+		Joins("JOIN (?) latest ON acknowledgments.alert_key = latest.alert_key AND acknowledgments.created_at = latest.max_created", latest).
+		Where("acknowledgments.alert_key IN ?", alertKeys).
 		Find(&acks).Error
 
 	if err != nil {
 		return nil, err
 	}
 
-	result := make(map[string]models.AcknowledgmentWithUser)
 	for _, ack := range acks {
 		result[ack.AlertKey] = ack
 	}
--- a/internal/backend/database/gorm_db_test.go
+++ b/internal/backend/database/gorm_db_test.go
@@ -1,0 +1,81 @@
+package database
+
+import (
+	"testing"
+	"time"
+
+	"gorm.io/driver/sqlite"
+	"gorm.io/gorm"
+	"gorm.io/gorm/logger"
+
+	"notificator/internal/backend/models"
+)
+
+func newTestDB(t *testing.T) *GormDB {
+	t.Helper()
+	db, err := gorm.Open(sqlite.Open("file::memory:"), &gorm.Config{
+		Logger: logger.Default.LogMode(logger.Silent),
+	})
+	if err != nil {
+		t.Fatalf("open sqlite: %v", err)
+	}
+	if err := db.AutoMigrate(&models.User{}, &models.Acknowledgment{}); err != nil {
+		t.Fatalf("migrate: %v", err)
+	}
+	return &GormDB{db: db, dbType: "sqlite"}
+}
+
+func TestGetAllAcknowledgedAlerts(t *testing.T) {
+	gdb := newTestDB(t)
+
+	alice := models.User{ID: "u1", Username: "alice", Email: "alice@example.com"}
+	bob := models.User{ID: "u2", Username: "bob", Email: "bob@example.com"}
+	if err := gdb.db.Create([]*models.User{&alice, &bob}).Error; err != nil {
+		t.Fatalf("create users: %v", err)
+	}
+
+	base := time.Now().UTC().Truncate(time.Second)
+	acks := []models.Acknowledgment{
+		{ID: "a1", AlertKey: "key-a", UserID: alice.ID, Reason: "old", CreatedAt: base.Add(-time.Hour)},
+		{ID: "a2", AlertKey: "key-a", UserID: bob.ID, Reason: "latest", CreatedAt: base},
+		{ID: "b1", AlertKey: "key-b", UserID: alice.ID, Reason: "only", CreatedAt: base},
+		{ID: "c1", AlertKey: "key-not-cached", UserID: alice.ID, Reason: "stale", CreatedAt: base},
+	}
+	for i := range acks {
+		if err := gdb.db.Create(&acks[i]).Error; err != nil {
+			t.Fatalf("create ack: %v", err)
+		}
+	}
+
+	result, err := gdb.GetAllAcknowledgedAlerts([]string{"key-a", "key-b", "key-missing"})
+	if err != nil {
+		t.Fatalf("GetAllAcknowledgedAlerts: %v", err)
+	}
+	if len(result) != 2 {
+		t.Fatalf("expected 2 acks, got %d: %v", len(result), result)
+	}
+	if got := result["key-a"]; got.Reason != "latest" || got.Username != "bob" {
+		t.Errorf("key-a: expected latest ack by bob, got reason=%q user=%q", got.Reason, got.Username)
+	}
+	if got := result["key-b"]; got.Reason != "only" || got.Username != "alice" {
+		t.Errorf("key-b: expected ack by alice, got reason=%q user=%q", got.Reason, got.Username)
+	}
+	if _, ok := result["key-not-cached"]; ok {
+		t.Errorf("key-not-cached must not be returned when not requested")
+	}
+
+	empty, err := gdb.GetAllAcknowledgedAlerts(nil)
+	if err != nil {
+		t.Fatalf("empty keys: %v", err)
+	}
+	if len(empty) != 0 {
+		t.Fatalf("expected no acks for empty key list, got %v", empty)
+	}
+}
+
+func TestAcknowledgmentCompositeIndexExists(t *testing.T) {
+	gdb := newTestDB(t)
+	if !gdb.db.Migrator().HasIndex(&models.Acknowledgment{}, "idx_acknowledgments_alert_key_created_at") {
+		t.Fatal("composite index idx_acknowledgments_alert_key_created_at missing after migration")
+	}
+}
--- a/internal/backend/proto/alert/alert.pb.go
+++ b/internal/backend/proto/alert/alert.pb.go
@@ -841,6 +841,7 @@ func (x *GetAcknowledgmentsResponse) GetCount() int32 {
 
 type GetAllAcknowledgedAlertsRequest struct {
 	state         protoimpl.MessageState `protogen:"open.v1"`
+	AlertKeys     []string               `protobuf:"bytes,1,rep,name=alert_keys,json=alertKeys,proto3" json:"alert_keys,omitempty"` // Only return acknowledgments for these alert keys
 	unknownFields protoimpl.UnknownFields
 	sizeCache     protoimpl.SizeCache
 }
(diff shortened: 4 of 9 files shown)


SoulKyu (commented):
@SoulKyu Clean pass — nice, tightly-scoped perf fix. I verified the bounded query end to end: the `alert_key IN (...)` filter is applied to both the outer query and the latest-per-key subquery in `GetAllAcknowledgedAlerts`, the empty-input path returns without a DB round-trip, and the new `(alert_key, created_at)` composite index is created by the existing AutoMigrate path. The WebUI side correctly snapshots fingerprints under `RLock` and calls gRPC with no lock held, mirroring the established `loadCommentCountsEfficiently` pattern, and acknowledgments are keyed by fingerprint so the filter matches what the cache holds. The proto addition (`repeated string alert_keys = 1` in a previously-empty message) is wire-compatible, and all in-repo callers are plumbed through. `go build ./...` and the affected package tests pass locally, and the new sqlite-backed test covers latest-per-key selection, key filtering, empty input, and index presence. Solid, well-tested work that follows the codebase's existing batch pattern — thanks!



SoulKyu:
## 🧪 QA Report

Verified against a full `make test` docker-compose stack (postgres + backend + webui + fake alertmanager), driving the real WebUI in headless Chromium and the backend over gRPC (`grpcurl`), with `log_statement='all'` on postgres to capture the hot-path SQL.

|---|---|---|
| A refresh with N cached alerts sends at most N keys and receives at most N acks | ✅ | With 26 alerts cached (`fetched 26 alerts from Alertmanager`), the captured statement carried exactly 26 bind params in the subquery (`$1..$26`) and 26 in the outer query (`$27..$52`). DB held 5 ack rows over 4 distinct keys; webui logged `Received 2 acknowledged alerts from backend` — only the keys currently cached. |
| Acks for uncached alerts never cross the gRPC boundary | ✅ | Inserted `alert_key='qa-ghost-uncached-key'` directly in postgres. Over 3 consecutive refresh cycles the webui kept reporting `Received 1 acknowledged alerts` (the cached one only). Requesting that key explicitly over gRPC returns it (`count: 1`), proving the row was live and the key filter — not row absence — excluded it. |
| Badge shown after ack | ✅ | Logged in as `qaalice` in a real browser, clicked the row's ✓ Acknowledge, submitted a reason. Acknowledged tab renders the row with a `✓ ACK` badge; client state reads `isAcknowledged:true, acknowledgedBy:"qaalice"`. Row also disappears from Classic as expected. |
| Un-ack clears the badge | ✅ | Un-acknowledged `db-18.example.com` from the alert-details modal. Row deleted from `acknowledgments`; a **fresh** login one cycle later finds it absent from the Acknowledged view (`[]`) — the flag is gone, not cached. |
| Ack by another user visible after one cycle | ✅ | `qabob` acknowledged `db-16.example.com` in a separate browser session; `qaalice`'s session showed it within one 10 s cycle with `acknowledgedBy:"qabob"`, `acknowledgeReason:"QA-66 acked by BOB"`. Latest-per-key also verified: a second, later ack on an already-acked key flipped the UI from `qaalice`/`first ack alice` to `qabob`/`second ack BOB wins`. |
| Hot-path SQL bounded by `alert_key IN (...)` on both outer query and `latest` subquery | ✅ | Captured live: `SELECT acknowledgments.*, users.username FROM "acknowledgments" JOIN users ... JOIN (SELECT alert_key, MAX(created_at) as max_created FROM "acknowledgments" WHERE alert_key IN ($1..$22) GROUP BY "alert_key") latest ON ... WHERE acknowledgments.alert_key IN ($23..$44)`. |
| Composite `(alert_key, created_at)` index exists after migration | ✅ | `\d acknowledgments` on the AutoMigrate-created postgres schema: `"idx_acknowledgments_alert_key_created_at" btree (alert_key, created_at)` (alongside the pre-existing single-column index). |
| Empty cache makes no DB round-trip | ✅ | Stopped the webui to silence the 10 s poller, confirmed a quiet postgres log, then called `GetAllAcknowledgedAlerts{"alert_keys":[]}` over gRPC → `{}` and **0** `FROM "acknowledgments"` statements. Control call with one key → exactly **1** statement. WebUI side also logs `No alerts to load acknowledgments for` and skips the RPC when the cache is empty. |

**Exploratory:** Adjacent flows intact — the batched comment-count path still resolves (`commentCount:1` from the ack audit-trail comment), acknowledging still writes its audit comment, Classic/Acknowledged/Resolved tab filtering behaves as before, and registration/login/logout are unaffected. One quirk observed: on the Acknowledged tab, the client-side live refresh drops `isAcknowledged`/`acknowledgedBy` from already-rendered rows and lets un-acked alerts leak into the list (correct on server-rendered load, degrades after the first in-page refresh). I rebuilt and booted `main` to check, and it reproduces identically there (`ACK/qaalice` on load → `noack` after live refresh) — pre-existing client-side behaviour, untouched by this diff, not a regression from this PR.

**Verdict: PASS**

SoulKyu:

internal/backend/models/models.go
```diff
@@ -91,10 +91,10 @@ func (c *Comment) BeforeCreate(tx *gorm.DB) error {
 
 type Acknowledgment struct {
 	ID        string    `gorm:"primaryKey;type:varchar(32)" json:"id"`
-	AlertKey  string    `gorm:"not null;size:500;index" json:"alert_key"`
+	AlertKey  string    `gorm:"not null;size:500;index;index:idx_acknowledgments_alert_key_created_at,priority:1" json:"alert_key"`
```
SoulKyu, line 94: **Non-blocking: the standalone `index` on `AlertKey` is now redundant.**

Problem: this tag keeps the original single-column index (`idx_acknowledgments_alert_key`) alongside the new composite `idx_acknowledgments_alert_key_created_at`. Because `alert_key` is the composite's leading column, every acknowledgments query that filters on `alert_key` alone is served by a leftmost-prefix lookup on the composite — including `GetAcknowledgments`, `DeleteAcknowledgment`, and both `IN` filters in the new `GetAllAcknowledgedAlerts` (`internal/backend/database/gorm_db.go:490-498`).

Why it matters: the PR description itself notes the `acknowledgments` table is never pruned, so keeping both indexes means every ack insert maintains two B-trees and duplicates storage that grows with deployment history — pure write amplification with no read benefit, on the exact table this PR is optimizing.

Suggested change — drop the standalone `index` from the tag so fresh installs only create the composite:

```go
AlertKey  string    `gorm:"not null;size:500;index:idx_acknowledgments_alert_key_created_at,priority:1" json:"alert_key"`
```

Note: `AutoMigrate` won't drop the old index on already-migrated databases (that would need a one-off `DROP INDEX idx_acknowledgments_alert_key`), but removing the tag stops new deployments from creating it.


SoulKyu (commented):
Solid, tightly-scoped perf fix — I verified the `alert_key IN (...)` bound is applied to both the outer query and the latest-per-key subquery, the empty-key path short-circuits before any DB round-trip, `loadAcknowledgmentsEfficiently` collects fingerprints under `RLock` and calls gRPC with no lock held, and the new tests pass locally (`go build ./...` clean; the `internal/webui` vet failure is the pre-existing `router.go:52` issue, untouched here).

One non-blocking index-hygiene suggestion inline on `internal/backend/models/models.go` — safe to merge as is.



SoulKyu:
## 🧪 QA Report

Re-test after the rebase (`head=bd2168c`). Fresh `make test` stack (postgres + backend + webui + fake alertmanager, empty DB), real WebUI driven in headless Chromium (Playwright), backend hit directly over gRPC (`grpcurl`), postgres running with `log_statement='all'` to capture the hot-path SQL.

|---|---|---|
| A refresh with N cached alerts sends at most N keys and receives at most N acks | ✅ | Captured statements track the cache size cycle by cycle (`$1..$21` / `$22..$42` with 21 cached alerts, then 22/23/24/25 as the fake AM added alerts). With 5 ack rows over 4 distinct keys in the DB, the webui logged `Received 2 acknowledged alerts from backend` — exactly the keys then cached. |
| Acks for uncached alerts never cross the gRPC boundary | ✅ | Inserted `alert_key='qa-ghost-uncached-key'` straight into postgres; over 6+ consecutive 10 s cycles the webui kept reporting `Received 2` (never 3). After the alertmanager restart rotated fingerprints, all 3 stored keys became uncached and the webui reported `Received 0` while an explicit gRPC call for those same keys returned `count: 3` — the rows were live, the key filter excluded them. |
| Badge shown after ack | ✅ | Logged in as `qaalice` in a real browser, clicked the row's ✓ Acknowledge, submitted a reason. Acknowledged tab renders `LogErrorRateHigh / app-14.example.com` with an `ACK` badge; client state `isAcknowledged:true, acknowledgedBy:"qaalice", acknowledgeReason:"QA-66 acked by ALICE"`. Row left the Classic view as expected. |
| Un-ack clears the badge | ✅ | Un-acknowledged `service-12.example.com` from the alert-details modal → row deleted from `acknowledgments`, and a fresh login one cycle later no longer lists it under Acknowledged (backend `Received` count dropped accordingly). |
| Ack by another user visible after one cycle | ✅ | `qabob` acknowledged `backup-9.example.com` in a separate browser session; `qaalice`'s next session showed it with `acknowledgedBy:"qabob"`, `reason:"QA-66 acked by BOB"`. Latest-per-key also checked: a later ack on that same key flipped both the UI and the gRPC response to `qaalice` / `QA-66 LATEST wins ALICE`. |
| Hot-path SQL bounded by `alert_key IN (...)` on both outer query and `latest` subquery | ✅ | Live capture: `SELECT acknowledgments.*, users.username FROM "acknowledgments" JOIN users ... JOIN (SELECT alert_key, MAX(created_at) as max_created FROM "acknowledgments" WHERE alert_key IN ($1..$21) GROUP BY "alert_key") latest ON ... WHERE acknowledgments.alert_key IN ($22..$42)`. |
| Composite `(alert_key, created_at)` index exists after migration | ✅ | `\d acknowledgments` on the AutoMigrate-created schema: `"idx_acknowledgments_alert_key_created_at" btree (alert_key, created_at)`, alongside the pre-existing single-column index. |
| Empty cache makes no DB round-trip | ✅ | WebUI stopped to silence the poller (baseline: 0 ack statements). `GetAllAcknowledgedAlerts{"alert_keys":[]}` and `{}` over gRPC → `{}` with **0** `FROM "acknowledgments"` statements; control call with one key → exactly **1**. Same at the webui level: restarted with the alertmanager down so the cache stayed empty → 0 ack statements in postgres. |

**Exploratory:** Adjacent flows intact — Classic / Resolved / Acknowledged / Hidden tab switching, the search filter (`HighCPU` → 2 matching rows), the batched comment-count path (`commentCount:1` from the ack audit-trail comment), acknowledging still writing its audit comment, login/logout (`200`), no page errors. The Acknowledged tab still loses `isAcknowledged`/`acknowledgedBy` from already-rendered rows after the first in-page live refresh (correct on server-rendered load) — `/api/v1/alerts` returns raw Alertmanager-shaped alerts with no acknowledgment fields at all, an endpoint this diff does not touch, so it is pre-existing client-side behaviour and not a regression here (matches what the previous QA run reproduced on `main`).

**Verdict: PASS**

PR #70, opened 2026-07-24
fix(webui): alert cache hands out live alert pointers — return snapshots, add MutateAlert

### LINKED ISSUE #54: fix(webui): alert cache hands out live alert pointers — background refresh races with request handlers
> All line references below are against `e452236` (they shifted after `fe4e9fa` landed).

## Problem

`AlertCache` protects the *map* with `ac.mu` but hands out raw pointers to the alert structs, which
the background refresher keeps mutating after the reader has released the lock.

Readers escape the lock:

- `internal/webui/services/alert_cache.go:528-538` — `GetAllAlerts` copies the slice of pointers
  under `RLock` and returns those pointers
- `internal/webui/services/alert_cache.go:582-590` — `GetAlert` returns the cache-resident pointer
  and `RUnlock`s before the caller dereferences anything
- `internal/webui/handlers/dashboard_handlers.go:346-357` — `getStandardAlerts` reads
  `alert.IsAcknowledged` / `alert.IsResolved` with no lock held (same for `getAcknowledgedAlerts`,
  `:359-370`)
- `internal/webui/handlers/dashboard_handlers.go:372` onward — `applyDashboardFilters` reads
  `Labels`, `Status`, `Severity`, then the alerts are converted to response structs

Writers mutate the same structs concurrently:

- `internal/webui/services/alert_cache.go:359-372` — `updateExistingAlert` writes `Status`,
  `EndsAt`, `Duration`, `IsResolved`, `Annotations`
- `internal/webui/services/alert_cache.go:463-466` — `loadAcknowledgmentsEfficiently` writes
  `IsAcknowledged`, `AcknowledgedBy`, `AcknowledgedAt`, `AcknowledgeReason` (own goroutine, spawned
  at `alert_cache.go:445`)
- `internal/webui/services/alert_cache.go:503-519` — `loadCommentCountsEfficiently` writes
  `CommentCount` (own goroutine, spawned at `alert_cache.go:446`)
- `internal/webui/services/alert_cache.go:241-244` — the resolve pass writes `IsResolved`,
  `ResolvedAt`, `Status.State`, `EndsAt`

Taking `ac.mu` on the writer side does not help: the reader dropped the lock before dereferencing.
These are unsynchronized concurrent read/writes on the same memory — a data race the Go race
detector reports.

Same file, `alert_cache.go:269-271`: `refreshAlerts` logs `len(ac.alerts)` after `ac.mu.Unlock()`,
an unsynchronized map read — a candidate for `fatal error: concurrent map read and map write`, which
no `recover` catches.

The hazard is already recognized on one path — `alertCopy := *alert` (`alert_cache.go:258-259`)
snapshots the struct before handing it to a goroutine, with a comment saying exactly why. The reader
side never got the same fix.

**A shallow struct copy is sufficient.** `Labels` is never written after construction anywhere in
`internal/webui`, and `Annotations` is only ever replaced wholesale (`alert_cache.go:371`, `:425`).
No deep copy of the maps is needed.

## Impact

Failure scenario: a browser polls `/api/v1/dashboard/data` while the 10s refresh cycle runs.
`GetAllAlerts` returns pointers, then `loadAcknowledgmentsEfficiently` writes `IsAcknowledged = true`
and `AcknowledgedBy = "alice"` on the same struct while the handler is serializing it. The handler
can observe `IsAcknowledged` true together with an empty `AcknowledgedBy`, or a `Status.State` of
`resolved` next to a stale `EndsAt` — inconsistent rows served to users. Unsynchronized access to
the string fields (`Status.State`, `AcknowledgedBy`, `Summary`) can also tear the ptr/len pair and
yield garbage.

The `len(ac.alerts)` read at `:271` is the sharper edge: a concurrent map read/write is a runtime
`fatal error`, not a recoverable panic.

Because nothing in the existing suite exercises concurrency (`alert_cache_test.go` is
single-goroutine — no `go func`, no `WaitGroup`, no `t.Parallel`), `go test -race` passes today and
hides all of this.

## Proposed approach

**Read path — return snapshots.**

- `GetAllAlerts` and `GetAlert` copy each struct under the lock and return the copies instead of
  cache-resident pointers.
- Compute the count while the lock is held at `alert_cache.go:269-271` instead of reading
  `len(ac.alerts)` after `Unlock()`.
- `GetAlertByFingerprint` (`alert_cache.go:660-665`) wraps `GetAlert` and therefore also returns a
  copy. Every one of its callers is **read-only** today —
  `dashboard_handlers.go:1237` (alert details), `:1335` (cache status), `:1402` (add comment
  existence check), `:1451` (delete comment existence check), `:1993` (unsilence, reads
  `Status.SilencedBy` only), `sentry_handlers.go:46`, and `GetAlertColors`
  (`alert_cache.go:667-668`) — and must stay read-only. If a future caller needs to mutate, it uses
  the mutation API below.

**Write path — this is the part that breaks if you only do the above.** `processAlertAction`
(`dashboard_handlers.go:836`) takes the pointer from `GetAlert` and **writes through it into the
cache**, under a comment that says so:

- `dashboard_handlers.go:864-869` — acknowledge: `IsAcknowledged`, `AcknowledgedBy`,
  `AcknowledgedAt`, `CommentCount++`
- `dashboard_handlers.go:899-904` — unacknowledge: same fields reset, `CommentCount++`
- `dashboard_handlers.go:906-911` — resolve: `Status.State`, `EndsAt`, `IsResolved`, `ResolvedAt`.
  This case writes **nowhere else** — the comment reads "this is more of a UI state than backend"

Snapshot the reads without touching this and the resolve action becomes a silent no-op, while
acknowledge degrades to "works after the next 10s refresh". So the change must add a mutation API
and move those three cases onto it:

```go
// MutateAlert applies fn to the cached alert under ac.mu; returns false if absent.
func (ac *AlertCache) MutateAlert(fingerprint string, fn func(*webuimodels.DashboardAlert)) bool
```

The acknowledge case also passes `alert` to a statistics goroutine
(`dashboard_handlers.go:872-878`); that goroutine must receive a snapshot, not the cache pointer —
the same treatment `alert_cache.go:258` already applies.

## Acceptance criteria

- [ ] No `AlertCache` accessor returns a pointer that the background refresher continues to mutate:
      `GetAllAlerts`, `GetAlert` and `GetAlertByFingerprint` return copies.
- [ ] `processAlertAction` no longer writes through a returned pointer; the acknowledge,
      unacknowledge and resolve cases go through `MutateAlert` (or an equivalent method that holds
      `ac.mu` for the mutation).
- [ ] Acknowledge → unacknowledge → resolve via `POST /api/v1/dashboard/action` each change the
      cached alert observably, verified by an immediately following read, with no dependency on a
      refresh cycle. In particular resolve still sets `Status.State = "resolved"` and `IsResolved`.
- [ ] `len(ac.alerts)` is no longer read outside the lock in `refreshAlerts`.
- [ ] A new test runs refresh-style writes concurrently with `GetAllAlerts` / `GetAlert` reads,
      passes under `go test -race ./internal/webui/services/...`, and fails on the current code.
- [ ] `go test -race ./...` stays green.
- [ ] Dashboard JSON output is unchanged (same fields, same values).

SoulKyu:
## Problem

`AlertCache` protected the map with `ac.mu` but returned raw pointers to the cached alert structs. The 10s background refresh (`updateExistingAlert`, `loadAcknowledgmentsEfficiently`, `loadCommentCountsEfficiently`, the resolve pass) kept mutating those structs after readers had released the lock — unsynchronized concurrent read/writes that the race detector flags, serving torn alert rows (e.g. `IsAcknowledged=true` with empty `AcknowledgedBy`). `refreshAlerts` also read `len(ac.alerts)` after `Unlock()`, a candidate for an unrecoverable `fatal error: concurrent map read and map write`.

## Changes

**Read path — snapshots**
- `GetAllAlerts` and `GetAlert` copy each struct under the lock and return the copies; `GetAlertByFingerprint` inherits this. Shallow copy is sufficient: `Labels` is never written after construction and `Annotations` is only replaced wholesale.
- `refreshAlerts` computes the active-alert count before `Unlock()`.

**Write path — `MutateAlert`**
- New `MutateAlert(fingerprint, fn)` applies `fn` to the cached alert under `ac.mu`; returns `false` if absent.
- `processAlertAction` no longer writes through returned pointers: acknowledge, unacknowledge and resolve go through `MutateAlert`, so each is observable on an immediately following read (resolve still sets `Status.State = "resolved"` / `IsResolved` with no refresh-cycle dependency).
- Two additional write-through sites beyond the issue's list got the same treatment: `AddAlertComment` (`CommentCount`/`LastCommentAt`) and the silence-comment count bump in `processSilenceAction`.
- The acknowledge statistics goroutine receives the handler-local snapshot instead of cache memory.

Dashboard JSON output is unchanged — same fields, same values, only copies instead of shared memory.

## Validation

- New `TestAlertCache_ConcurrentRefreshAndReads` runs refresh-style writes concurrently with `GetAlert`/`GetAllAlerts` readers: verified it **fails with `DATA RACE` on the old accessors** and passes after the fix.
- New `TestAlertCache_MutateAlert` checks mutation visibility on immediate read and that writing through a returned snapshot does not touch the cache.
- `go test -race ./...` — 48 packages green.

Closes #54

### DIFF
--- a/internal/webui/handlers/dashboard_handlers.go
+++ b/internal/webui/handlers/dashboard_handlers.go
@@ -861,12 +861,19 @@ func processAlertAction(c *gin.Context, fingerprint, action, comment, userID str
 			}
 		}
 
-		// Update local cache
-		alert.IsAcknowledged = true
-		alert.AcknowledgedBy = userID
-		alert.AcknowledgedAt = time.Now()
-		// Always increment comment count since we add an acknowledgment comment
-		alert.CommentCount++
+		// Update local cache under the cache lock (GetAlert returns a snapshot).
+		// Mirror the change on the snapshot so the statistics goroutine below
+		// sees the acknowledged state without touching cache-resident memory.
+		ackAt := time.Now()
+		applyAck := func(a *webuimodels.DashboardAlert) {
+			a.IsAcknowledged = true
+			a.AcknowledgedBy = userID
+			a.AcknowledgedAt = ackAt
+			// Always increment comment count since we add an acknowledgment comment
+			a.CommentCount++
+		}
+		alertCache.MutateAlert(fingerprint, applyAck)
+		applyAck(alert)
 
 		// Capture acknowledgment statistics
 		if backendClient != nil && backendClient.IsConnected() {
@@ -897,19 +904,24 @@ func processAlertAction(c *gin.Context, fingerprint, action, comment, userID str
 			}
 		}
 
-		// Update local cache
-		alert.IsAcknowledged = false
-		alert.AcknowledgedBy = ""
-		alert.AcknowledgedAt = time.Time{}
-		// Increment comment count for unacknowledgment comment
-		alert.CommentCount++
+		// Update local cache under the cache lock (GetAlert returns a snapshot)
+		alertCache.MutateAlert(fingerprint, func(a *webuimodels.DashboardAlert) {
+			a.IsAcknowledged = false
+			a.AcknowledgedBy = ""
+			a.AcknowledgedAt = time.Time{}
+			// Increment comment count for unacknowledgment comment
+			a.CommentCount++
+		})
 
 	case "resolve":
 		// Mark alert as resolved (this is more of a UI state than backend)
-		alert.Status.State = "resolved"
-		alert.EndsAt = time.Now()
-		alert.IsResolved = true
-		alert.ResolvedAt = time.Now()
+		resolvedAt := time.Now()
+		alertCache.MutateAlert(fingerprint, func(a *webuimodels.DashboardAlert) {
+			a.Status.State = "resolved"
+			a.EndsAt = resolvedAt
+			a.IsResolved = true
+			a.ResolvedAt = resolvedAt
+		})
 
 		// Add a comment about resolution for audit trail
 		if backendClient != nil && backendClient.IsConnected() {
@@ -924,7 +936,7 @@ func processAlertAction(c *gin.Context, fingerprint, action, comment, userID str
 				fmt.Printf("Warning: failed to add resolution comment: %v\n", err)
 			} else {
 				// Increment comment count if comment was added successfully
-				alert.CommentCount++
+				alertCache.MutateAlert(fingerprint, func(a *webuimodels.DashboardAlert) { a.CommentCount++ })
 			}
 		}
 
@@ -1412,9 +1424,11 @@ func AddAlertComment(c *gin.Context) {
 		return
 	}
 
-	// Update comment count in alert cache
-	alert.CommentCount++
-	alert.LastCommentAt = time.Now()
+	// Update comment count in alert cache (the lookup above returned a snapshot)
+	alertCache.MutateAlert(fingerprint, func(a *webuimodels.DashboardAlert) {
+		a.CommentCount++
+		a.LastCommentAt = time.Now()
+	})
 
 	c.JSON(http.StatusOK, webuimodels.SuccessResponse(gin.H{
 		"message": "Comment added successfully",
@@ -1984,7 +1998,7 @@ func processSilenceAction(c *gin.Context, fingerprint, comment, userID string) e
 			fmt.Printf("Warning: failed to add silence comment: %v\n", err)
 		} else {
 			// Increment comment count in cache only if comment was added successfully
-			alert.CommentCount++
+			alertCache.MutateAlert(fingerprint, func(a *webuimodels.DashboardAlert) { a.CommentCount++ })
 		}
 	}
 
--- a/internal/webui/services/alert_cache.go
+++ b/internal/webui/services/alert_cache.go
@@ -266,9 +266,10 @@ func (ac *AlertCache) refreshAlerts() {
 		}
 	}
 
+	activeCount := len(ac.alerts)
 	ac.mu.Unlock()
 
-	log.Printf("Alert cache refresh complete: %d active alerts, %d newly resolved", len(ac.alerts), resolvedCount)
+	log.Printf("Alert cache refresh complete: %d active alerts, %d newly resolved", activeCount, resolvedCount)
 
 	ac.loadBackendData()
 
@@ -430,6 +431,22 @@ func (ac *AlertCache) UpdateAlert(alert *webuimodels.DashboardAlert) {
 	}
 }
 
+// MutateAlert applies fn to the cached alert under ac.mu; returns false if the
+// fingerprint is not in the live cache. This is the only supported way to write
+// to a cached alert — the read accessors return snapshots, so writes through
+// their return values never reach the cache.
+func (ac *AlertCache) MutateAlert(fingerprint string, fn func(*webuimodels.DashboardAlert)) bool {
+	ac.mu.Lock()
+	defer ac.mu.Unlock()
+
+	alert, exists := ac.alerts[fingerprint]
+	if !exists {
+		return false
+	}
+	fn(alert)
+	return true
+}
+
 func (ac *AlertCache) loadBackendData() {
 	log.Printf("loadBackendData called - checking backend connection...")
 	if ac.backendClient == nil {
@@ -544,9 +561,14 @@ func (ac *AlertCache) GetAllAlerts() []*webuimodels.DashboardAlert {
 	ac.mu.RLock()
 	defer ac.mu.RUnlock()
 
+	// Return snapshots, not cache-resident pointers: the background refresher
+	// keeps mutating the cached structs after this lock is released. A shallow
+	// copy is enough — Labels is never written after construction and
+	// Annotations is only ever replaced wholesale.
 	alerts := make([]*webuimodels.DashboardAlert, 0, len(ac.alerts))
 	for _, alert := range ac.alerts {
-		alerts = append(alerts, alert)
+		alertCopy := *alert
+		alerts = append(alerts, &alertCopy)
 	}
 
 	return alerts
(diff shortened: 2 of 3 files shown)


SoulKyu (commented):
@SoulKyu — the core fix is correct and complete against issue #54: `GetAlert`/`GetAllAlerts` now return shallow snapshots taken under the lock, all six handler write-through sites (ack, unack, resolve, resolution-comment, `AddAlertComment`, silence-comment) go through the new `MutateAlert`, and the `len(ac.alerts)`-after-`Unlock` map read is gone. I verified the shallow-copy claim independently — `Labels` is never written after construction, `Annotations` and the `Status` slices are only ever replaced wholesale (`updateExistingAlert`, `UpdateAlert`) — and ran the new tests plus `go test -race` on the package: green. The `applyAck` mirror onto the handler-local snapshot for the statistics goroutine is a nice touch.

One follow-up in the same bug class. It sits outside this PR's changed ranges (the lines are unchanged and the race pre-dates this PR), so it is non-blocking here — fine to land this PR as-is and fix in a follow-up issue.

**`internal/webui/services/alert_cache.go:206-218` — `refreshAlerts` still publishes cache-resident pointers on two paths.**

- **Problem:** at lines 207–209 a brand-new `dashAlert` is stored into `ac.alerts` and the *same pointer* is appended to `newAlertsForSSE`; after `ac.mu.Unlock()` it is handed to SSE subscribers via `notifySubscribers` (line 287). Similarly, the `runBounded` goroutine at lines 212–218 captures the cache-resident `dashAlert` for `CaptureAlertFired` and reads it with no lock held.
- **Why it matters:** `internal/webui/handlers/sse_handler.go:59` `json.Marshal`s the update off-lock, and the very same refresh cycle then spawns `loadAcknowledgmentsEfficiently`/`loadCommentCountsEfficiently`, which write `IsAcknowledged`/`AcknowledgedBy`/`CommentCount` on those exact structs under `ac.mu` — an unsynchronized read/write pair identical to the one this PR removes from the accessors. A concurrent acknowledge request going through the new `MutateAlert` races the SSE marshal the same way.
- **Evidence:** contrast with the resolve pass at `alert_cache.go:256-259`, which snapshots (`alertCopy := *alert`) before handing the struct to a goroutine, with a comment explaining exactly this hazard — the new-alert branch never got that treatment.
- **Suggested fix:** mirror this PR's approach at the publish points: `alertCopy := *dashAlert; newAlertsForSSE = append(newAlertsForSSE, &alertCopy)` and pass a copy to `CaptureAlertFired` as well.



SoulKyu:
## 🧪 QA Report

Env: `make test` (docker compose: postgres + fake alertmanager + backend + webui on :8081), plus two local `go build -race` runs of the same webui code (PR head on :8082, parent `043062d` on :8083) against the same dockerized backend/alertmanager, driven by an identical 45s load harness (6 reader loops on `/dashboard/data` + `/dashboard/alert/:fp`, 2 writer loops doing acknowledge/unacknowledge on 13–15 fingerprints, spanning ~5 refresh cycles).

|---|---|---|
| `GetAllAlerts` / `GetAlert` / `GetAlertByFingerprint` return copies, not pointers the refresher keeps mutating | ✅ | Race-instrumented PR head under the load harness: **0 handler-side DATA RACE** reports (only 2 pre-existing ones in `refreshAlerts`, see below). Same harness on the parent commit: **44 DATA RACE** reports, including `GetDashboardData` / `GetAlertDetails` / `convertToResponseAlerts` (`encoding/json` reading an alert) racing with `updateExistingAlert` / `processAlertAction` — exactly the torn-row scenario in #54. Harness proven to detect the bug; the fix removes it. |
| `processAlertAction` no longer writes through a returned pointer (ack / unack / resolve go through `MutateAlert`) | ✅ | If the writes still went through the snapshot they would be silent no-ops now that reads are copies. All three are observable: see next row. `POST /dashboard/alert/:fp/comments` and the silence-comment bump were moved too — comment count went 3 → 4 on the immediately following `GET /dashboard/alert/:fp`. |
| ack → unack → resolve via the action endpoint each change the cached alert observably on an immediately following read, no refresh-cycle dependency | ✅ | `POST /api/v1/dashboard/bulk-action` on `f914f0868c600a7d388eb82615815bd6`, each followed instantly by `GET /api/v1/dashboard/alert/<fp>`:<br>acknowledge → `{"ack":true,"by":"09pyej2hld...","at":"2026-07-24T23:35:59.19Z","comments":1}`<br>unacknowledge → `{"ack":false,"by":null,"comments":2}`<br>resolve → `{"state":"resolved","resolved":true,"endsAt":"…23:35:59.263Z","resolvedAt":"…23:35:59.263Z"}` |
| `len(ac.alerts)` no longer read outside the lock in `refreshAlerts` | ✅ | Count captured before `Unlock()` (`alert_cache.go:269`). 19 refresh cycles logged under the concurrent load on the race build: no race report on that line and no `fatal error: concurrent map read and map write`. |

**Exploratory:** Browser pass (Playwright, dockerized :8081) — logged in, acknowledged `ServiceSlowResponse` from the row action + modal: it left Classic immediately (11 → 10 rows) and appears in the Acknowledged tab with the `✓ ACK` badge and comment count 1, without waiting for a refresh cycle; search filter still narrows correctly (11 → 4 rows on `cpu`); no new console/network errors (the two 401s on `/profile/timezone` and `/impersonate/status` reproduce identically on the parent commit).

**Note (out of scope, pre-existing — not introduced by this PR):** the race build still reports 2 DATA RACEs of the same family, untouched by the diff: `refreshAlerts` hands the cache-resident `dashAlert` pointer to the `CaptureAlertFired` goroutine (`alert_cache.go:212-214`) while the next cycle's `updateExistingAlert` (`:367`, `:372`) writes that struct. The sibling resolve path already snapshots (`alertCopy := *alert`, `:258`); this new-alert path does not. Worth a follow-up issue.

**Verdict: PASS**

PR #71, opened 2026-07-24
fix(webui): move color cache RPC out of the process-wide lock

### LINKED ISSUE #53: fix(webui): color cache holds the process-wide lock across a backend RPC — one cold lookup stalls every dashboard request
## Problem

`ColorService.getUserColorCache` takes the process-wide write lock and holds it across a backend
gRPC call:

- `internal/webui/services/color_service.go:119-120` — `cs.cacheMutex.Lock()` with a deferred
  `Unlock()`
- `internal/webui/services/color_service.go:130` — `cs.backendClient.GetUserColorPreferences(sessionID)`
  is called **while that lock is held**
- `internal/webui/client/backend_client.go:558-563` — that RPC uses a 10s context timeout

There is one `ColorService` for the whole process (`internal/webui/router.go:101`,
`alert_cache.go:102`) and the mutex is a single `sync.RWMutex` (`color_service.go:19`), not per user.
Go's RWMutex gives writers preference, so once a `Lock()` is pending, new `RLock()` callers park at
`color_service.go:111` — users whose entry is fresh wait behind a stranger's cold lookup.

This sits on the main dashboard path, not on a side endpoint:

- `internal/webui/handlers/dashboard_handlers.go:211` — `GetDashboardData` embeds colors via
  `computeAlertColorsMap`
- `internal/webui/handlers/dashboard_handlers.go:1702` — `computeAlertColorsMap` calls
  `GetAlertColorsOptimized`, which calls `getUserColorCache` (`color_service.go:81`)
- `internal/webui/handlers/dashboard_handlers.go:1763` — same for `GET /dashboard/alert-colors`,
  which the browser refetches after SSE updates

Two aggravating factors:

1. The cache TTL is 5 minutes (`color_service.go:53`), so every user re-enters the slow path
   regularly.
2. Failures are never cached — `getUserColorCache` returns `nil, err` at `color_service.go:131-133`
   without storing anything — so each subsequent request pays the full 10s timeout again.

Colour *degradation* already exists: on error, `GetAlertColors` (`color_service.go:64-67`) and
`GetAlertColorsOptimized` (`:78-88`) fall back to `getDefaultSeverityColors`. The problem is not
that a backend outage loses colours; it is that the process pays for the same failure on every
single request, holding a global lock while it does.

## Impact

Backend slow or unreachable: one request whose colour entry has expired holds the global write lock
for up to 10s. Every other dashboard request queues behind it, including users with a perfectly
fresh cache. With no negative caching the next request repeats the stall, so a backend hiccup turns
into a WebUI-wide freeze and a gin worker pileup instead of a degraded-colours experience.

Even in the healthy case, every TTL expiry serializes all colour computation in the process behind
one backend round-trip.

`HiddenAlertsService.LoadUserData` already got exactly this treatment — see the C4 fix comment at
`internal/webui/services/hidden_alerts_service.go:59-61` ("perform gRPC calls BEFORE acquiring the
write lock"), followed by an implementation that does the RPCs, compiles the regexes, and only then
takes `s.mu`. `ColorService` was never updated.

## Proposed approach

1. **Move the RPC out of the critical section.** Read under `RLock`; on miss or expiry call the
   backend with **no lock held**; take the write lock only to store the result, re-checking whether
   a concurrent caller already stored one. Same shape as `hidden_alerts_service.go:54-95`.
   The existing opportunistic `sweepExpiredLocked()` call (`color_service.go:128`) stays under the
   write lock.

2. **Negative caching, 15 seconds.** On RPC error, store a failure marker in the same
   `colorCache` entry with a 15s TTL and return the error immediately for subsequent callers within
   that window (they already degrade to `getDefaultSeverityColors`). Chosen over serve-stale: it is
   a single TTL field on an existing struct, it needs no "how stale is too stale" policy, and the
   colour fallback path already exists. `InvalidateUserCache` (`color_service.go:104-108`) deletes
   the map entry and therefore clears a cached failure too — the user-triggered "I changed my colour
   preferences" path is never blocked by it.

3. **Introduce the seam the test needs — declared scope, not a surprise.** `ColorService.backendClient`
   is a concrete `*client.BackendClient` (`color_service.go:17`, `backend_client.go:18`), so there is
   nothing to stub today. Add a minimal interface in `internal/webui/services`:

   ```go
   type colorPreferencesFetcher interface {
       GetUserColorPreferences(sessionID string, impersonateUserID ...string) ([]*alertpb.UserColorPreference, error)
   }
   ```

   `*client.BackendClient` satisfies it implicitly; `NewColorService` (`color_service.go:49`) keeps
   its `*client.BackendClient` parameter so the three construction sites
   (`router.go:101`, `alert_cache.go:102`, `session_cache_cleanup_test.go:12` — the last passes
   `nil`) are unchanged. Only the struct field type changes. No other type is restructured.

4. **Singleflight is explicitly out of scope.** Collapsing N concurrent misses for the same session
   into one RPC is a nice-to-have; with the lock released, N misses simply run in parallel, which is
   the behaviour this issue asks for. Do not add it here.

## Acceptance criteria

- [ ] `getUserColorCache` never holds `cacheMutex` while calling the backend: in the final code, the
      RPC call sits between an `RUnlock()`/`Unlock()` and the next `Lock()`.
- [ ] Package-level test in `internal/webui/services` with a stub `colorPreferencesFetcher` that
      blocks for 200ms: N=10 concurrent `getUserColorCache` calls for **distinct** sessions complete
      in roughly one block duration, not N.
- [ ] Package-level test: a session whose entry is fresh returns without blocking while another
      session's fetch is in flight (the stub signals when it is inside the fetch).
- [ ] Package-level test with a call-counting stub that always errors: 5 sequential
      `getUserColorCache` calls within 15s produce exactly **one** backend call; a call after the 15s
      window produces a second.
- [ ] `InvalidateUserCache` clears a cached failure: after an error is cached, calling it makes the
      next `getUserColorCache` issue a fresh RPC.
- [ ] Existing colour behaviour is unchanged: on error, callers still receive
      `getDefaultSeverityColors` output.
- [ ] `go test -race ./internal/webui/services/...` is green.

SoulKyu:
## Problem

`ColorService.getUserColorCache` took the process-wide write lock and held it across the backend `GetUserColorPreferences` gRPC call (10s timeout). With one `ColorService` per process and writer-preferring `RWMutex`, a single cold or expired lookup parked every dashboard request — including users with a fresh cache. Failures were never cached, so a backend outage meant every request re-paid the full timeout while holding the global lock.

## Changes

- **RPC moved out of the critical section** (`internal/webui/services/color_service.go`): read under `RLock`; on miss/expiry the backend call runs with no lock held; the write lock is only taken to store the result, re-checking for a concurrent store first (same shape as the earlier `HiddenAlertsService.LoadUserData` fix). Concurrent misses for distinct sessions now fetch in parallel.
- **Negative caching, 15s**: a failed fetch stores a failure marker (`FetchErr` + 15s TTL) in the existing cache entry, so subsequent callers within the window get the cached error immediately and degrade to default severity colors instead of stalling. `InvalidateUserCache` clears the marker.
- **Test seam**: `ColorService.backendClient` becomes a package-private `colorPreferencesFetcher` interface, satisfied implicitly by `*client.BackendClient`. `NewColorService` keeps its concrete parameter; all construction sites unchanged.

Singleflight for same-session concurrent misses is intentionally out of scope per the issue.

## Validation

- New package tests: parallel misses complete in ~one block duration (not N), fresh entries aren't blocked by an in-flight cold fetch, error path issues exactly one RPC per 15s window, `InvalidateUserCache` clears a cached failure, error callers still get `getDefaultSeverityColors` output.
- `go test -race ./internal/webui/services/...` — 44 tests green.
- `go build ./...` — clean.

Closes #53

### DIFF
--- a/internal/webui/services/color_service.go
+++ b/internal/webui/services/color_service.go
@@ -8,13 +8,25 @@ import (
 	"sync"
 	"time"
 
+	alertpb "notificator/internal/backend/proto/alert"
 	"notificator/internal/models"
 	"notificator/internal/webui/client"
 	webuimodels "notificator/internal/webui/models"
 )
 
+// colorPreferencesFetcher is the slice of *client.BackendClient the color
+// cache needs; it exists so tests can stub the backend RPC.
+type colorPreferencesFetcher interface {
+	GetUserColorPreferences(sessionID string, impersonateUserID ...string) ([]*alertpb.UserColorPreference, error)
+}
+
+// colorFetchErrTTL is the negative-cache window: after a failed backend
+// fetch, callers get the cached error (and degrade to severity colors)
+// instead of paying the RPC timeout on every request.
+const colorFetchErrTTL = 15 * time.Second
+
 type ColorService struct {
-	backendClient *client.BackendClient
+	backendClient colorPreferencesFetcher
 	colorCache    map[string]*ColorPreferenceCache // userID -> cache
 	cacheMutex    sync.RWMutex
 	defaultColors map[string]string
@@ -27,6 +39,7 @@ type ColorPreferenceCache struct {
 	LookupMap   map[string]*ColorMatch            `json:"-"` // Pre-computed lookups for performance
 	CachedAt    time.Time                         `json:"cachedAt"`
 	TTL         time.Duration                     `json:"ttl"`
+	FetchErr    error                             `json:"-"` // non-nil marks a negative-cache entry
 }
 
 type ColorMatch struct {
@@ -113,23 +126,44 @@ func (cs *ColorService) getUserColorCache(sessionID string) (*ColorPreferenceCac
 	cs.cacheMutex.RUnlock()
 
 	if exists && time.Since(cache.CachedAt) < cache.TTL {
+		if cache.FetchErr != nil {
+			return nil, cache.FetchErr
+		}
 		return cache, nil
 	}
+
+	// Fetch with NO lock held (issue #53): holding the write lock across this
+	// RPC (10s timeout) would stall every dashboard request in the process.
+	// Concurrent misses for the same session fetch in parallel; last store wins.
+	preferences, fetchErr := cs.backendClient.GetUserColorPreferences(sessionID)
 
 	cs.cacheMutex.Lock()
 	defer cs.cacheMutex.Unlock()
 
-	if cache, exists := cs.colorCache[sessionID]; exists && time.Since(cache.CachedAt) < cache.TTL {
-		return cache, nil
+	// Re-check: a concurrent caller may have stored a result while we fetched.
+	if existing, ok := cs.colorCache[sessionID]; ok && time.Since(existing.CachedAt) < existing.TTL {
+		if existing.FetchErr == nil {
+			return existing, nil
+		}
+		if fetchErr != nil {
+			return nil, existing.FetchErr
+		}
+		// Our fetch succeeded; overwrite the failure marker below.
 	}
 
 	// ponytail: opportunistic sweep instead of a janitor goroutine — expired
 	// entries for idle sessions are dropped whenever any entry is refreshed.
 	cs.sweepExpiredLocked()
 
-	preferences, err := cs.backendClient.GetUserColorPreferences(sessionID)
-	if err != nil {
-		return nil, fmt.Errorf("failed to get user color preferences: %w", err)
+	if fetchErr != nil {
+		err := fmt.Errorf("failed to get user color preferences: %w", fetchErr)
+		cs.colorCache[sessionID] = &ColorPreferenceCache{
+			UserID:   sessionID,
+			CachedAt: time.Now(),
+			TTL:      colorFetchErrTTL,
+			FetchErr: err,
+		}
+		return nil, err
 	}
 
 	var webuiPrefs []webuimodels.UserColorPreference
--- a/internal/webui/services/color_service_test.go
+++ b/internal/webui/services/color_service_test.go
@@ -1,0 +1,168 @@
+package services
+
+import (
+	"errors"
+	"fmt"
+	"sync"
+	"testing"
+	"time"
+
+	alertpb "notificator/internal/backend/proto/alert"
+	"notificator/internal/models"
+)
+
+type stubColorFetcher struct {
+	mu      sync.Mutex
+	calls   int
+	delay   time.Duration
+	err     error
+	inFetch chan string   // if set, receives sessionID when a fetch starts
+	release chan struct{} // if set, fetch blocks until closed
+}
+
+func (s *stubColorFetcher) GetUserColorPreferences(sessionID string, _ ...string) ([]*alertpb.UserColorPreference, error) {
+	s.mu.Lock()
+	s.calls++
+	s.mu.Unlock()
+	if s.inFetch != nil {
+		s.inFetch <- sessionID
+	}
+	if s.release != nil {
+		<-s.release
+	}
+	if s.delay > 0 {
+		time.Sleep(s.delay)
+	}
+	return nil, s.err
+}
+
+func (s *stubColorFetcher) callCount() int {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	return s.calls
+}
+
+func newTestColorService(fetcher colorPreferencesFetcher) *ColorService {
+	cs := NewColorService(nil)
+	cs.backendClient = fetcher
+	return cs
+}
+
+// Distinct cold sessions must fetch in parallel, not serialize behind the
+// process-wide lock: 10 concurrent 200ms fetches ≈ 200ms, not 2s.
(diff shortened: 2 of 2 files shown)


SoulKyu (commented):
@SoulKyu Nice fix — this closes out issue #53 cleanly. I verified the new lock discipline in `internal/webui/services/color_service.go`: the read path holds only `RLock`, the backend RPC runs with no lock held, and the write-lock re-check correctly handles all four combinations of local-fetch vs. concurrent-store success/failure (including overwriting a stale failure marker when the local fetch succeeded). Cache entries are immutable after insertion, so the unlocked `CachedAt`/`FetchErr` reads are race-free. The 15s negative cache stores the wrapped error without extending the window on repeat hits, and `InvalidateUserCache` clears it. I ran `go test -race -count=1 ./internal/webui/services/` (44 tests green) and `go build ./...` locally, and confirmed all `NewColorService` call sites still pass the concrete client, so the `colorPreferencesFetcher` seam is backward compatible. The test coverage is genuinely good — parallel misses, fresh-entry non-blocking, negative-cache windowing, invalidation, and severity fallback are all exercised with sane timing margins. Solid, well-scoped work.



SoulKyu:
## 🧪 QA Report

Env: full `make test` rebuild (`docker compose` — postgres + fake alertmanager + backend + webui on :8081). Runtime fault injection: `LOCK TABLE user_color_preferences IN ACCESS EXCLUSIVE MODE` in postgres, which stalls the backend's `GetUserColorPreferences` query while `ValidateSession` (different table) keeps working — i.e. exactly the "backend slow on the colour path" scenario from #53, exercised through real HTTP requests on the running app.

|---|---|---|
| `getUserColorCache` never holds `cacheMutex` across the backend RPC | ✅ | `color_service.go`: `RUnlock()` at :126 → RPC at :138 → `Lock()` at :140. Runtime proof: with the colour query stalled, 10 concurrent `GET /dashboard/alert-colors` (10 distinct logged-in sessions, all cold) each returned in **10.01–10.02s** — i.e. all ran in parallel against the 10s RPC timeout, not 10×10s serialized. |
| Package test: 10 concurrent misses for distinct sessions ≈ one block duration | ✅ | `TestColorCacheConcurrentMissesRunInParallel` passes with `-race`. Ran the same test against the pre-PR `color_service.go` (scratch copy, test seam only): **FAIL — "10 concurrent misses took 2.004s; fetches are serialized"**, so the test genuinely detects the regression. |
| Package test: fresh entry not blocked by another session's in-flight fetch | ✅ | `TestColorCacheFreshEntryNotBlockedByColdFetch` passes; FAILs on pre-PR code ("fresh session blocked behind another session's fetch"). Runtime: during the 10s stall, a warm session's `GET /dashboard/alert-colors` returned in **9.9ms**, and a full dashboard reload for that session took **168ms** while another user's colour fetch was stuck. |
| Negative caching: 5 calls in 15s → exactly 1 backend call, 1 more after the window | ✅ | `TestColorCacheNegativeCaching` passes; FAILs on pre-PR code ("got 5"). Runtime: cold session with the colour query blocked → req1 **10.007s**, req2/3/4 within the window **2.3ms / 1.9ms / 1.9ms**, all HTTP 200 with severity colours. |
| `InvalidateUserCache` clears a cached failure | ✅ | `TestInvalidateUserCacheClearsNegativeEntry` passes; FAILs on pre-PR code. Runtime equivalent: `POST /dashboard/color-preferences` (which calls `InvalidateUserCache`) made the new `#ff00ff` critical colour appear on the very next request instead of waiting out the TTL. |
| Existing colour behaviour unchanged — errors degrade to `getDefaultSeverityColors` | ✅ | `TestGetAlertColorsFallsBackOnError` passes. Runtime: during the stall every alert came back `"colorSource":"severity"` (`#dc2626` critical / `#d97706` warning), HTTP 200; a browser session logging in during the stall still rendered the full dashboard (11.6s, one timeout) with default severity colours — screenshot confirms coloured rows, no error page. After the 15s window with the lock released, the same session got `"colorSource":"user"` + `#ff00ff` back — no stale poisoning. |
| `go test -race ./internal/webui/services/...` green | ✅ | 44 tests pass, `-count=1 -race`, on a pristine checkout of 6cbc7f5. `go build ./...` clean. |

**Exploratory:** adjacent colour flows all still work end-to-end — dashboard page renders user colours in a real headless Chromium (`background-color: #ff00ff19; border-left-color: #ff00ff`), `GET /dashboard/color-preferences`, `POST /dashboard/alerts/bulk-colors` (same cache, returned `"colorSource":"user"`), SSE stream and login/auth unaffected. Two pre-existing quirks unrelated to this diff, noted only for the record: `DELETE /dashboard/color-preferences/:id` never calls `InvalidateUserCache`, so a deleted preference stays visible until the 5min TTL (handler untouched by this PR); and the login page fires two `401`s on `/api/v1/profile/timezone` + `/api/impersonate/status` before authentication.

**Verdict: PASS**

PR #82, opened 2026-07-25
perf(webui): serve hidden alerts from cache within a 30s TTL

### LINKED ISSUE #78: perf(webui): every dashboard request re-fetches hidden alerts and rules over gRPC — LoadUserData never consults its own cache
## Problem

`HiddenAlertsService.LoadUserData` is written as a cache-fill function, but it has no freshness check: every call unconditionally performs two blocking gRPC round-trips to the backend and recompiles every regex rule.

`internal/webui/services/hidden_alerts_service.go:54-124`:

```go
func (s *HiddenAlertsService) LoadUserData(sessionID string) error {
	hiddenAlerts, hiddenAlertsErr := s.GetUserHiddenAlerts(sessionID)   // gRPC round-trip
	rules, hiddenRulesErr := s.GetUserHiddenRules(sessionID)            // gRPC round-trip
	// recompile every regex rule
	// s.mu.Lock() and overwrite the maps
	s.lastAccess[sessionID] = time.Now()
```

`s.lastAccess` is written here but only ever read by `sweepIdleSessionsLocked` (`:128`) for the 7-day idle eviction — it is never used to decide whether the fetch can be skipped. There is no TTL field. So the cache is populated and immediately invalidated by the next request.

Both hot dashboard endpoints call it on **every** request, unconditionally:

- `internal/webui/handlers/dashboard_handlers.go:132-134` (`GetDashboardData`, `GET /api/v1/dashboard/data`)
- `internal/webui/handlers/dashboard_handlers.go:1141-1144` (`GetDashboardIncremental`, `GET /api/v1/dashboard/incremental`)

Each of those gRPC calls costs two backend DB queries, not one, because the backend has no auth interceptor and re-validates the session by hand:

- `internal/backend/services/services.go:1693` — `GetUserBySession(req.SessionId)`, then `:1708` — `GetUserHiddenAlerts(targetUserID)` (`internal/backend/database/gorm_db.go:694-705`)
- `internal/backend/services/services.go:1880-1893` — same shape for `GetUserHiddenRules`

**Total per dashboard request: 2 sequential gRPC calls + 4 DB queries + N regex compiles, before a single alert is looked at.**

There is a second, subtler cost. `LoadUserData` takes `s.mu.Lock()` (`:91`) while the request-serving path takes `s.mu.RLock()` **twice per alert** inside the filter loop — `applyDashboardFilters` → `IsAlertHidden` (`dashboard_handlers.go:384` → `hidden_alerts_service.go:142-152`). Go's `sync.RWMutex` blocks new readers once a writer is queued, so a concurrent request's `LoadUserData` write-lock stalls another request's filter loop mid-iteration. The more tabs are open, the worse it gets.

## Impact

Concrete scenario: SSE fails or is unavailable, so the client falls back to polling `loadDashboardIncremental()` every `currentInterval` — 5000 ms by default (`internal/webui/templates/scripts/dashboard_core.templ:124`, `:516-521`). With 20 open tabs that is **8 gRPC calls/s and 16 DB queries/s** dedicated entirely to re-reading a hidden-alerts list that only changes when someone clicks "hide".

Even on the SSE path, every filter change, sort change, page change and debounced search keystroke hits `/dashboard/data` and pays the same two round-trips, adding backend RTT to the perceived latency of every single UI interaction.

This is pure waste: the data is per-user, tiny, mutated only through `HideAlert` / `UnhideAlert` / `SaveHiddenRule` / `RemoveHiddenRule` — all of which already update or invalidate the cache in-process (`:277`, `:293`, `:353`, `:366`, `:372-380`).

## Proposed approach

Make `LoadUserData` behave like the cache it already almost is — the mutation paths already keep it coherent, so a short TTL is safe:

1. Add a TTL check at the top of `LoadUserData`: under `RLock`, if `time.Since(s.lastAccess[sessionID])` is below a TTL (e.g. 30 s, matching the pattern already used by `ColorService.cacheTTL` in `internal/webui/services/color_service.go:56`) **and** `s.userHiddenAlerts[sessionID] != nil`, return immediately without any gRPC call.
2. Keep the existing write path unchanged (gRPC outside the lock, then `Lock()` to publish) for the cold/expired case, and keep `sweepIdleSessionsLocked` as-is.
3. Add an explicit `ForceLoadUserData` (or a `force bool`) if any caller genuinely needs a bypass; the mutation paths do not — they already call `InvalidateCache`.
4. Optional follow-up, separate concern: `IsAlertHidden` takes `RLock`/`RUnlock` twice per alert. Hoisting one snapshot of the per-session maps out of `applyDashboardFilters`' loop would remove ~2×N lock operations per request.

## Acceptance criteria

- [ ] Two consecutive `GET /api/v1/dashboard/data` requests within the TTL for the same session produce exactly **one** pair of `GetUserHiddenAlerts` / `GetUserHiddenRules` gRPC calls, not two. Verifiable by counting backend log lines or with a fake `BackendClient` counting invocations in a unit test.
- [ ] Hiding an alert via `POST /api/v1/dashboard/hide-alert` is reflected in the very next `/dashboard/data` response with no TTL wait (mutation paths still update/invalidate the cache synchronously).
- [ ] Deleting or disabling a hidden rule is reflected on the next request (existing `InvalidateCache` path still exercised).
- [ ] A first request for an unknown session still loads the data (cold path unchanged).
- [ ] Regexes are compiled once per load, not once per request.
- [ ] `go build -tags "nogui,webui" ./...` and `go vet ./...` clean; existing hidden-alerts tests still pass.

SoulKyu:
## Problem

`HiddenAlertsService.LoadUserData` was a cache-fill function with no freshness check: every call did two blocking gRPC round-trips (4 backend DB queries, since the backend re-validates the session per call) and recompiled every regex rule. Both hot dashboard endpoints (`GET /api/v1/dashboard/data`, `GET /api/v1/dashboard/incremental`) call it unconditionally, so the cache was populated and immediately invalidated by the next request. `s.lastAccess` was already written but only read by the 7-day idle sweep.

## Change

- Add `cacheTTL` (30s, mirroring `ColorService.cacheTTL`) and a check at the top of `LoadUserData`: under `RLock`, if the session has a populated snapshot younger than the TTL, return with no gRPC call.
- Cold, expired and invalidated sessions take the existing fetch-outside-the-lock / publish-under-`Lock` path unchanged; `sweepIdleSessionsLocked` unchanged.
- No `ForceLoadUserData`: the mutation paths (`HideAlert`, `UnhideAlert`, `SaveHiddenRule`, `RemoveHiddenRule`) already update or `InvalidateCache` the entry synchronously, so hides and rule changes are still reflected on the very next request with no TTL wait.

The write-lock contention noted in the issue drops out as a side effect — the write lock is now taken once per 30s per session instead of once per request, so it no longer stalls concurrent filter loops mid-iteration.

## Tests

`internal/webui/services/session_cache_cleanup_test.go`:
- `TestHiddenAlertsServiceLoadUserDataServesCacheWithinTTL` — fresh entry, nil backend client, no panic ⇒ no fetch.
- `TestHiddenAlertsServiceLoadUserDataRefetchesWhenStale` — expired / cold / invalidated all still attempt the backend fetch.

`go build -tags "nogui,webui" ./...` clean, `go test ./internal/webui/services/` passes (51 tests). The one `go vet` finding (`internal/webui/router.go:52`, `log.Fatalf` with `%w`) is pre-existing and untouched here.

Not done (optional follow-up from the issue, separate concern): hoisting the per-session map snapshot out of `applyDashboardFilters`' loop to drop the 2×N `RLock` pairs in `IsAlertHidden`.

Closes #78

### DIFF
--- a/internal/webui/services/hidden_alerts_service.go
+++ b/internal/webui/services/hidden_alerts_service.go
@@ -310,13 +310,23 @@ func (s *HiddenAlertsService) CompileFilterRules(rules []webuimodels.FilterHidde
 	return compiledRules
 }
 
+// isImpersonating mirrors the check client.BackendClient uses to decide whether
+// a mutation targets another user. The cache is keyed by sessionID and holds the
+// session owner's hidden set, so an impersonated mutation says nothing about it.
+func isImpersonating(impersonateUserID []string) bool {
+	return len(impersonateUserID) > 0 && impersonateUserID[0] != ""
+}
+
 // HideAlert hides a specific alert for a user
 func (s *HiddenAlertsService) HideAlert(sessionID string, alert *webuimodels.DashboardAlert, reason string, impersonateUserID ...string) error {
 	err := s.backendClient.HideAlert(sessionID, alert.Fingerprint, alert.AlertName, alert.Instance, reason, impersonateUserID...)
 	if err != nil {
 		return fmt.Errorf("failed to hide alert in backend: %w", err)
 	}
-	
+	if isImpersonating(impersonateUserID) {
+		return nil
+	}
+
 	// Update the cache
 	s.mu.Lock()
 	defer s.mu.Unlock()
@@ -335,7 +345,10 @@ func (s *HiddenAlertsService) UnhideAlert(sessionID, fingerprint string, imperso
 	if err != nil {
 		return fmt.Errorf("failed to unhide alert in backend: %w", err)
 	}
-	
+	if isImpersonating(impersonateUserID) {
+		return nil
+	}
+
 	// Update the cache
 	s.mu.Lock()
 	defer s.mu.Unlock()
@@ -427,7 +440,10 @@ func (s *HiddenAlertsService) InvalidateCache(sessionID string) {
 	delete(s.userHiddenAlerts, sessionID)
 	delete(s.userHiddenRules, sessionID)
 	delete(s.compiledRegexRules, sessionID)
-	delete(s.lastAccess, sessionID)
+	// Backdate instead of deleting: the entry can never read as fresh (the
+	// snapshot is gone and LoadUserData checks that first) but stays visible to
+	// sweepIdleSessionsLocked, which is what reclaims the generation counter.
+	s.lastAccess[sessionID] = time.Now().Add(-s.cacheTTL)
 	s.generation[sessionID]++
 }
 
@@ -465,14 +481,18 @@ func (s *HiddenAlertsService) ClearAllHiddenAlerts(sessionID string, impersonate
 	if err != nil {
 		return fmt.Errorf("failed to clear hidden alerts in backend: %w", err)
 	}
-	
+	if isImpersonating(impersonateUserID) {
+		return nil
+	}
+
 	// Clear the cache
 	s.mu.Lock()
 	defer s.mu.Unlock()
 	if s.userHiddenAlerts[sessionID] != nil {
 		s.userHiddenAlerts[sessionID] = make(map[string]bool)
 	}
-	
+	s.generation[sessionID]++
+
 	return nil
 }
 
--- a/internal/webui/services/hidden_alerts_service.go
+++ b/internal/webui/services/hidden_alerts_service.go
@@ -17,6 +17,11 @@ import (
 // idle longer than this are unreachable (session expired) and can be dropped.
 const sessionIdleTTL = 7 * 24 * time.Hour
 
+// cacheTTL bounds how long a cached hidden-alerts snapshot is served without
+// re-fetching. Mutations (hide/unhide/save/remove rule) update or invalidate the
+// cache in-process, so this only covers changes made by another webui replica.
+const cacheTTL = 30 * time.Second
+
 // HiddenAlertsService manages hidden alerts and rules for users
 type HiddenAlertsService struct {
 	backendClient       *client.BackendClient
@@ -24,7 +29,8 @@ type HiddenAlertsService struct {
 	userHiddenAlerts    map[string]map[string]bool             // userID -> fingerprint -> hidden
 	userHiddenRules     map[string][]models.UserHiddenRule     // userID -> rules
 	compiledRegexRules  map[string]map[string]*regexp.Regexp   // userID -> ruleID -> compiled regex
-	lastAccess          map[string]time.Time                   // userID -> last LoadUserData call
+	lastAccess          map[string]time.Time                   // userID -> last LoadUserData fetch
+	cacheTTL            time.Duration
 }
 
 // NewHiddenAlertsService creates a new hidden alerts service
@@ -35,6 +41,7 @@ func NewHiddenAlertsService(backendClient *client.BackendClient) *HiddenAlertsSe
 		userHiddenRules:    make(map[string][]models.UserHiddenRule),
 		compiledRegexRules: make(map[string]map[string]*regexp.Regexp),
 		lastAccess:         make(map[string]time.Time),
+		cacheTTL:           cacheTTL,
 	}
 	
 	// Load initial data
@@ -50,11 +57,22 @@ func (s *HiddenAlertsService) LoadAllUserData() {
 	log.Println("HiddenAlertsService initialized")
 }
 
-// LoadUserData loads hidden alerts and rules for a specific user using sessionID
+// LoadUserData loads hidden alerts and rules for a specific user using sessionID,
+// serving the cached snapshot when it is younger than cacheTTL.
 func (s *HiddenAlertsService) LoadUserData(sessionID string) error {
 	// Get userID from session for cache key
 	// Note: We'll need to pass userID separately or get it from session
 	// For now, we'll use sessionID as the cache key
+
+	s.mu.RLock()
+	fresh := s.userHiddenAlerts[sessionID] != nil && time.Since(s.lastAccess[sessionID]) < s.cacheTTL
+	s.mu.RUnlock()
+	if fresh {
+		return nil
+	}
+
+	// ponytail: no single-flight — concurrent misses for the same session just
+	// fetch twice and publish the same data; add one if that shows up in traces.
 
 	// C4 fix: perform gRPC calls BEFORE acquiring the write lock to avoid
 	// holding the lock across potentially long-running I/O operations.
--- a/internal/webui/services/hidden_alerts_service.go
+++ b/internal/webui/services/hidden_alerts_service.go
@@ -37,14 +37,14 @@ type hiddenAlertsBackend interface {
 
 // HiddenAlertsService manages hidden alerts and rules for users
 type HiddenAlertsService struct {
-	backendClient       hiddenAlertsBackend
-	mu                  sync.RWMutex
-	userHiddenAlerts    map[string]map[string]bool             // userID -> fingerprint -> hidden
-	userHiddenRules     map[string][]models.UserHiddenRule     // userID -> rules
-	compiledRegexRules  map[string]map[string]*regexp.Regexp   // userID -> ruleID -> compiled regex
-	lastAccess          map[string]time.Time                   // userID -> last successful LoadUserData fetch
-	generation          map[string]uint64                      // userID -> bumped by every mutation/invalidation
-	cacheTTL            time.Duration
+	backendClient      hiddenAlertsBackend
+	mu                 sync.RWMutex
+	userHiddenAlerts   map[string]map[string]bool           // userID -> fingerprint -> hidden
+	userHiddenRules    map[string][]models.UserHiddenRule   // userID -> rules
+	compiledRegexRules map[string]map[string]*regexp.Regexp // userID -> ruleID -> compiled regex
+	lastAccess         map[string]time.Time                 // userID -> last successful LoadUserData fetch
+	generation         map[string]uint64                    // userID -> bumped by every mutation/invalidation
+	cacheTTL           time.Duration
 }
 
 // NewHiddenAlertsService creates a new hidden alerts service
--- a/internal/webui/services/hidden_alerts_service.go
+++ b/internal/webui/services/hidden_alerts_service.go
@@ -9,7 +9,6 @@ import (
 
 	"notificator/internal/backend/models"
 	alertpb "notificator/internal/backend/proto/alert"
-	"notificator/internal/webui/client"
 	webuimodels "notificator/internal/webui/models"
 )
 
@@ -18,29 +17,45 @@ import (
 const sessionIdleTTL = 7 * 24 * time.Hour
 
 // cacheTTL bounds how long a cached hidden-alerts snapshot is served without
-// re-fetching. Mutations (hide/unhide/save/remove rule) update or invalidate the
-// cache in-process, so this only covers changes made by another webui replica.
+// re-fetching. The cache is keyed by sessionID and mutations only touch the
+// acting session's entry, so a change made from any other session (the same
+// user on a second device, an admin acting through impersonation, or another
+// webui replica) becomes visible after at most cacheTTL.
 const cacheTTL = 30 * time.Second
+
+// hiddenAlertsBackend is the slice of client.BackendClient this service calls.
+// It exists as a seam so tests can inject a backend that returns errors.
+type hiddenAlertsBackend interface {
+	GetUserHiddenAlerts(sessionID string, impersonateUserID ...string) ([]*alertpb.UserHiddenAlert, error)
+	GetUserHiddenRules(sessionID string, impersonateUserID ...string) ([]*alertpb.UserHiddenRule, error)
+	HideAlert(sessionID, fingerprint, alertName, instance, reason string, impersonateUserID ...string) error
+	UnhideAlert(sessionID, fingerprint string, impersonateUserID ...string) error
+	SaveHiddenRule(sessionID string, rule *alertpb.UserHiddenRule, impersonateUserID ...string) (*alertpb.UserHiddenRule, error)
+	RemoveHiddenRule(sessionID, ruleID string, impersonateUserID ...string) error
+	ClearAllHiddenAlerts(sessionID string, impersonateUserID ...string) error
+}
 
 // HiddenAlertsService manages hidden alerts and rules for users
 type HiddenAlertsService struct {
-	backendClient       *client.BackendClient
+	backendClient       hiddenAlertsBackend
 	mu                  sync.RWMutex
 	userHiddenAlerts    map[string]map[string]bool             // userID -> fingerprint -> hidden
 	userHiddenRules     map[string][]models.UserHiddenRule     // userID -> rules
 	compiledRegexRules  map[string]map[string]*regexp.Regexp   // userID -> ruleID -> compiled regex
-	lastAccess          map[string]time.Time                   // userID -> last LoadUserData fetch
+	lastAccess          map[string]time.Time                   // userID -> last successful LoadUserData fetch
+	generation          map[string]uint64                      // userID -> bumped by every mutation/invalidation
 	cacheTTL            time.Duration
 }
 
 // NewHiddenAlertsService creates a new hidden alerts service
-func NewHiddenAlertsService(backendClient *client.BackendClient) *HiddenAlertsService {
+func NewHiddenAlertsService(backendClient hiddenAlertsBackend) *HiddenAlertsService {
 	service := &HiddenAlertsService{
 		backendClient:      backendClient,
 		userHiddenAlerts:   make(map[string]map[string]bool),
 		userHiddenRules:    make(map[string][]models.UserHiddenRule),
 		compiledRegexRules: make(map[string]map[string]*regexp.Regexp),
 		lastAccess:         make(map[string]time.Time),
+		generation:         make(map[string]uint64),
 		cacheTTL:           cacheTTL,
 	}
 	
@@ -66,6 +81,7 @@ func (s *HiddenAlertsService) LoadUserData(sessionID string) error {
 
 	s.mu.RLock()
 	fresh := s.userHiddenAlerts[sessionID] != nil && time.Since(s.lastAccess[sessionID]) < s.cacheTTL
+	gen := s.generation[sessionID]
 	s.mu.RUnlock()
 	if fresh {
 		return nil
@@ -109,6 +125,13 @@ func (s *HiddenAlertsService) LoadUserData(sessionID string) error {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
+	// A mutation or invalidation landed while the fetch was in flight: this
+	// snapshot predates it. Drop it and leave the entry cold so the next call
+	// refetches, rather than republishing the pre-mutation state for a TTL.
+	if s.generation[sessionID] != gen {
+		return nil
+	}
+
 	// Replace cached hidden alerts wholesale so entries removed in the backend
 	// (or stale regexes for deleted rules) do not accumulate.
 	if hiddenAlertsErr == nil {
@@ -133,7 +156,14 @@ func (s *HiddenAlertsService) LoadUserData(sessionID string) error {
 		}
 	}
 
-	s.lastAccess[sessionID] = time.Now()
+	// Only a complete fetch is cacheable: on error the fallbacks above published
+	// an empty snapshot, so backdate the clock to keep the entry sweepable but
+	// never fresh, and let the next request retry the backend.
+	if hiddenAlertsErr == nil && hiddenRulesErr == nil {
+		s.lastAccess[sessionID] = time.Now()
+	} else {
+		s.lastAccess[sessionID] = time.Now().Add(-s.cacheTTL)
+	}
 	// ponytail: opportunistic sweep instead of a janitor goroutine — sessions
 	// idle past the backend session TTL are dropped on the next load by anyone.
 	s.sweepIdleSessionsLocked()
@@ -150,6 +180,7 @@ func (s *HiddenAlertsService) sweepIdleSessionsLocked() {
 			delete(s.userHiddenRules, sessionID)
 			delete(s.compiledRegexRules, sessionID)
 			delete(s.lastAccess, sessionID)
+			delete(s.generation, sessionID)
 		}
 	}
 }
@@ -293,7 +324,8 @@ func (s *HiddenAlertsService) HideAlert(sessionID string, alert *webuimodels.Das
 		s.userHiddenAlerts[sessionID] = make(map[string]bool)
 	}
 	s.userHiddenAlerts[sessionID][alert.Fingerprint] = true
-	
+	s.generation[sessionID]++
+
 	return nil
 }
 
@@ -310,7 +342,8 @@ func (s *HiddenAlertsService) UnhideAlert(sessionID, fingerprint string, imperso
 	if s.userHiddenAlerts[sessionID] != nil {
 		delete(s.userHiddenAlerts[sessionID], fingerprint)
 	}
-	
+	s.generation[sessionID]++
+
 	return nil
 }
 
@@ -395,6 +428,7 @@ func (s *HiddenAlertsService) InvalidateCache(sessionID string) {
 	delete(s.userHiddenRules, sessionID)
 	delete(s.compiledRegexRules, sessionID)
 	delete(s.lastAccess, sessionID)
+	s.generation[sessionID]++
 }
 
 // GetUserHiddenRules gets all hidden rules for a user
--- a/internal/webui/services/hidden_alerts_service.go
+++ b/internal/webui/services/hidden_alerts_service.go
@@ -16,12 +16,12 @@ import (
 // idle longer than this are unreachable (session expired) and can be dropped.
 const sessionIdleTTL = 7 * 24 * time.Hour
 
-// cacheTTL bounds how long a cached hidden-alerts snapshot is served without
-// re-fetching. The cache is keyed by sessionID and mutations only touch the
-// acting session's entry, so a change made from any other session (the same
+// hiddenAlertsCacheTTL bounds how long a cached hidden-alerts snapshot is served
+// without re-fetching. The cache is keyed by sessionID and mutations only touch
+// the acting session's entry, so a change made from any other session (the same
 // user on a second device, an admin acting through impersonation, or another
-// webui replica) becomes visible after at most cacheTTL.
-const cacheTTL = 30 * time.Second
+// webui replica) becomes visible after at most this TTL.
+const hiddenAlertsCacheTTL = 30 * time.Second
 
 // hiddenAlertsBackend is the slice of client.BackendClient this service calls.
 // It exists as a seam so tests can inject a backend that returns errors.
@@ -56,7 +56,7 @@ func NewHiddenAlertsService(backendClient hiddenAlertsBackend) *HiddenAlertsServ
 		compiledRegexRules: make(map[string]map[string]*regexp.Regexp),
 		lastAccess:         make(map[string]time.Time),
 		generation:         make(map[string]uint64),
-		cacheTTL:           cacheTTL,
+		cacheTTL:           hiddenAlertsCacheTTL,
 	}
 	
 	// Load initial data
@@ -73,7 +73,7 @@ func (s *HiddenAlertsService) LoadAllUserData() {
 }
 
 // LoadUserData loads hidden alerts and rules for a specific user using sessionID,
-// serving the cached snapshot when it is younger than cacheTTL.
+// serving the cached snapshot when it is younger than hiddenAlertsCacheTTL.
 func (s *HiddenAlertsService) LoadUserData(sessionID string) error {
 	// Get userID from session for cache key
 	// Note: We'll need to pass userID separately or get it from session
@@ -126,9 +126,14 @@ func (s *HiddenAlertsService) LoadUserData(sessionID string) error {
 	defer s.mu.Unlock()
 
 	// A mutation or invalidation landed while the fetch was in flight: this
-	// snapshot predates it. Drop it and leave the entry cold so the next call
+	// snapshot predates it. Drop it and clear the entry so the next call
 	// refetches, rather than republishing the pre-mutation state for a TTL.
+	// Clearing matters for HideAlert, which creates the snapshot map before
+	// bumping the generation: leaving it behind would look "loaded" to
+	// IsAlertHidden while carrying no rules. Losing the optimistic hide is safe,
+	// the backend already persisted it and the refetch brings it back.
 	if s.generation[sessionID] != gen {
+		s.invalidateLocked(sessionID)
 		return nil
 	}
 
@@ -436,7 +441,13 @@ func (s *HiddenAlertsService) RemoveHiddenRule(sessionID, ruleID string, imperso
 func (s *HiddenAlertsService) InvalidateCache(sessionID string) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
-	
+
+	s.invalidateLocked(sessionID)
+	s.generation[sessionID]++
+}
+
+// invalidateLocked drops a session's cached snapshot. Caller must hold s.mu.
+func (s *HiddenAlertsService) invalidateLocked(sessionID string) {
 	delete(s.userHiddenAlerts, sessionID)
 	delete(s.userHiddenRules, sessionID)
 	delete(s.compiledRegexRules, sessionID)
@@ -444,7 +455,6 @@ func (s *HiddenAlertsService) InvalidateCache(sessionID string) {
 	// snapshot is gone and LoadUserData checks that first) but stays visible to
 	// sweepIdleSessionsLocked, which is what reclaims the generation counter.
 	s.lastAccess[sessionID] = time.Now().Add(-s.cacheTTL)
-	s.generation[sessionID]++
 }
 
 // GetUserHiddenRules gets all hidden rules for a user
--- a/internal/webui/services/session_cache_cleanup_test.go
+++ b/internal/webui/services/session_cache_cleanup_test.go
@@ -8,6 +8,7 @@ import (
 
 	"notificator/internal/backend/models"
 	alertpb "notificator/internal/backend/proto/alert"
+	webuimodels "notificator/internal/webui/models"
 )
 
 // stubBackend implements only the two fetch methods LoadUserData needs; the
@@ -31,6 +32,10 @@ func (b *stubBackend) GetUserHiddenAlerts(string, ...string) ([]*alertpb.UserHid
 func (b *stubBackend) GetUserHiddenRules(string, ...string) ([]*alertpb.UserHiddenRule, error) {
 	return nil, b.err
 }
+
+func (b *stubBackend) ClearAllHiddenAlerts(string, ...string) error { return nil }
+
+func (b *stubBackend) HideAlert(string, string, string, string, string, ...string) error { return nil }
 
 func TestColorServiceSweepExpired(t *testing.T) {
 	cs := NewColorService(nil)
@@ -151,8 +156,43 @@ func TestHiddenAlertsServiceLoadUserDataDropsSnapshotInvalidatedMidFetch(t *test
 	if _, ok := s.userHiddenAlerts["sess"]; ok {
 		t.Error("snapshot predating the mid-fetch invalidation should not be published")
 	}
-	if _, ok := s.lastAccess["sess"]; ok {
-		t.Error("discarded snapshot should leave the entry cold, not fresh")
+	if time.Since(s.lastAccess["sess"]) < s.cacheTTL {
+		t.Error("discarded snapshot should leave the entry stale, not fresh")
+	}
+}
+
+// Clear-all is a mutation like any other: an in-flight fetch that predates it
+// must not republish the pre-clear fingerprint set for a full TTL.
+func TestHiddenAlertsServiceLoadUserDataDropsSnapshotClearedMidFetch(t *testing.T) {
+	backend := &stubBackend{alerts: []*alertpb.UserHiddenAlert{{Fingerprint: "stale"}}}
+	s := NewHiddenAlertsService(backend)
+	backend.onFetch = func() { _ = s.ClearAllHiddenAlerts("sess") }
+
+	if err := s.LoadUserData("sess"); err != nil {
+		t.Fatalf("LoadUserData: %v", err)
+	}
+
+	if _, ok := s.userHiddenAlerts["sess"]; ok {
+		t.Error("snapshot predating the mid-fetch clear-all should not be published")
+	}
+}
+
+// An impersonated mutation targets another user's hidden set, so it must not be
+// written into the acting session's snapshot.
+func TestHiddenAlertsServiceHideAlertIgnoresCacheWhenImpersonating(t *testing.T) {
+	s := NewHiddenAlertsService(&stubBackend{})
+	s.userHiddenAlerts["sess"] = map[string]bool{}
+	s.lastAccess["sess"] = time.Now()
+
+	if err := s.HideAlert("sess", &webuimodels.DashboardAlert{Fingerprint: "other-user-fp"}, "", "impersonated-user"); err != nil {
+		t.Fatalf("HideAlert: %v", err)
+	}
+
+	if len(s.userHiddenAlerts["sess"]) != 0 {
+		t.Error("impersonated hide should not touch the acting session's snapshot")
+	}
+	if s.generation["sess"] != 0 {
+		t.Error("impersonated hide should not bump the acting session's generation")
 	}
 }
 
@@ -165,7 +205,12 @@ func TestHiddenAlertsServiceInvalidateCacheRemovesAllMaps(t *testing.T) {
 
 	s.InvalidateCache("sess")
 
-	if len(s.userHiddenAlerts)+len(s.userHiddenRules)+len(s.compiledRegexRules)+len(s.lastAccess) != 0 {
-		t.Error("InvalidateCache should remove the session from all maps")
+	if len(s.userHiddenAlerts)+len(s.userHiddenRules)+len(s.compiledRegexRules) != 0 {
+		t.Error("InvalidateCache should remove the session's cached data")
+	}
+	// lastAccess is kept but backdated, so the idle sweep can still reclaim the
+	// generation counter for sessions whose last action was an invalidation.
+	if time.Since(s.lastAccess["sess"]) < s.cacheTTL {
+		t.Error("InvalidateCache should leave lastAccess stale, not fresh")
 	}
 }
--- a/internal/webui/services/session_cache_cleanup_test.go
+++ b/internal/webui/services/session_cache_cleanup_test.go
@@ -105,6 +105,23 @@ func TestHiddenAlertsServiceLoadUserDataServesCacheWithinTTL(t *testing.T) {
 	}
 }
 
+// A successful load must mark the entry fresh, so the next call inside the TTL
+// is served from cache. This is the behaviour the whole PR exists for.
+func TestHiddenAlertsServiceLoadUserDataCachesAfterSuccessfulFetch(t *testing.T) {
+	backend := &stubBackend{alerts: []*alertpb.UserHiddenAlert{{Fingerprint: "fp"}}}
+	s := NewHiddenAlertsService(backend)
+
+	for range 2 {
+		if err := s.LoadUserData("sess"); err != nil {
+			t.Fatalf("LoadUserData: %v", err)
+		}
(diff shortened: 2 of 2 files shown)


internal/webui/services/hidden_alerts_service.go
```diff
@@ -50,12 +57,23 @@ func (s *HiddenAlertsService) LoadAllUserData() {
...
 	// Get userID from session for cache key
 	// Note: We'll need to pass userID separately or get it from session
 	// For now, we'll use sessionID as the cache key
 
+	s.mu.RLock()
+	fresh := s.userHiddenAlerts[sessionID] != nil && time.Since(s.lastAccess[sessionID]) < s.cacheTTL
+	s.mu.RUnlock()
+	if fresh {
+		return nil
+	}
```
SoulKyu, line 88: **A failed backend fetch is now cached as "fresh" for a full 30s**

`LoadUserData` bumps `s.lastAccess[sessionID] = time.Now()` unconditionally (line 136), on the error path as well as the success path. Before this PR that was harmless, because `lastAccess` only fed the 7-day idle sweep. Now it is the freshness clock for this new gate, so the error path becomes negative caching.

Concrete failure: backend restart or a transient gRPC error while a user's first dashboard request comes in.

1. `hiddenAlertsErr != nil` and `hiddenRulesErr != nil`, both logged and swallowed.
2. The `else if s.userHiddenAlerts[sessionID] == nil` branch (lines 120-122) publishes an empty **non-nil** map, and the rules/regex fallbacks publish empty collections.
3. `s.lastAccess[sessionID] = time.Now()`.
4. Every `LoadUserData` call for the next 30s hits `fresh == true` here and returns at line 71 without retrying.

Result: that session's hidden alerts and hide rules are all ignored for 30 seconds, and they stay ignored even after the backend is healthy again. Previously the very next dashboard request refetched and self-healed within one poll cycle.

Suggested fix — only a successful fetch should be cacheable, while `lastAccess` stays usable as the sweep clock:

```go
if hiddenAlertsErr == nil && hiddenRulesErr == nil {
	s.lastAccess[sessionID] = time.Now()
} else {
	// Keep the entry sweepable, but not fresh: retry on the next call.
	s.lastAccess[sessionID] = time.Now().Add(-s.cacheTTL)
}
```

Writing the zero time instead would make `sweepIdleSessionsLocked` drop the entry immediately, which also forces a refetch but throws away a snapshot that may still be good.

This is worth a regression test in `session_cache_cleanup_test.go`, and the current nil-client trick cannot express it, since a nil client panics rather than returning an error. The smallest seam is to type the `backendClient` field as a local interface covering the methods actually used (`GetUserHiddenAlerts`, `GetUserHiddenRules`, `HideAlert`, `UnhideAlert`, `SaveHiddenRule`, `RemoveHiddenRule`, `ClearAllHiddenAlerts`) so a stub can return errors; the test then asserts that after a failing fetch a second `LoadUserData` still calls the backend.


internal/webui/services/hidden_alerts_service.go
```diff
@@ -50,12 +57,23 @@ func (s *HiddenAlertsService) LoadAllUserData() {
...
+
+	// ponytail: no single-flight — concurrent misses for the same session just
+	// fetch twice and publish the same data; add one if that shows up in traces.
```
SoulKyu, line 91: **A mutation that lands during an in-flight fetch is republished away, and now sticks for 30s**

The note here is right that two concurrent *readers* are harmless. The case it does not cover is a writer racing a reader, and the TTL is what makes it matter.

Interleaving with `SaveHiddenRule` / `RemoveHiddenRule`:

1. A dashboard poll calls `LoadUserData`, misses the gate above, and starts the two gRPC fetches (lines 82 and 88) with no lock held.
2. The user saves a rule. `SaveHiddenRule` writes to the backend and calls `InvalidateCache` (line 390), which deletes the session's alerts, rules, regexes and `lastAccess`.
3. The in-flight fetch takes the write lock and publishes its **pre-save** snapshot, then sets `lastAccess = time.Now()` (line 136).

The entry is fresh again holding the old rule set, so the gate short-circuits for the next 30s. The same shape applies to `HideAlert`: it sets `s.userHiddenAlerts[sessionID][fp] = true` (line 295), and the wholesale map replacement at line 119 drops it if a fetch was in flight, so the alert un-hides for a TTL.

This contradicts the PR description's "hides and rule changes are still reflected on the very next request with no TTL wait". The interleaving existed before too, but it self-healed on the next dashboard request; with `GET /api/v1/dashboard/incremental` polling continuously, a mutation landing inside a fetch window is not exotic, and 30s of a rule appearing not to apply reads as a broken feature.

Suggested fix — a per-session generation counter, roughly six lines:

```go
// struct field, initialised in NewHiddenAlertsService
generation map[string]uint64

// InvalidateCache, under the existing write lock
s.generation[sessionID]++

// LoadUserData, read in the same RLock as the freshness check
gen := s.generation[sessionID]

// after taking the write lock, before publishing anything
if s.generation[sessionID] != gen {
	return nil // invalidated mid-fetch: leave the cache cold so the next call refetches
}
```

Having `HideAlert` / `UnhideAlert` bump the same counter closes their variant as well, at the cost of one extra refetch after each mutation.


internal/webui/services/hidden_alerts_service.go
```diff
@@ -17,14 +17,20 @@ import (
 // idle longer than this are unreachable (session expired) and can be dropped.
 const sessionIdleTTL = 7 * 24 * time.Hour
 
+// cacheTTL bounds how long a cached hidden-alerts snapshot is served without
+// re-fetching. Mutations (hide/unhide/save/remove rule) update or invalidate the
+// cache in-process, so this only covers changes made by another webui replica.
+const cacheTTL = 30 * time.Second
```
SoulKyu: **The TTL comment understates which changes go stale**

"Mutations (hide/unhide/save/remove rule) update or invalidate the cache in-process, so this only covers changes made by another webui replica" holds only for mutations issued from the *same session*. The cache is keyed by `sessionID`, not by user ID, so within a single replica:

- the same user logged in from a second browser or device has a distinct `sessionID`, so a hide made there is invisible to the first session for up to 30s;
- admin impersonation is affected in both directions. `HideAlert` writes into the **acting** session's map (`s.userHiddenAlerts[sessionID][alert.Fingerprint] = true`, line 295) while `impersonateUserID` only reaches the backend call, so the admin's own dashboard now carries that foreign fingerprint for up to a TTL instead of until their next request, and the impersonated user's session picks up the real change up to a TTL late.

The session keying and the impersonation write are both pre-existing, but the TTL is what turns "wrong until the next request" into "wrong for up to 30 seconds", so the comment should say so rather than pointing only at multi-replica deployments. Suggested wording:

```go
// cacheTTL bounds how long a cached hidden-alerts snapshot is served without
// re-fetching. The cache is keyed by sessionID and mutations only touch the
// acting session's entry, so a change made from any other session (the same
// user on a second device, an admin acting through impersonation, or another
// webui replica) becomes visible after at most cacheTTL.
```


SoulKyu (commented):
Nice, tightly-scoped perf change: the TTL gate is in the right place (checked under `RLock`, fetch still outside the write lock), the cold/expired/invalidated paths are untouched, and `go build -tags "nogui,webui" ./...` plus `go test ./internal/webui/services/` (51 tests) pass locally on this head.

Three non-blocking findings, all in `internal/webui/services/hidden_alerts_service.go`, left inline: backend fetch errors are now cached as fresh for a full TTL, a mutation landing during an in-flight fetch can be republished away for 30s, and the `cacheTTL` doc comment understates which changes go stale.




internal/webui/services/hidden_alerts_service.go
SoulKyu: @SoulKyu This has been fixed in 37c5738.

LoadUserData now only sets lastAccess=time.Now() when both fetches succeed; on error it writes time.Now().Add(-s.cacheTTL) so the entry stays sweepable but never fresh and the next request retries. backendClient is now typed as a local hiddenAlertsBackend interface, and TestHiddenAlertsServiceLoadUserDataRetriesAfterFailedFetch in session_cache_cleanup_test.go asserts two backend calls after a failing fetch.




internal/webui/services/hidden_alerts_service.go
SoulKyu: @SoulKyu This has been fixed in 37c5738.

Added a per-session generation map bumped by InvalidateCache, HideAlert and UnhideAlert; LoadUserData reads it in the same RLock as the freshness check and returns without publishing if it changed while the fetch was in flight, leaving the entry cold for a refetch. Covered by TestHiddenAlertsServiceLoadUserDataDropsSnapshotInvalidatedMidFetch.




internal/webui/services/hidden_alerts_service.go
SoulKyu: @SoulKyu This has been fixed in 37c5738.

Replaced the cacheTTL doc comment with the suggested wording explaining that the cache is keyed by sessionID and mutations only touch the acting session, so changes from a second device, admin impersonation or another webui replica become visible after at most cacheTTL.




SoulKyu:
**Looper fixer round complete** — 37c5738

- ✅ Review comment on `internal/webui/services/hidden_alerts_service.go:88` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/82#discussion_r3649542114)
  - LoadUserData now only sets lastAccess=time.Now() when both fetches succeed; on error it writes time.Now().Add(-s.cacheTTL) so the entry stays sweepable but never fresh and the next request retries. backendClient is now typed as a local hiddenAlertsBackend interface, and TestHiddenAlertsServiceLoadUserDataRetriesAfterFailedFetch in session_cache_cleanup_test.go asserts two backend calls after a failing fetch.
- ✅ Review comment on `internal/webui/services/hidden_alerts_service.go:91` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/82#discussion_r3649542115)
  - Added a per-session generation map bumped by InvalidateCache, HideAlert and UnhideAlert; LoadUserData reads it in the same RLock as the freshness check and returns without publishing if it changed while the fetch was in flight, leaving the entry cold for a refetch. Covered by TestHiddenAlertsServiceLoadUserDataDropsSnapshotInvalidatedMidFetch.
- ✅ Review comment on `internal/webui/services/hidden_alerts_service.go:24` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/82#discussion_r3649542116)
  - Replaced the cacheTTL doc comment with the suggested wording explaining that the cache is keyed by sessionID and mutations only touch the acting session, so changes from a second device, admin impersonation or another webui replica become visible after at most cacheTTL.


SoulKyu:

internal/webui/services/hidden_alerts_service.go
```diff
@@ -292,7 +342,8 @@ func (s *HiddenAlertsService) UnhideAlert(sessionID, fingerprint string, imperso
 	if s.userHiddenAlerts[sessionID] != nil {
 		delete(s.userHiddenAlerts[sessionID], fingerprint)
 	}
-	
+	s.generation[sessionID]++
```
SoulKyu, line 363: **`ClearAllHiddenAlerts` is the one mutation left out of the generation counter**

`HideAlert` (line 327), `UnhideAlert` (this line) and `InvalidateCache` (line 431) all bump `s.generation[sessionID]`, but `ClearAllHiddenAlerts` (lines 463-477) mutates the very same cache entry — `s.userHiddenAlerts[sessionID] = make(map[string]bool)` — without bumping the counter and without touching `lastAccess`.

That means the mid-fetch guard added at line 131 never fires for clear-all. Concrete sequence: request A calls `LoadUserData` for a cold or expired session and captures `gen = N` at line 84; while its two gRPC calls are in flight the user hits `DELETE /api/v1/dashboard/hidden-alerts` (wired at `internal/webui/router.go:273` -> `handlers.ClearAllHiddenAlerts` -> this service), which clears the backend and empties the map but leaves `generation` at `N`; A then takes the write lock, the guard at line 131 passes, and it republishes the pre-clear fingerprint set with `s.lastAccess[sessionID] = time.Now()`. Every dashboard request for the next 30s is served that snapshot, so alerts the user just un-hid stay hidden and there is no request the UI can make to force a refresh. Before this PR the same interleaving self-corrected on the very next dashboard request, because `LoadUserData` refetched unconditionally — the TTL is what turns a one-request glitch into a 30-second one.

Suggested change: add `s.generation[sessionID]++` next to the map reset in `ClearAllHiddenAlerts`, so all four mutators are consistent. A regression test mirroring `TestHiddenAlertsServiceLoadUserDataDropsSnapshotInvalidatedMidFetch` would catch it — `backend.onFetch = func() { _ = s.ClearAllHiddenAlerts("sess") }`, with a `ClearAllHiddenAlerts` method added to `stubBackend` returning `nil` (the embedded nil interface would otherwise panic), then assert `s.userHiddenAlerts["sess"]` is absent.


internal/webui/services/hidden_alerts_service.go
```diff
@@ -275,7 +324,8 @@ func (s *HiddenAlertsService) HideAlert(sessionID string, alert *webuimodels.Das
 		s.userHiddenAlerts[sessionID] = make(map[string]bool)
 	}
 	s.userHiddenAlerts[sessionID][alert.Fingerprint] = true
-	
+	s.generation[sessionID]++
```
SoulKyu, line 342: **Hides performed while impersonating now poison the acting session's snapshot for a full TTL**

This optimistic write plus the new generation bump run unconditionally, but the mutation may not be about the session owner at all. `internal/webui/handlers/hidden_alerts_handlers.go:61` passes `middleware.GetImpersonatedUserID(c)` straight into `HideAlert`, and `client.BackendClient.HideAlert` (`internal/webui/client/backend_client.go:1017`) routes the mutation to that impersonated user. The cache, however, is keyed purely by `sessionID`, and `LoadUserData` fetches the session owner's hidden alerts with no impersonation argument (line 98). So hiding an alert for user X writes X's fingerprint into the admin's own snapshot.

Before this PR that wrong entry was wiped by the next dashboard request, since `LoadUserData` refetched every time. Now the `fresh` check at line 83 short-circuits, so the admin's own dashboard — `IsAlertHidden(sessionID, alert)` at `internal/webui/handlers/dashboard_handlers.go:384` is session-keyed, not impersonation-aware — filters out an alert that is not in the admin's hidden set for up to 30s, and that persists past the end of the impersonation session.

Suggested change: gate the cache mutation on the same condition the backend client already uses, in both `HideAlert` and `UnhideAlert` (line 345), and in `ClearAllHiddenAlerts`:

```go
if len(impersonateUserID) == 0 || impersonateUserID[0] == "" {
	if s.userHiddenAlerts[sessionID] == nil {
		s.userHiddenAlerts[sessionID] = make(map[string]bool)
	}
	s.userHiddenAlerts[sessionID][alert.Fingerprint] = true
	s.generation[sessionID]++
}
```

An impersonated mutation says nothing about the acting session's own hidden set, so the correct action is to leave that entry untouched rather than to write a value the next refetch will contradict.


internal/webui/services/hidden_alerts_service.go
```diff
@@ -377,6 +428,7 @@ func (s *HiddenAlertsService) InvalidateCache(sessionID string) {
 	delete(s.userHiddenRules, sessionID)
 	delete(s.compiledRegexRules, sessionID)
 	delete(s.lastAccess, sessionID)
+	s.generation[sessionID]++
```
SoulKyu: **`generation` entries survive `InvalidateCache` in a state the idle sweep can never reach**

`InvalidateCache` deletes `lastAccess[sessionID]` (line 430) and then increments `generation[sessionID]` here. But `sweepIdleSessionsLocked` (line 176) iterates `s.lastAccess` only, so the `delete(s.generation, sessionID)` you added at line 183 can only ever fire for sessions that still have a `lastAccess` entry. Any session whose last action is `SaveHiddenRule` or `RemoveHiddenRule` — both call `InvalidateCache` (lines 404 and 417) — leaves a `generation` entry behind that nothing reclaims for the lifetime of the process. Each entry is tiny, but the growth is monotonic and keyed by session ID, which is exactly the unbounded dimension the 7-day sweep was added to bound.

Suggested change: rather than deleting `lastAccess` in `InvalidateCache`, backdate it the way the error path at line 165 does — `s.lastAccess[sessionID] = time.Now().Add(-s.cacheTTL)`. The entry stays visible to the sweep and can never read as fresh, because `s.userHiddenAlerts[sessionID]` has just been deleted and the `fresh` check at line 83 tests that first. Two tests assert key absence today and would need to assert staleness instead: `TestHiddenAlertsServiceInvalidateCacheRemovesAllMaps` (`internal/webui/services/session_cache_cleanup_test.go:168`) and `TestHiddenAlertsServiceLoadUserDataDropsSnapshotInvalidatedMidFetch` (same file, line 154).

Worth flagging the fix that looks obvious but is not safe: do not sweep `generation` keys that have no `lastAccess` entry. A sweep runs inside another session's publish, so it could delete the counter for a session that was invalidated while its own fetch was still in flight, resetting it to 0 and re-opening precisely the race the guard at line 131 closes.


SoulKyu (commented):
The TTL gate itself looks right: the freshness check runs under `RLock`, the fetch still happens outside the write lock, and the two fixes from the previous round (backdating `lastAccess` on a failed fetch, and the generation guard for mutations landing mid-fetch) are both in and covered by tests. `go build -tags "nogui,webui" ./...` and `go test ./internal/webui/services/` both pass locally on 848d791 (53 tests).

Three non-blocking follow-ups, all in `internal/webui/services/hidden_alerts_service.go` and all about the mutation paths around the new cache: `ClearAllHiddenAlerts` was left out of the generation counter, impersonated hides/unhides now pin a wrong snapshot for a full TTL, and `generation` entries survive `InvalidateCache` in a way the idle sweep cannot reach. Details inline.




internal/webui/services/hidden_alerts_service.go
SoulKyu: @SoulKyu This has been fixed in f1e76f7.

ClearAllHiddenAlerts now bumps s.generation[sessionID] alongside the map reset (internal/webui/services/hidden_alerts_service.go), so the mid-fetch guard in LoadUserData drops snapshots that predate a clear-all. Added TestHiddenAlertsServiceLoadUserDataDropsSnapshotClearedMidFetch plus a ClearAllHiddenAlerts method on stubBackend.




internal/webui/services/hidden_alerts_service.go
SoulKyu: @SoulKyu This has been fixed in f1e76f7.

Added an isImpersonating helper mirroring the backend client's check and returned early before the cache mutation in HideAlert, UnhideAlert and ClearAllHiddenAlerts, so an impersonated mutation leaves the acting session's snapshot and generation untouched. Covered by TestHiddenAlertsServiceHideAlertIgnoresCacheWhenImpersonating.




internal/webui/services/hidden_alerts_service.go
SoulKyu: @SoulKyu This has been fixed in f1e76f7.

InvalidateCache now sets lastAccess[sessionID] = time.Now().Add(-s.cacheTTL) instead of deleting it, keeping the entry visible to sweepIdleSessionsLocked while never reading as fresh. Updated TestHiddenAlertsServiceInvalidateCacheRemovesAllMaps and TestHiddenAlertsServiceLoadUserDataDropsSnapshotInvalidatedMidFetch to assert staleness rather than key absence.




SoulKyu:
**Looper fixer round complete** — f1e76f7

- ✅ Review comment on `internal/webui/services/hidden_alerts_service.go:358` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/82#discussion_r3649832960)
  - ClearAllHiddenAlerts now bumps s.generation[sessionID] alongside the map reset (internal/webui/services/hidden_alerts_service.go), so the mid-fetch guard in LoadUserData drops snapshots that predate a clear-all. Added TestHiddenAlertsServiceLoadUserDataDropsSnapshotClearedMidFetch plus a ClearAllHiddenAlerts method on stubBackend.
- ✅ Review comment on `internal/webui/services/hidden_alerts_service.go:337` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/82#discussion_r3649832963)
  - Added an isImpersonating helper mirroring the backend client's check and returned early before the cache mutation in HideAlert, UnhideAlert and ClearAllHiddenAlerts, so an impersonated mutation leaves the acting session's snapshot and generation untouched. Covered by TestHiddenAlertsServiceHideAlertIgnoresCacheWhenImpersonating.
- ✅ Review comment on `internal/webui/services/hidden_alerts_service.go:447` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/82#discussion_r3649832968)
  - InvalidateCache now sets lastAccess[sessionID] = time.Now().Add(-s.cacheTTL) instead of deleting it, keeping the entry visible to sweepIdleSessionsLocked while never reading as fresh. Updated TestHiddenAlertsServiceInvalidateCacheRemovesAllMaps and TestHiddenAlertsServiceLoadUserDataDropsSnapshotInvalidatedMidFetch to assert staleness rather than key absence.


SoulKyu:

internal/webui/services/hidden_alerts_service.go
```diff
@@ -91,6 +125,13 @@ func (s *HiddenAlertsService) LoadUserData(sessionID string) error {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
+	// A mutation or invalidation landed while the fetch was in flight: this
+	// snapshot predates it. Drop it and leave the entry cold so the next call
+	// refetches, rather than republishing the pre-mutation state for a TTL.
+	if s.generation[sessionID] != gen {
+		return nil
+	}
```
SoulKyu, line 138: **The mid-fetch guard does not actually leave the entry cold when the racing mutation was a `HideAlert`**

**Problem.** The comment states the intent as "leave the entry cold so the next call refetches", but the guard returns without touching any map. `HideAlert` creates the snapshot map before it bumps the counter (`if s.userHiddenAlerts[sessionID] == nil { ... = make(map[string]bool) }`, then `s.generation[sessionID]++`). So when the racing mutation is a hide against a session that had no snapshot, the guard fires and leaves behind a half-populated entry: `userHiddenAlerts[sessionID]` holds exactly one fingerprint, while `userHiddenRules[sessionID]` and `compiledRegexRules[sessionID]` are still nil.

**Why it matters.** `IsAlertHidden` gates its lazy reload on `s.userHiddenAlerts[sessionID] != nil` (line 192), not on freshness. The entry therefore reads as "loaded" while carrying no rules, and the reload never fires. `internal/webui/handlers/dashboard_handlers.go:133` calls `LoadUserData` and then loops `IsAlertHidden` at line 384, so the very request that lost the race renders alerts that an enabled hidden rule should have filtered out. Before this PR the window did not exist, because `LoadUserData` always published its snapshot. It self-heals on the next request (`lastAccess` is never written on this path, so the entry is stale) but the incorrect dashboard response has already been served.

**Repro.** Cold cache entry (first dashboard load after login, or immediately after `SaveHiddenRule`/`RemoveHiddenRule` call `InvalidateCache`), a `GET /api/v1/dashboard/data` with its fetch in flight, a concurrent `POST` hide for the same session, and at least one enabled `UserHiddenRule` for that user. The rule-hidden alerts appear in that response.

**Suggested change.** Make the guard do what the comment says by reusing the invalidation semantics. Extract the body of `InvalidateCache` into a lock-free helper and call it from both places:

```go
func (s *HiddenAlertsService) invalidateLocked(sessionID string) {
	delete(s.userHiddenAlerts, sessionID)
	delete(s.userHiddenRules, sessionID)
	delete(s.compiledRegexRules, sessionID)
	s.lastAccess[sessionID] = time.Now().Add(-s.cacheTTL)
}
```

then here:

```go
if s.generation[sessionID] != gen {
	s.invalidateLocked(sessionID)
	return nil
}
```

Discarding the optimistic `HideAlert` write is safe: the backend already persisted it, and the forced refetch on the next `LoadUserData` (or the next `IsAlertHidden`, which now sees a nil map and reloads) brings it back.

**Test.** Add a case next to `TestHiddenAlertsServiceLoadUserDataDropsSnapshotClearedMidFetch` that covers the cold-entry hide specifically, since that is the interleaving the current tests miss — `onFetch` calling `s.HideAlert("sess", &webuimodels.DashboardAlert{Fingerprint: "fpA"}, "")` on a service with no primed snapshot, then asserting `s.userHiddenAlerts["sess"]` is absent so `IsAlertHidden` is forced to reload.


internal/webui/services/session_cache_cleanup_test.go
```diff
@@ -58,6 +87,115 @@ func TestHiddenAlertsServiceSweepIdleSessions(t *testing.T) {
...
+// The nil backend client is the assertion: any gRPC fetch dereferences it and
+// panics, so "did not panic" means the cached snapshot was served.
+func TestHiddenAlertsServiceLoadUserDataServesCacheWithinTTL(t *testing.T) {
+	s := NewHiddenAlertsService(nil)
+	s.userHiddenAlerts["sess"] = map[string]bool{"fp": true}
+	s.lastAccess["sess"] = time.Now()
+
+	if err := s.LoadUserData("sess"); err != nil {
+		t.Fatalf("LoadUserData: %v", err)
+	}
```
SoulKyu: **Nit: assert on the stub's call counter rather than on a nil-pointer panic**

**Problem.** The only assertion in `TestHiddenAlertsServiceLoadUserDataServesCacheWithinTTL` is "the test did not panic", and its sibling `TestHiddenAlertsServiceLoadUserDataRefetchesWhenStale` asserts `recover() != nil`. Both are coupled to the incidental fact that a nil backend dereferences, not to the behaviour under test ("no fetch happened" / "a fetch happened").

**Why it matters.** Any future change that stops the fetch path from dereferencing the backend — an early return for an empty session ID, a nil-client guard, the single-flight noted a few lines below — turns the within-TTL test into a vacuous pass that no longer exercises the TTL gate at all, while the stale test starts failing for an unrelated reason. Since `stubBackend` in this file already counts invocations (`calls` is incremented in `GetUserHiddenAlerts`), the direct assertion is right there.

**Suggested change.** Use the stub in both tests:

```go
backend := &stubBackend{}
s := NewHiddenAlertsService(backend)
s.userHiddenAlerts["sess"] = map[string]bool{"fp": true}
s.lastAccess["sess"] = time.Now()

if err := s.LoadUserData("sess"); err != nil {
	t.Fatalf("LoadUserData: %v", err)
}
if backend.calls != 0 {
	t.Errorf("fresh cache should not fetch: got %d backend calls, want 0", backend.calls)
}
```

and in the expired/cold/invalidated matrix, replace the `recover()` guard with `if backend.calls != 1 { ... }`. That also lets each case assert on the snapshot that was published, not merely that a fetch was attempted — for example that the `expired` case ends up with the fingerprints the stub returned rather than the primed `"fp"`.


SoulKyu (commented):
@SoulKyu Third pass over the TTL gate, and the two fixes from the previous round hold up well: the failed-fetch backdating means an outage can no longer blank a session's hidden set for 30s, and `ClearAllHiddenAlerts` is now inside the generation counter alongside hide/unhide. The impersonation guard is the right call too, since `LoadUserData` fetches without an impersonated user, so the acting session's snapshot genuinely says nothing about the impersonated user's hidden set.

Verified locally on this head: `go build -tags "nogui,webui" ./...` is clean and `go test -race -count=2 ./internal/webui/services/` is green (55 tests). I also checked that every hidden-alert mutation still routes through `HiddenAlertsService` — no handler calls `backendClient.HideAlert` / `UnhideAlert` / `ClearAllHiddenAlerts` / `SaveHiddenRule` / `RemoveHiddenRule` directly — so nothing can mutate the backend behind the cache's back and go unnoticed for a TTL.

One non-blocking correctness gap remains in the mid-fetch guard, plus two nits on test assertions and naming. Details inline.



internal/webui/services/hidden_alerts_service.go
```diff
@@ -9,32 +9,54 @@ import (
...
 // sessionIdleTTL matches the backend session TTL: cache entries for sessions
 // idle longer than this are unreachable (session expired) and can be dropped.
 const sessionIdleTTL = 7 * 24 * time.Hour
 
+// cacheTTL bounds how long a cached hidden-alerts snapshot is served without
+// re-fetching. The cache is keyed by sessionID and mutations only touch the
+// acting session's entry, so a change made from any other session (the same
+// user on a second device, an admin acting through impersonation, or another
+// webui replica) becomes visible after at most cacheTTL.
+const cacheTTL = 30 * time.Second
```
SoulKyu: **Nit: `cacheTTL` is a package-scoped name in a package that holds several caches**

`package services` also contains `ColorService` (whose `cacheTTL` field is 5 minutes, `color_service.go:66`), `AlertCache` and `SentryService`, so a bare package-level `const cacheTTL` reads as "the TTL for caches in this package" rather than "the hidden-alerts snapshot TTL". It is referenced exactly once, at line 59.

Related: the PR description says this value mirrors `ColorService.cacheTTL`, but that one is 5 minutes while this is 30 seconds — the two disagree, and a future reader syncing them would silently change behaviour.

**Suggested change.** Rename the constant to `hiddenAlertsCacheTTL` (single call site at line 59, plus the two doc-comment mentions at lines 19 and 76), and drop the "mirroring `ColorService.cacheTTL`" claim from the PR description so the 30s value is documented on its own terms.


internal/webui/services/hidden_alerts_service.go
SoulKyu: @SoulKyu This has been fixed in 652b421.

Extracted InvalidateCache's body into invalidateLocked (hidden_alerts_service.go:436-455) and called it from the generation guard at line 131, so a HideAlert racing a cold fetch no longer leaves a half-populated userHiddenAlerts entry that IsAlertHidden reads as loaded. Added TestHiddenAlertsServiceLoadUserDataDropsSnapshotHiddenMidFetchOnColdEntry covering that interleaving.




internal/webui/services/session_cache_cleanup_test.go
SoulKyu: @SoulKyu This has been fixed in 652b421.

Both tests now use stubBackend and assert on its call counter: the within-TTL case checks backend.calls == 0, and the expired/cold/invalidated matrix checks backend.calls == 1 plus that the published snapshot holds the fetched fingerprint rather than the primed one. The recover()-based guard is gone.




internal/webui/services/hidden_alerts_service.go
SoulKyu: @SoulKyu This has been fixed in 652b421.

Renamed the package-level constant to hiddenAlertsCacheTTL in hidden_alerts_service.go (declaration, the single call site in NewHiddenAlertsService, and both doc-comment mentions). The PR body's 'mirroring ColorService.cacheTTL' claim was left untouched because this run is not permitted to edit PR metadata.




SoulKyu:
**Looper fixer round complete** — 652b421

- ✅ Review comment on `internal/webui/services/hidden_alerts_service.go:138` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/82#discussion_r3649993861)
  - Extracted InvalidateCache's body into invalidateLocked (hidden_alerts_service.go:436-455) and called it from the generation guard at line 131, so a HideAlert racing a cold fetch no longer leaves a half-populated userHiddenAlerts entry that IsAlertHidden reads as loaded. Added TestHiddenAlertsServiceLoadUserDataDropsSnapshotHiddenMidFetchOnColdEntry covering that interleaving.
- ✅ Review comment on `internal/webui/services/session_cache_cleanup_test.go` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/82#discussion_r3649993864)
  - Both tests now use stubBackend and assert on its call counter: the within-TTL case checks backend.calls == 0, and the expired/cold/invalidated matrix checks backend.calls == 1 plus that the published snapshot holds the fetched fingerprint rather than the primed one. The recover()-based guard is gone.
- ✅ Review comment on `internal/webui/services/hidden_alerts_service.go` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/82#discussion_r3649993866)
  - Renamed the package-level constant to hiddenAlertsCacheTTL in hidden_alerts_service.go (declaration, the single call site in NewHiddenAlertsService, and both doc-comment mentions). The PR body's 'mirroring ColorService.cacheTTL' claim was left untouched because this run is not permitted to edit PR metadata.


SoulKyu:

internal/webui/services/session_cache_cleanup_test.go
```diff
@@ -58,6 +87,146 @@ func TestHiddenAlertsServiceSweepIdleSessions(t *testing.T) {
 	}
 }
 
+func TestHiddenAlertsServiceLoadUserDataServesCacheWithinTTL(t *testing.T) {
+	backend := &stubBackend{}
+	s := NewHiddenAlertsService(backend)
+	s.userHiddenAlerts["sess"] = map[string]bool{"fp": true}
+	s.lastAccess["sess"] = time.Now()
```
SoulKyu, line 94: **No test covers the write that makes the cache a cache**

**Problem.** Every test in this file exercises the freshness *check* but none exercises the freshness *write*. This test primes `s.lastAccess["sess"]` by hand before calling `LoadUserData`, so it proves the gate honours a fresh timestamp — it never proves that a successful fetch produces one. `TestHiddenAlertsServiceLoadUserDataRefetchesWhenStale` asserts `calls == 1` after a single load, `TestHiddenAlertsServiceLoadUserDataRetriesAfterFailedFetch` asserts `calls == 2` on the error path, and the two mid-fetch tests assert the entry is absent. None of them calls `LoadUserData` twice on a healthy backend.

**Why it matters.** The one line that implements this PR is `s.lastAccess[sessionID] = time.Now()` on the success branch of `hidden_alerts_service.go:167-171`. I checked this by mutation: replacing that whole `if/else` with the unconditional backdate `s.lastAccess[sessionID] = time.Now().Add(-s.cacheTTL)` — which reverts the PR's entire behaviour, restoring two gRPC round-trips per dashboard request — leaves `go test ./internal/webui/services/` fully green. A silent perf regression back to the pre-PR state is exactly the kind of thing that ships unnoticed, and issue #78 would quietly reopen with no red test to catch it.

**Suggested change.** Add one test that drives the full round-trip through the public method instead of priming internals, so the success-path write is load-bearing:

```go
// A successful load must mark the entry fresh, so the next call inside the TTL
// is served from cache. This is the behaviour the whole PR exists for.
func TestHiddenAlertsServiceLoadUserDataCachesAfterSuccessfulFetch(t *testing.T) {
	backend := &stubBackend{alerts: []*alertpb.UserHiddenAlert{{Fingerprint: "fp"}}}
	s := NewHiddenAlertsService(backend)

	for range 2 {
		if err := s.LoadUserData("sess"); err != nil {
			t.Fatalf("LoadUserData: %v", err)
		}
	}

	if backend.calls != 1 {
		t.Errorf("second load inside the TTL should be served from cache: got %d backend calls, want 1", backend.calls)
	}
}
```

That pairs symmetrically with the existing `RetriesAfterFailedFetch` test (two loads, `want 2` on the error path / `want 1` on the success path), and it fails under the mutation above. It also makes the existing `ServesCacheWithinTTL` test the narrower unit it already is, which is fine to keep alongside it.


internal/webui/services/hidden_alerts_service.go
```diff
@@ -9,32 +9,54 @@ import (
...
-	backendClient       *client.BackendClient
+	backendClient       hiddenAlertsBackend
 	mu                  sync.RWMutex
 	userHiddenAlerts    map[string]map[string]bool             // userID -> fingerprint -> hidden
 	userHiddenRules     map[string][]models.UserHiddenRule     // userID -> rules
 	compiledRegexRules  map[string]map[string]*regexp.Regexp   // userID -> ruleID -> compiled regex
-	lastAccess          map[string]time.Time                   // userID -> last LoadUserData call
+	lastAccess          map[string]time.Time                   // userID -> last successful LoadUserData fetch
+	generation          map[string]uint64                      // userID -> bumped by every mutation/invalidation
+	cacheTTL            time.Duration
```
SoulKyu: **Nit: the struct block is no longer gofmt-aligned after the two new fields**

**Problem.** The comment and type columns here are still spaced for the old `*client.BackendClient` type, so with `backendClient` narrowed to `hiddenAlertsBackend` and `generation` / `cacheTTL` added, `gofmt -d` wants to pull all eight fields left by one column and the three trailing comments left by two.

**Why it matters.** It is cosmetic, but it means anyone who runs `gofmt -w` or has format-on-save enabled will produce an unrelated eight-line diff on top of this change, and the next reader cannot tell whether the misalignment was deliberate. Worth settling now while the block is already being touched.

**Suggested change.** Realign only these eight field lines to what `gofmt` produces (each field one space left, the three inline comments back to a single aligned column). Please do not run `gofmt -w` on the whole file: it is already gofmt-dirty on `main` from pre-existing trailing whitespace on lines such as 61, 64 and 207, and reformatting those would bury this PR's real diff in unrelated churn. Fixing the file wholesale is a fine separate `chore:` commit if you want it.


SoulKyu (commented):
@SoulKyu Fourth pass, and all three findings from the previous round are genuinely closed on this head: `invalidateLocked` now backs both `InvalidateCache` and the mid-fetch guard (with the backdated `lastAccess` correctly chosen over a delete, so a cold session's generation counter cannot reset to 0 and let a pre-mutation snapshot republish itself), `TestHiddenAlertsServiceLoadUserDataDropsSnapshotHiddenMidFetchOnColdEntry` pins the exact interleaving that was missing, the stub-based call counters replaced the panic assertions, and the constant is now `hiddenAlertsCacheTTL`.

Verified locally on 07a8277: `go build -tags "nogui,webui" ./...` clean, `go vet ./internal/webui/services/` clean, `go test -race -count=1 ./internal/webui/services/` green. I also re-walked the generation protocol against every mutation path and could not find a hole: the three cache-touching mutations bump the counter, the two rule mutations route through `InvalidateCache` which also bumps it, the impersonation guard matches the check `client.BackendClient` uses at `backend_client.go:1017`, and `sweepIdleSessionsLocked` deleting a counter mid-fetch only ever makes the guard more conservative. The interface seam is safe for the single production caller at `internal/webui/router.go:105`, which passes a `backendClient` that is always non-nil.

Two non-blocking items inline: one real coverage gap I confirmed by mutation testing, and one formatting nit. Nice work getting this one tight over the rounds — the failure-path and race reasoning in the comments is the kind of thing that saves the next reader a lot of time.




SoulKyu (commented):
@SoulKyu Fifth pass over the hidden-alerts TTL cache, and this head closes out the two remaining threads from the previous round.

What I verified on 6bef66d:

- **The success-path freshness write is now load-bearing.** `TestHiddenAlertsServiceLoadUserDataCachesAfterSuccessfulFetch` drives `LoadUserData` twice through the public API on a healthy stub and asserts one backend call. I re-ran the exact mutation from the previous round — replacing the `if hiddenAlertsErr == nil && hiddenRulesErr == nil` block at `internal/webui/services/hidden_alerts_service.go:167-171` with an unconditional backdate, which reverts the whole PR — and the suite now fails on that test with "got 2 backend calls, want 1". The regression that would silently reopen issue #78 is caught.
- **The gofmt nit is settled.** `gofmt -d internal/webui/services/hidden_alerts_service.go` no longer touches the `HiddenAlertsService` struct block; the only remaining hunks are the pre-existing trailing-whitespace lines outside this PR's ranges, which you correctly left alone.
- **Cache-coherence invariants hold.** I traced the states the freshness gate can observe: a successful fetch always publishes both alerts and rules together, the error path publishes empty collections and backdates `lastAccess`, and `invalidateLocked` deletes the snapshot before backdating — so "fresh timestamp with a half-populated entry" is unreachable, and `IsAlertHidden`'s `userHiddenAlerts[sessionID] != nil` lazy-reload check can never latch onto a rules-less entry. The generation guard covers all four mutation paths (hide, unhide, clear-all, and save/remove-rule via `InvalidateCache`), and `isImpersonating` mirrors the check in `client.BackendClient.HideAlert` at `internal/webui/client/backend_client.go:1017` exactly.
- **Build and tests.** `go build -tags "nogui,webui" ./...` clean, `go vet` on the package clean, and `go test -race -count=1 ./internal/webui/services/` green.

Really nice job working this one down — the change started as a one-line TTL gate and ended up surfacing four genuine coherence bugs, and you fixed each with the narrower option rather than the convenient one. Backdating `lastAccess` in `invalidateLocked` instead of deleting it, so the idle sweep can still reclaim the generation counter, is the kind of fix that is easy to get subtly wrong and you got it right, with the reasoning left in the comment for the next reader. Happy to see this merge.




internal/webui/services/session_cache_cleanup_test.go
SoulKyu: @SoulKyu This has been fixed in 6bef66d.

TestHiddenAlertsServiceLoadUserDataCachesAfterSuccessfulFetch was added in internal/webui/services/session_cache_cleanup_test.go:110 — two loads on a healthy stubBackend asserting a single fetch. Verified load-bearing: replacing the success branch with the unconditional backdate makes it fail with 'got 2 backend calls, want 1'.




internal/webui/services/hidden_alerts_service.go
SoulKyu: @SoulKyu This has been fixed in 6bef66d.

The HiddenAlertsService field block at internal/webui/services/hidden_alerts_service.go:39-47 was realigned to gofmt output; gofmt -d on the file now reports no diff in that block, only the pre-existing trailing whitespace elsewhere which was deliberately left untouched.




SoulKyu:
## 🧪 QA Report

Env: `make docker-build-all` + `docker compose up -d` (postgres/alertmanager/backend/webui). gRPC fetches counted by running postgres with `log_statement=all` and counting `execute … SELECT * FROM "user_hidden_alerts"` / `"user_hidden_rules"` — one per backend RPC. Counters below are `alerts/rules`.

|---|---|---|
| Two consecutive `GET /dashboard/data` within the TTL ⇒ **one** pair of RPCs, not two | ✅ | Fresh session: cold req#1 → `2/2`→`3/3`. 7 more `/dashboard/data` + 5 `/dashboard/incremental` within 30 s → still `3/3` (13 requests, 1 fetch). After `sleep 31` the next req → `4/4`, and immediately again → `4/4` — TTL, not a permanent cache. |
| Hiding an alert reflected in the very next `/dashboard/data`, no TTL wait | ✅ | Cache warm (`4/4`), `POST /api/v1/dashboard/hidden-alerts` fp `ebe5b4cd…`; next `GET /dashboard/data` <1 s later: 15→14 alerts, fingerprint absent, counters still `4/4` (served from the in-place-updated cache — no refetch needed). `DELETE` unhide → alert back on the next request. |
| Deleting / disabling a hidden rule reflected on the next request | ✅ | Rule `alertname=HighCPUUsage` created → next request: HighCPUUsage 2→0 alerts, counters `5/5`→`6/6` (`InvalidateCache` → refetch). `DELETE` rule → next request: HighCPUUsage 0→2, counters `9/10`. ⚠️ *Disable* (`PUT enabled=false`) also invalidates + refetches correctly, but never takes effect: `internal/backend/database/gorm_db.go:736` uses `Updates(rule)` with a struct, so GORM drops the `false` zero-value and the DB still reads `is_enabled=t` after the PUT. **Pre-existing backend bug, untouched by this PR** — the cache faithfully mirrors what the DB returns. |
| First request for an unknown session still loads (cold path) | ✅ | Brand-new login (new sessionID): counters `14/17`→`15/18` on its first `/dashboard/data`, and the alert that user had hidden was already excluded from that very first response. |
| Regexes compiled once per load, not once per request | ✅ | Inserted a rule with an invalid pattern `[unclosed` (`is_regex=t`) straight into postgres, so every compile logs. webui log `Failed to compile regex for rule qainvalidregex…`: **1** after 8 `/dashboard/data` requests; **2** after 6 further requests following a TTL expiry. One compile per load. |
| `go build -tags "nogui,webui" ./...` + `go vet ./...` clean, hidden-alerts tests pass | ✅ | build exit 0; `go vet ./...` exit 0, no findings (the `router.go:52` finding mentioned in the PR body did not reproduce); `go test ./internal/webui/services/` → 57 passed. |

**Exploratory:** Playwright headless against the running UI — login, table renders (13 rows), search `DiskSpaceLow` filters to 5 all-matching rows and clearing restores, 5 consecutive reloads (every one a cache hit) render normally, hide → row gone from Classic and present in the Hidden tab, unhide → row back; 0 page errors. Two 401s fire at page load (`/api/v1/profile/timezone`, `/api/impersonate/status`) but both return 200 with the same session via curl — a pre-existing page-load race, unrelated to this change.

**Verdict: PASS**

PR #84, opened 2026-07-25
perf(alertmanager): fan out multi-Alertmanager polls concurrently

### LINKED ISSUE #80: perf(alertmanager): FetchAllAlertsDetailed polls every Alertmanager serially — one slow source stalls the whole refresh cycle
## Problem

The multi-Alertmanager fan-out is a plain sequential loop. `internal/alertmanager/client.go:674-697`:

```go
func (mc *MultiClient) FetchAllAlertsDetailed() ([]AlertWithSource, map[string]error) {
	mc.mutex.RLock()
	defer mc.mutex.RUnlock()

	for name, client := range mc.clients {
		alerts, err := client.FetchAlerts()   // blocking HTTP GET /api/v2/alerts
		if err != nil {
			failedSources[name] = err
			continue
		}
		...
	}
}
```

Each `FetchAlerts` is a blocking `GET /api/v2/alerts` with a **10 s** client timeout (`internal/alertmanager/client.go:204`, `:255`, `:261`, `:270`) and no retry. So the wall-clock cost of one fan-out is the **sum** of every source's latency, and an unreachable source contributes its full 10 s timeout.

This function is the single input to the whole live dashboard. `AlertCache.refreshAlerts` calls it first thing (`internal/webui/services/alert_cache.go:178`), driven by a ticker whose default period is **10 s** (`config/config.go:191`, wired at `internal/webui/router.go:95`). Everything downstream — the fingerprint diff, `CaptureAlertFired`, resolve detection, and the SSE broadcast at `alert_cache.go:279-287` — waits on it.

Concrete failure path: 3 Alertmanagers configured, each answering in ~400 ms, and one of them becomes unreachable (network partition, or a hung Mimir/Cortex tenant endpoint).

1. Tick fires, `refreshAlerts` calls `FetchAllAlertsDetailed`.
2. The loop iterates in Go's randomised map order. It blocks 10 s on the dead source, plus ~800 ms for the two healthy ones.
3. `refreshAlerts` returns after ~10.8 s. `backgroundRefresh` (`alert_cache.go:148-161`) only reads the next tick after the current refresh returns, so intervening ticks are coalesced.
4. The effective refresh period silently becomes ~11 s instead of 10 s — and with two dead sources, ~21 s.

Even with everything healthy, N sources at ~400 ms each turn a 400 ms fan-out into an N×400 ms one, for work that has no data dependency between sources.

The same serial shape repeats across every fan-out in the file: `FetchAllSilences` (`:727-745`), `TestAllConnections` (`:762-764`), `GetHealthStatus` (`:833-843`), `GetHealthyClients` (`:845-857`). `GetHealthStatus`/`GetHealthyClients` are worse than they look, because `IsHealthy` → `TestConnection` (`:576-598`) fetches the **full alerts payload** from each instance and, on the success path, never reads the body — so the response cannot be reused by keep-alive.

A secondary effect of the same code: `mc.mutex.RLock()` is held (deferred) across the entire serial fan-out, so `UpdateFromConfig` (`:881`, needs `Lock()`) can be blocked for the full N×timeout window.

## Impact

Multi-Alertmanager support is the product's headline feature, and this is the code path that makes it slow precisely when it matters. Every user of every open tab sees stale alerts for the duration: SSE events for new alerts, resolutions, and status changes are only emitted after the slowest source answers. During an incident where one Alertmanager is degraded, alert delivery for the *healthy* ones is delayed too — the exact opposite of what an operator needs.

It also inflates the statistics record: `CaptureAlertFired` runs off the diff computed after the fan-out (`alert_cache.go:212-218`), so `fired_at` timestamps drift by the accumulated fan-out latency.

## Proposed approach

Fan out concurrently — the loop bodies are independent and the result assembly is trivial:

1. In `FetchAllAlertsDetailed`, snapshot `mc.clients` into a local slice under `RLock`, **release the lock**, then run one goroutine per client with a `sync.WaitGroup`, collecting `[]AlertWithSource` and `failedSources` under a small local mutex (or via per-index result slots, which avoids the mutex entirely). Releasing the config lock before the I/O also unblocks `UpdateFromConfig`.
2. Preserve today's partial-failure contract exactly: sources that error land in `failedSources` and their alerts are omitted. `refreshAlerts` already depends on this to avoid mass-resolving an unreachable source (`alert_cache.go:236-238`) — that behaviour must not change.
3. Apply the same treatment to `FetchAllSilences`, `TestAllConnections`, `GetHealthStatus`, `GetHealthyClients`.
4. Separately, make `TestConnection` cheap: hit a lighter endpoint or use `HEAD`, and drain/close the body on the success path so connections are reusable. Fetching every alert to answer "is this up?" is wasted work on both ends.
5. No new dependency needed — `sync.WaitGroup` covers it. Bounding concurrency is unnecessary here: the fleet size is the number of configured Alertmanagers, which is small by construction.

## Acceptance criteria

- [ ] `FetchAllAlertsDetailed` wall-clock time for N healthy sources is ~max(latency), not ~sum(latency). Verifiable with a table test using `httptest` servers that sleep different durations.
- [ ] With one source hanging past its 10 s timeout and two healthy, the healthy sources' alerts are returned in ~their own latency and the dead one appears in `failedSources`; total time is bounded by the single timeout, not timeout + others.
- [ ] Partial-failure semantics unchanged: a failing source yields an entry in `failedSources` and no alerts, so `refreshAlerts` still leaves its cached alerts untouched instead of mass-resolving them (`alert_cache.go:236-238` behaviour preserved — the existing `alert_cache_test.go` coverage for this must still pass).
- [ ] `mc.mutex` is not held across any HTTP call in the fan-out functions.
- [ ] `FetchAllSilences`, `TestAllConnections`, `GetHealthStatus`, `GetHealthyClients` are concurrent too, with unchanged return shapes.
- [ ] `TestConnection` drains/closes the response body on the success path.
- [ ] `go build -tags "nogui,webui" ./...`, `go vet ./...`, and `go test -race ./internal/alertmanager/... ./internal/webui/services/...` all clean.

SoulKyu:
The multi-Alertmanager fan-outs in `internal/alertmanager/client.go` were plain sequential loops, so a refresh cycle cost `sum(latency)` across sources and an unreachable Alertmanager contributed its full 10s client timeout. `AlertCache.refreshAlerts` is the only caller that matters — it runs off a 10s ticker and coalesces ticks while a refresh is in flight, so one dead source stretched the effective refresh period to ~21s and delayed SSE delivery for the *healthy* sources too.

## Changes

- `snapshotClients()` copies `mc.clients` under `RLock` and releases it before any I/O; a generic `fanOut()` helper runs one goroutine per client with a `sync.WaitGroup` and writes results into per-index slots (no extra mutex). Concurrency is deliberately unbounded — the fleet is the number of configured Alertmanagers.
- Converted `FetchAllAlertsDetailed`, `FetchAllSilences`, `TestAllConnections`, `GetHealthStatus` and `GetHealthyClients`. Return shapes are unchanged.
- Partial-failure contract preserved exactly: a failing source yields an entry in `failedSources` and contributes no alerts, so `refreshAlerts` still leaves its cached alerts untouched instead of mass-resolving them.
- `mc.mutex` is no longer held across any HTTP call, so `UpdateFromConfig` can no longer block for the whole N x timeout window.
- `TestConnection` probes `/api/v2/alerts` with a filter that cannot match (keeps compatibility with Cortex/Mimir, which do not all expose `/api/v2/status`) and drains the body before closing, so the health check stops pulling the full alerts payload and the connection returns to the keep-alive pool.
- Drive-by: `log.Fatalf("... %w", err)` in `internal/webui/router.go` was failing `go vet ./...`; switched to `%v`.

## Tests

New `internal/alertmanager/client_test.go` with `httptest` servers:

- table test asserting the fan-out wall-clock is ~max(latency), not ~sum(latency), for 2 and 5 sources
- one hanging source plus two healthy: healthy alerts returned, dead source in `failedSources`, total time bounded by the single timeout
- `UpdateFromConfig` completes while a fan-out is in flight
- `TestConnection` sends the health filter and reuses the connection on a second probe
- `GetHealthStatus`/`GetHealthyClients` concurrency and correct name to client mapping

`go build -tags "nogui,webui" ./...`, `go vet ./...` and `go test -race ./internal/alertmanager/... ./internal/webui/services/...` all clean (53 tests).

Closes #80

### DIFF
--- a/internal/alertmanager/client.go
+++ b/internal/alertmanager/client.go
@@ -585,6 +585,11 @@ func (c *Client) DeleteSilence(silenceID string) error {
 // of the full payload.
 const healthCheckFilter = `alertname="__notificator_health_check__"`
 
+// healthCheckDrainLimit caps how much of the probe response we read. A backend
+// that honours the filter answers with a handful of bytes; one that silently
+// ignores it would otherwise have us pull the entire alert set on every probe.
+const healthCheckDrainLimit = 4 << 10
+
 func (c *Client) TestConnection() error {
 	url := fmt.Sprintf("%s/api/v2/alerts?filter=%s", c.BaseURL, neturl.QueryEscape(healthCheckFilter))
 
@@ -606,8 +611,10 @@ func (c *Client) TestConnection() error {
 		return fmt.Errorf("alertmanager returned status %d: %s", resp.StatusCode, string(body[:min(200, len(body))]))
 	}
 
-	// Drain the body so the connection goes back to the keep-alive pool.
-	_, _ = io.Copy(io.Discard, resp.Body)
+	// Drain a bounded amount so the connection goes back to the keep-alive pool
+	// without downloading a full payload from a backend that ignored the filter.
+	// Past that ceiling, letting the transport abandon the connection is cheaper.
+	_, _ = io.CopyN(io.Discard, resp.Body, healthCheckDrainLimit)
 
 	return nil
 }
--- a/internal/alertmanager/client.go
+++ b/internal/alertmanager/client.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	neturl "net/url"
 	"strings"
 	"sync"
 	"time"
@@ -578,8 +579,14 @@ func (c *Client) DeleteSilence(silenceID string) error {
 	return nil
 }
 
+// healthCheckFilter keeps the health probe cheap: the alerts endpoint is the
+// only one guaranteed to exist across Alertmanager, Cortex and Mimir, but a
+// filter that cannot match anything makes it answer with an empty list instead
+// of the full payload.
+const healthCheckFilter = `alertname="__notificator_health_check__"`
+
 func (c *Client) TestConnection() error {
-	url := fmt.Sprintf("%s/api/v2/alerts", c.BaseURL) // v2 API doesn't have dedicated status endpoint
+	url := fmt.Sprintf("%s/api/v2/alerts?filter=%s", c.BaseURL, neturl.QueryEscape(healthCheckFilter))
 
 	req, err := http.NewRequest("GET", url, nil)
 	if err != nil {
@@ -598,6 +605,9 @@ func (c *Client) TestConnection() error {
 		body, _ := io.ReadAll(resp.Body)
 		return fmt.Errorf("alertmanager returned status %d: %s", resp.StatusCode, string(body[:min(200, len(body))]))
 	}
+
+	// Drain the body so the connection goes back to the keep-alive pool.
+	_, _ = io.Copy(io.Discard, resp.Body)
 
 	return nil
 }
@@ -673,27 +683,76 @@ type SilenceWithSource struct {
 	Source  string // Name of the Alertmanager instance
 }
 
+type namedClient struct {
+	name   string
+	client *Client
+}
+
+type fanOutResult[T any] struct {
+	name  string
+	value T
+	err   error
+}
+
+// snapshotClients copies the client map under RLock so fan-outs can do their
+// HTTP work without holding mc.mutex, which would otherwise block
+// UpdateFromConfig for the whole N x timeout window.
+func (mc *MultiClient) snapshotClients() []namedClient {
+	mc.mutex.RLock()
+	defer mc.mutex.RUnlock()
+
+	snapshot := make([]namedClient, 0, len(mc.clients))
+	for name, client := range mc.clients {
+		snapshot = append(snapshot, namedClient{name: name, client: client})
+	}
+
+	return snapshot
+}
+
+// fanOut calls fn against every client concurrently and returns the results in
+// snapshot order, so the wall-clock cost is max(latency) rather than
+// sum(latency). Concurrency is unbounded on purpose: the fleet size is the
+// number of configured Alertmanagers, which is small by construction.
+func fanOut[T any](clients []namedClient, fn func(*Client) (T, error)) []fanOutResult[T] {
+	results := make([]fanOutResult[T], len(clients))
+
+	var wg sync.WaitGroup
+	for i, nc := range clients {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			value, err := fn(nc.client)
+			results[i] = fanOutResult[T]{name: nc.name, value: value, err: err}
+		}()
+	}
+	wg.Wait()
+
+	return results
+}
+
+func testConnections(clients []namedClient) []fanOutResult[struct{}] {
+	return fanOut(clients, func(c *Client) (struct{}, error) {
+		return struct{}{}, c.TestConnection()
+	})
+}
+
 // FetchAllAlertsDetailed fetches alerts from every configured Alertmanager and
 // reports per-source failures instead of collapsing them into a single error,
 // so callers can tell a partial fetch from a genuinely empty one.
 func (mc *MultiClient) FetchAllAlertsDetailed() ([]AlertWithSource, map[string]error) {
-	mc.mutex.RLock()
-	defer mc.mutex.RUnlock()
-
 	var allAlerts []AlertWithSource
 	failedSources := make(map[string]error)
 
-	for name, client := range mc.clients {
-		alerts, err := client.FetchAlerts()
-		if err != nil {
-			failedSources[name] = err
+	for _, result := range fanOut(mc.snapshotClients(), (*Client).FetchAlerts) {
+		if result.err != nil {
+			failedSources[result.name] = result.err
 			continue
 		}
 
-		for _, alert := range alerts {
+		for _, alert := range result.value {
 			allAlerts = append(allAlerts, AlertWithSource{
 				Alert:  alert,
-				Source: name,
+				Source: result.name,
 			})
 		}
 	}
@@ -733,23 +792,19 @@ func (mc *MultiClient) FetchAllActiveAlerts() ([]AlertWithSource, error) {
 // reports per-source failures instead of collapsing them into a single error, so an
 // unreachable Alertmanager degrades that source instead of blanking the inventory.
 func (mc *MultiClient) FetchAllSilencesDetailed() ([]SilenceWithSource, map[string]error) {
-	mc.mutex.RLock()
-	defer mc.mutex.RUnlock()
-
 	var allSilences []SilenceWithSource
 	failedSources := make(map[string]error)
 
-	for name, client := range mc.clients {
-		silences, err := client.FetchSilences()
-		if err != nil {
-			failedSources[name] = err
+	for _, result := range fanOut(mc.snapshotClients(), (*Client).FetchSilences) {
+		if result.err != nil {
+			failedSources[result.name] = result.err
 			continue
 		}
 
-		for _, silence := range silences {
+		for _, silence := range result.value {
 			allSilences = append(allSilences, SilenceWithSource{
 				Silence: silence,
-				Source:  name,
+				Source:  result.name,
 			})
 		}
 	}
@@ -758,13 +813,10 @@ func (mc *MultiClient) FetchAllSilencesDetailed() ([]SilenceWithSource, map[stri
 }
 
 func (mc *MultiClient) TestAllConnections() map[string]error {
-	mc.mutex.RLock()
-	defer mc.mutex.RUnlock()
-
 	results := make(map[string]error)
 
-	for name, client := range mc.clients {
-		results[name] = client.TestConnection()
+	for _, result := range testConnections(mc.snapshotClients()) {
+		results[result.name] = result.err
 	}
 
 	return results
@@ -835,25 +887,21 @@ func (c *Client) IsHealthy() bool {
 }
 
 func (mc *MultiClient) GetHealthStatus() map[string]bool {
-	mc.mutex.RLock()
-	defer mc.mutex.RUnlock()
-
 	status := make(map[string]bool)
-	for name, client := range mc.clients {
-		status[name] = client.IsHealthy()
+	for _, result := range testConnections(mc.snapshotClients()) {
+		status[result.name] = result.err == nil
 	}
 
 	return status
 }
 
 func (mc *MultiClient) GetHealthyClients() map[string]*Client {
-	mc.mutex.RLock()
-	defer mc.mutex.RUnlock()
+	clients := mc.snapshotClients()
 
 	healthy := make(map[string]*Client)
-	for name, client := range mc.clients {
-		if client.IsHealthy() {
-			healthy[name] = client
+	for i, result := range testConnections(clients) {
+		if result.err == nil {
+			healthy[result.name] = clients[i].client
 		}
 	}
 
--- a/internal/alertmanager/client_test.go
+++ b/internal/alertmanager/client_test.go
@@ -91,28 +91,34 @@ func TestFetchAllAlertsDetailedIsConcurrent(t *testing.T) {
 }
 
 func TestFetchAllAlertsDetailedPartialFailure(t *testing.T) {
-	// A source whose handler outlives the client timeout stands in for an
-	// unreachable Alertmanager.
+	// Sources whose handlers outlive the client timeout stand in for unreachable
+	// Alertmanagers. Two of them, so serialising stacks both timeouts on top of
+	// the healthy latencies (~1.1s) and trips the bound below, while a concurrent
+	// fan-out stays at roughly one timeout.
 	hang := make(chan struct{})
-	dead := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+	tarpit := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		<-hang
-	}))
+	})
+	deadA := httptest.NewServer(tarpit)
+	deadB := httptest.NewServer(tarpit)
 	defer func() {
 		close(hang)
-		dead.Close()
+		deadA.Close()
+		deadB.Close()
 	}()
 
 	urls := map[string]string{
 		"healthy-a": alertmanagerStub(t, "healthy-a", 50*time.Millisecond).URL,
 		"healthy-b": alertmanagerStub(t, "healthy-b", 50*time.Millisecond).URL,
-		"dead":      dead.URL,
+		"dead-a":    deadA.URL,
+		"dead-b":    deadB.URL,
 	}
 
 	const deadTimeout = 500 * time.Millisecond
 
 	mc := multiClientFor(urls)
 	for _, nc := range mc.snapshotClients() {
-		if nc.name == "dead" {
+		if strings.HasPrefix(nc.name, "dead") {
 			nc.client.HTTPClient.Timeout = deadTimeout
 		}
 	}
@@ -125,19 +131,21 @@ func TestFetchAllAlertsDetailedPartialFailure(t *testing.T) {
 		t.Fatalf("got %d alerts, want the 2 healthy ones", len(alerts))
 	}
 	for _, a := range alerts {
-		if a.Source == "dead" {
+		if strings.HasPrefix(a.Source, "dead") {
 			t.Fatalf("dead source contributed alerts: %+v", a)
 		}
 	}
-	if _, ok := failed["dead"]; !ok {
-		t.Fatalf("dead source missing from failedSources: %v", failed)
+	for _, name := range []string{"dead-a", "dead-b"} {
+		if _, ok := failed[name]; !ok {
+			t.Fatalf("%s missing from failedSources: %v", name, failed)
+		}
 	}
-	if len(failed) != 1 {
+	if len(failed) != 2 {
 		t.Fatalf("unexpected failures: %v", failed)
 	}
-	// Bounded by the single timeout, not timeout + healthy latencies. The bound
-	// is derived from the timeout so the intent stays "did not serialise" rather
-	// than a hand-tuned number that a slow runner can trip.
+	// Bounded by a single timeout, not both timeouts plus the healthy latencies.
+	// The bound is derived from the timeout so the intent stays "did not
+	// serialise" rather than a hand-tuned number that a slow runner can trip.
 	if elapsed > 2*deadTimeout {
 		t.Fatalf("fan-out took %s, want it bounded by the dead source timeout %s", elapsed, deadTimeout)
 	}
--- a/internal/alertmanager/client_test.go
+++ b/internal/alertmanager/client_test.go
@@ -4,12 +4,17 @@ import (
 	"fmt"
 	"net/http"
 	"net/http/httptest"
+	"strings"
 	"sync"
 	"testing"
 	"time"
 
 	"notificator/config"
 )
+
+func alertJSON(name string) string {
+	return fmt.Sprintf(`[{"fingerprint":"fp-%s","labels":{"alertname":"%s"},"status":{"state":"active"}}]`, name, name)
+}
 
 // alertmanagerStub serves a single alert named after the instance, after delay.
 func alertmanagerStub(t *testing.T, name string, delay time.Duration) *httptest.Server {
@@ -18,7 +23,24 @@ func alertmanagerStub(t *testing.T, name string, delay time.Duration) *httptest.
 	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		time.Sleep(delay)
 		w.Header().Set("Content-Type", "application/json")
-		fmt.Fprintf(w, `[{"fingerprint":"fp-%s","labels":{"alertname":"%s"},"status":{"state":"active"}}]`, name, name)
+		fmt.Fprint(w, alertJSON(name))
+	}))
+	t.Cleanup(srv.Close)
+
+	return srv
+}
+
+// barrierStub only answers once every peer stub is also serving, so a fan-out
+// that regressed to a sequential loop deadlocks on the first source and fails
+// on its client timeout instead of merely being slow.
+func barrierStub(t *testing.T, name string, barrier *sync.WaitGroup) *httptest.Server {
+	t.Helper()
+
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		barrier.Done()
+		barrier.Wait()
+		w.Header().Set("Content-Type", "application/json")
+		fmt.Fprint(w, alertJSON(name))
 	}))
 	t.Cleanup(srv.Close)
 
@@ -35,8 +57,6 @@ func multiClientFor(urls map[string]string) *MultiClient {
 }
 
 func TestFetchAllAlertsDetailedIsConcurrent(t *testing.T) {
-	const delay = 300 * time.Millisecond
-
 	tests := []struct {
 		name    string
 		sources int
@@ -47,25 +67,24 @@ func TestFetchAllAlertsDetailedIsConcurrent(t *testing.T) {
 
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
+			// Every source must be in flight at once for any of them to answer,
+			// so this asserts observed concurrency rather than wall-clock.
+			var barrier sync.WaitGroup
+			barrier.Add(tc.sources)
+
 			urls := make(map[string]string, tc.sources)
 			for i := 0; i < tc.sources; i++ {
 				name := fmt.Sprintf("am-%d", i)
-				urls[name] = alertmanagerStub(t, name, delay).URL
+				urls[name] = barrierStub(t, name, &barrier).URL
 			}
 
-			start := time.Now()
 			alerts, failed := multiClientFor(urls).FetchAllAlertsDetailed()
-			elapsed := time.Since(start)
 
 			if len(failed) != 0 {
-				t.Fatalf("unexpected failures: %v", failed)
+				t.Fatalf("sources did not run concurrently (a serial fan-out stalls on the barrier): %v", failed)
 			}
 			if len(alerts) != tc.sources {
 				t.Fatalf("got %d alerts, want %d", len(alerts), tc.sources)
-			}
-			// Serial would be sources*delay; concurrent stays near one delay.
-			if max := time.Duration(float64(delay) * 1.8); elapsed > max {
-				t.Fatalf("fan-out took %s, want under %s (serial would be %s)", elapsed, max, time.Duration(tc.sources)*delay)
 			}
 		})
 	}
@@ -122,7 +141,12 @@ func TestFetchAllAlertsDetailedPartialFailure(t *testing.T) {
 
 func TestFanOutDoesNotHoldMutexAcrossHTTP(t *testing.T) {
 	release := make(chan struct{})
+	// serving proves the fan-out is inside the HTTP call before the writer races
+	// it; without it the write lock may be taken before the fetch ever starts and
+	// the test passes regardless of where the mutex is held.
+	serving := make(chan struct{})
 	slow := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		close(serving)
 		<-release
 		fmt.Fprint(w, `[]`)
 	}))
@@ -130,12 +154,18 @@ func TestFanOutDoesNotHoldMutexAcrossHTTP(t *testing.T) {
 
 	mc := multiClientFor(map[string]string{"slow": slow.URL})
 
-	var wg sync.WaitGroup
-	wg.Add(1)
-	go func() {
-		defer wg.Done()
-		mc.FetchAllAlertsDetailed()
-	}()
+	// Deferred so a t.Fatal below still unblocks the handler, letting the
+	// deferred slow.Close() return; otherwise a regression reads as a hung
+	// package instead of a failed assertion.
+	defer close(release)
+
+	go mc.FetchAllAlertsDetailed()
+
+	select {
+	case <-serving:
+	case <-time.After(5 * time.Second):
+		t.Fatal("fan-out never reached the HTTP handler")
+	}
 
 	// UpdateFromConfig needs the write lock; it must not wait on the in-flight fetch.
 	updated := make(chan struct{})
@@ -149,15 +179,17 @@ func TestFanOutDoesNotHoldMutexAcrossHTTP(t *testing.T) {
 	case <-time.After(2 * time.Second):
 		t.Fatal("UpdateFromConfig blocked behind an in-flight fan-out")
 	}
-
-	close(release)
-	wg.Wait()
 }
 
-func TestTestConnectionDrainsBody(t *testing.T) {
+func TestTestConnectionUsesHealthFilterAndReusesConnection(t *testing.T) {
+	var mu sync.Mutex
 	var filter string
+	var addrs []string
 	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		mu.Lock()
 		filter = r.URL.Query().Get("filter")
+		addrs = append(addrs, r.RemoteAddr)
+		mu.Unlock()
 		fmt.Fprint(w, `[]`)
 	}))
 	defer srv.Close()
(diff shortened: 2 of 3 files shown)


internal/alertmanager/client.go
```diff
@@ -594,6 +601,9 @@ func (c *Client) TestConnection() error {
 		return fmt.Errorf("alertmanager returned status %d: %s", resp.StatusCode, string(body[:min(200, len(body))]))
 	}
 
+	// Drain the body so the connection goes back to the keep-alive pool.
+	_, _ = io.Copy(io.Discard, resp.Body)
```
SoulKyu: **Bound the drain — against a server that ignores `filter`, this probe now downloads the entire alerts payload where the old one downloaded nothing.**

Problem: before this change `TestConnection` never read the response body — `defer resp.Body.Close()` on an unread response makes `net/http` abandon the connection rather than pull the remainder. The new `io.Copy(io.Discard, resp.Body)` reads to EOF with no ceiling.

Why it matters: on a real Alertmanager the `filter` added on line 584 makes the response `[]`, so the drain is free and the keep-alive win is real. But any endpoint that accepts the request and silently ignores an unrecognised `filter` — a reverse proxy that strips the query string, a compatibility shim in front of Cortex/Mimir, an older API surface — answers 200 with the full alert set, and this probe will now pull all of it. `AlertmanagerHealthCheck` calls `GetHealthStatus()` per HTTP request (internal/webui/handlers/handlers.go:362), which after this PR fans out to every configured source, so the worst case is a full alerts fetch per source per health-check request — precisely the cost this PR set out to remove.

Evidence: line 605 is an unbounded `io.Copy`; the only thing keeping the body small is the assumption on lines 577-581 that every backend honours `filter`, and no test covers a backend that does not.

Suggested change: cap the drain. A few KB is all the transport needs to decide the connection is reusable, and past that abandoning the connection is the cheaper outcome anyway.

```go
// Drain a bounded amount so the connection goes back to the keep-alive pool
// without downloading a full payload from a backend that ignored the filter.
_, _ = io.CopyN(io.Discard, resp.Body, 4<<10)
```

Worth pairing with a stub in `client_test.go` that ignores `filter` and returns a large body, asserting the probe reads only a bounded number of bytes.


internal/alertmanager/client_test.go
```diff
@@ -0,0 +1,218 @@
...
+			if len(failed) != 0 {
+				t.Fatalf("unexpected failures: %v", failed)
+			}
+			if len(alerts) != tc.sources {
+				t.Fatalf("got %d alerts, want %d", len(alerts), tc.sources)
+			}
+			// Serial would be sources*delay; concurrent stays near one delay.
+			if max := time.Duration(float64(delay) * 1.8); elapsed > max {
+				t.Fatalf("fan-out took %s, want under %s (serial would be %s)", elapsed, max, time.Duration(tc.sources)*delay)
+			}
```
SoulKyu: **Assert observed concurrency instead of wall-clock, otherwise this is a latency test living in the unit suite.**

Problem: `elapsed > delay*1.8` gives a 540 ms budget for a 300 ms stub delay. Two things go wrong with that. It flakes: the package is run under `-race`, and on a throttled or shared CI runner with a low `GOMAXPROCS`, five goroutines plus five `httptest` servers can overshoot 540 ms with the code perfectly correct — and the failure reads as a `fanOut` regression rather than a slow runner. It is also weak: the 2-source case passes with a concurrency of 2 only marginally better than serial, and nothing here distinguishes "fully parallel" from "parallel enough to fit the budget".

Why it matters: a flaky perf assertion in the default test path gets muted or `-short`-skipped after the second false alarm, and the property the PR is actually about stops being guarded.

Evidence: lines 67-69 assert only an upper bound on `time.Since(start)`; there is no assertion anywhere in this test that more than one request was ever in flight.

Suggested change: turn the stubs into a barrier so serial execution deadlocks rather than merely running slowly. All `tc.sources` handlers must be in flight simultaneously to make progress; if `fanOut` regressed to a sequential loop the first handler blocks, the client hits its 10 s timeout, and the test fails deterministically with a clear message.

```go
var barrier sync.WaitGroup
barrier.Add(tc.sources)
// inside the stub handler, before writing the body:
barrier.Done()
barrier.Wait()
```

That needs a variant of `alertmanagerStub` that takes the barrier instead of a `delay`. Keep the wall-clock check afterwards if you want it as a smoke signal, but let the barrier be the assertion that fails the build.


internal/alertmanager/client_test.go
```diff
@@ -0,0 +1,218 @@
...
+	go func() {
+		mc.UpdateFromConfig(&config.Config{})
+		close(updated)
+	}()
+
+	select {
+	case <-updated:
+	case <-time.After(2 * time.Second):
+		t.Fatal("UpdateFromConfig blocked behind an in-flight fan-out")
+	}
```
SoulKyu: **`TestFanOutDoesNotHoldMutexAcrossHTTP` can pass without ever exercising the lock — it would go green on the pre-PR code too.**

Problem: nothing orders the fetch goroutine (lines 135-138) against the `UpdateFromConfig` goroutine (lines 142-145). If the writer reaches `mc.mutex.Lock()` before the fetcher has entered `snapshotClients`, the write lock is uncontended, `updated` closes immediately, and the `select` on lines 147-151 succeeds. That outcome is independent of whether the fan-out holds `mc.mutex` across HTTP.

Why it matters: this is the only test guarding the specific regression the PR fixes — `mc.mutex` no longer held across the N x timeout window. As written it is a coin flip on goroutine start order, so it would also have passed against the old `defer mc.mutex.RUnlock()` implementation. A future refactor that reintroduces the lock-across-I/O would not be caught reliably.

Evidence: the only synchronisation between the two goroutines is the scheduler; `release` (line 124) gates the handler's *return*, not its *entry*, so it cannot be used to establish that the fetch is mid-flight.

Suggested change: have the stub signal that it is actually serving, and wait for that before racing the writer.

```go
serving := make(chan struct{})
slow := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
    close(serving)
    <-release
    fmt.Fprint(w, `[]`)
}))
...
go func() { defer wg.Done(); mc.FetchAllAlertsDetailed() }()
<-serving // fan-out is now inside the HTTP call; if it held mc.mutex, Lock() below would block
go func() { mc.UpdateFromConfig(&config.Config{}); close(updated) }()
```

With the single stub used here `close(serving)` runs at most once; wrap it in a `sync.Once` if you ever add a second source to this test.


internal/alertmanager/client_test.go
```diff
@@ -0,0 +1,218 @@
...
+		t.Fatalf("TestConnection: %v", err)
+	}
+	if filter != healthCheckFilter {
+		t.Fatalf("health probe filter = %q, want %q", filter, healthCheckFilter)
+	}
+
+	// A drained + closed body is reused, so the second probe rides the same connection.
+	if err := client.TestConnection(); err != nil {
+		t.Fatalf("second TestConnection: %v", err)
+	}
```
SoulKyu: **`TestTestConnectionDrainsBody` does not test the drain — the comment on line 173 is the only thing asserting connection reuse.**

Problem: lines 173-176 claim "A drained + closed body is reused, so the second probe rides the same connection", but the assertion is only that the second `TestConnection()` returns nil. That is equally true when the transport discards the first connection and dials a fresh one, which is exactly what happens without the `io.Copy` drain.

Why it matters: deleting `io.Copy(io.Discard, resp.Body)` from internal/alertmanager/client.go:605 leaves this test green, so the drain — one of the two behaviours the PR description credits to this test — is unguarded. The test's name promises coverage that is not there, which is worse than no test, because the next person reads the name and assumes the behaviour is pinned.

Evidence: the only assertions are the `filter` equality check (lines 169-171) and two nil-error checks; nothing observes the transport.

Suggested change: record the peer address in the handler and assert both probes came from the same client port. This works without touching `TestConnection`'s request construction, and it also removes the unsynchronised cross-goroutine write to `filter` on line 160.

```go
var mu sync.Mutex
var addrs []string
srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
    mu.Lock()
    filter = r.URL.Query().Get("filter")
    addrs = append(addrs, r.RemoteAddr)
    mu.Unlock()
    fmt.Fprint(w, `[]`)
}))
...
if addrs[0] != addrs[1] {
    t.Fatalf("probe dialled a new connection instead of reusing one: %v", addrs)
}
```

An `httptrace.ClientTrace` with a `GotConn` hook checking `info.Reused` is the other option, but it needs `TestConnection` to accept a context, so the `RemoteAddr` check is the cheaper one here. Also consider renaming the test to say what it pins, e.g. `TestTestConnectionUsesHealthFilterAndReusesConnection`.


SoulKyu (commented):
@SoulKyu — reviewed the fan-out conversion in `internal/alertmanager/client.go`, the new `internal/alertmanager/client_test.go`, and the `log.Fatalf` verb fix in `internal/webui/router.go`.

The core change holds up. `snapshotClients()` copies under `RLock` and releases before any I/O; `UpdateFromConfig` replaces the whole map with freshly built `*Client` values rather than mutating existing ones, so a snapshotted client is never written underneath an in-flight fan-out; `fanOut` writes into per-index slots, so there is no shared-write hazard and no extra mutex needed. `IsHealthy()` was already just `TestConnection() == nil`, so routing `GetHealthStatus`/`GetHealthyClients` through `testConnections` preserves the previous semantics, and `GetHealthyClients` indexing `clients[i]` is safe because `fanOut` returns results in snapshot order. `go.mod` is at Go 1.23, so the per-iteration loop variables the goroutine closure relies on are guaranteed. Locally `go vet ./internal/alertmanager/... ./internal/webui/...` and `go test -race -count=3 ./internal/alertmanager/...` are both clean.

Four non-blocking items, all inline: one on the unbounded body drain in `TestConnection`, and three on the new tests (a wall-clock assertion that will flake, a lock test that can pass without exercising the lock, and a connection-reuse claim that nothing asserts). None of these block merge.



internal/alertmanager/client.go
SoulKyu: @SoulKyu This has been fixed in d88827c.

Replaced the unbounded io.Copy with io.CopyN(io.Discard, resp.Body, healthCheckDrainLimit) at internal/alertmanager/client.go:611, with the 4 KiB ceiling as a named const at line 584. Also added TestTestConnectionBoundsDrainOfIgnoredFilter, a stub that ignores the filter and returns 256 KiB: the probe must leave the connection unreusable, which fails if the cap is reverted to io.Copy.




internal/alertmanager/client_test.go
SoulKyu: @SoulKyu This has been fixed in d88827c.

Dropped the wall-clock budget and added barrierStub (client_test.go:36), which calls barrier.Done() then barrier.Wait() so all tc.sources handlers must be in flight before any responds. Verified by mutation: making fanOut sequential fails both subtests deterministically on the client timeout with a clear message.




internal/alertmanager/client_test.go
SoulKyu: @SoulKyu This has been fixed in d88827c.

Added a serving channel closed by the handler and awaited before starting the UpdateFromConfig goroutine, so the writer only races a fetch already inside the HTTP call. Also moved the handler release into a defer so a regression fails at 2s instead of hanging on slow.Close(). Verified by mutation: holding mc.mutex across the fan-out now fails with 'UpdateFromConfig blocked behind an in-flight fan-out'.




internal/alertmanager/client_test.go
SoulKyu: @SoulKyu This has been fixed in d88827c.

Renamed to TestTestConnectionUsesHealthFilterAndReusesConnection, recorded r.RemoteAddr per probe under a mutex (which also removes the unsynchronised filter write), and asserted addrs[0] == addrs[1]. Verified by mutation: deleting the drain from client.go makes the second probe dial a new port and the test fails.




SoulKyu:
**Looper fixer round complete** — d88827c

- ✅ Review comment on `internal/alertmanager/client.go` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/84#discussion_r3649544795)
  - Replaced the unbounded io.Copy with io.CopyN(io.Discard, resp.Body, healthCheckDrainLimit) at internal/alertmanager/client.go:611, with the 4 KiB ceiling as a named const at line 584. Also added TestTestConnectionBoundsDrainOfIgnoredFilter, a stub that ignores the filter and returns 256 KiB: the probe must leave the connection unreusable, which fails if the cap is reverted to io.Copy.
- ✅ Review comment on `internal/alertmanager/client_test.go` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/84#discussion_r3649544797)
  - Dropped the wall-clock budget and added barrierStub (client_test.go:36), which calls barrier.Done() then barrier.Wait() so all tc.sources handlers must be in flight before any responds. Verified by mutation: making fanOut sequential fails both subtests deterministically on the client timeout with a clear message.
- ✅ Review comment on `internal/alertmanager/client_test.go` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/84#discussion_r3649544799)
  - Added a serving channel closed by the handler and awaited before starting the UpdateFromConfig goroutine, so the writer only races a fetch already inside the HTTP call. Also moved the handler release into a defer so a regression fails at 2s instead of hanging on slow.Close(). Verified by mutation: holding mc.mutex across the fan-out now fails with 'UpdateFromConfig blocked behind an in-flight fan-out'.
- ✅ Review comment on `internal/alertmanager/client_test.go` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/84#discussion_r3649544801)
  - Renamed to TestTestConnectionUsesHealthFilterAndReusesConnection, recorded r.RemoteAddr per probe under a mutex (which also removes the unsynchronised filter write), and asserted addrs[0] == addrs[1]. Verified by mutation: deleting the drain from client.go makes the second probe dial a new port and the test fails.


SoulKyu:

Rebased onto `origin/main`. One conflict, in `internal/alertmanager/client.go`.

**Conflict:** `MultiClient.FetchAllSilences`
- `main` had renamed it to `FetchAllSilencesDetailed()` returning `([]SilenceWithSource, map[string]error)`, so an unreachable Alertmanager degrades only its own source instead of blanking the whole silence inventory (matching `FetchAllAlertsDetailed`).
- This branch had rewritten the same function to poll every Alertmanager concurrently via the generic `fanOut()` helper.

**Resolution:** kept both. The function keeps `main`'s `FetchAllSilencesDetailed` name/signature and per-source `failedSources` map, and its body now iterates `fanOut(mc.snapshotClients(), (*Client).FetchSilences)`, mapping `result.err` into `failedSources[result.name]`. The explicit `mc.mutex.RLock()` from `main` was dropped because `snapshotClients()` already takes the read lock and releases it before the fan-out — holding it across the concurrent HTTP calls would have serialized nothing and blocked writers for the full poll duration.

`main`'s only caller (`internal/webui/handlers/silence_handlers.go:188`) already uses the `Detailed` variant, so no call sites changed.

**Verification:** `go build ./...` and `go vet ./...` pass; `go test ./internal/alertmanager/` passes (8 tests).

internal/alertmanager/client_test.go
```diff
@@ -0,0 +1,293 @@
...
+
+	if len(status) != len(urls) {
+		t.Fatalf("got %d statuses, want %d", len(status), len(urls))
+	}
+	if status["broken"] {
+		t.Fatal("broken source reported healthy")
+	}
+	if elapsed > time.Duration(float64(delay)*1.8) {
+		t.Fatalf("health fan-out took %s, want near %s", elapsed, delay)
+	}
```
SoulKyu: **The wall-clock budget that got removed from `TestFetchAllAlertsDetailedIsConcurrent` is still the only concurrency assertion here — and in `TestFetchAllAlertsDetailedPartialFailure` (line 137).**

Problem: `elapsed > delay*1.8` allows 540 ms for a 300 ms stub delay across five `httptest` servers, and it is the sole check that `GetHealthStatus` fans out at all. It is weak in the same way the old fetch test was: it passes when only two of the five probes overlap, so a partial regression in `testConnections` (say, a semaphore of 2 added later) stays green. And it is flake-prone: the package runs under `-race`, and on a throttled or oversubscribed CI runner five goroutines plus five servers can exceed 540 ms with the code perfectly correct — reported as a fan-out regression rather than a slow runner.

Why it matters: this is now the *only* remaining guard on the health-probe half of the PR, since `GetHealthStatus` and `GetHealthyClients` were converted to `testConnections` in the same change (client.go:896 and client.go:905) and no other test observes their concurrency. A timing assertion that false-alarms is the one most likely to get widened or `-short`-skipped, and when that happens the property stops being guarded at all. `barrierStub` was introduced precisely to avoid this, so leaving one test on the old pattern is also an inconsistency a future reader has to reconcile.

Evidence: lines 280-282 assert only an upper bound on `time.Since(start)`; nothing in this test observes how many probes were in flight. Line 137 has the same shape (`elapsed > 900*time.Millisecond` against a 500 ms client timeout plus 50 ms stubs, so roughly a 350 ms margin).

Suggested change: reuse the barrier you already built. Arm it over the four delayed sources only — `broken` answers immediately and must not be a barrier participant — and re-arm before the `GetHealthyClients` round, which probes a second time. A `WaitGroup` is legal to reuse once the counter has reached zero and every `Wait` has returned, which is guaranteed here because `GetHealthStatus` only returns after every response has completed.

```go
var barrier sync.WaitGroup
barrier.Add(4)
urls := map[string]string{}
for i := 0; i < 4; i++ {
    name := fmt.Sprintf("am-%d", i)
    urls[name] = barrierStub(t, name, &barrier).URL
}
// ... broken stub as today, not a barrier participant ...

status := mc.GetHealthStatus() // serial probing now deadlocks on the barrier
                               // and fails on the 10s client timeout
...
barrier.Add(4) // re-arm for the second probe round
healthy := mc.GetHealthyClients()
```

Then drop lines 280-282 entirely rather than keeping them as a smoke signal. If you would rather not thread a re-armable barrier through both rounds, the cheaper alternative is a peak-in-flight counter in the stub (`inFlight++` / track max / `inFlight--` under a mutex) with `if peak < 4 { t.Fatalf(...) }` — it survives a slow runner and still fails a serial regression, though less deterministically than the barrier. For line 137 the barrier does not apply, since the point there is that a dead source's timeout does not stack onto the healthy latencies; that one is defensible as-is, but consider widening it to a value derived from the timeout (for example `2*deadTimeout`) so the intent is "did not serialise" rather than a hand-tuned number.


SoulKyu (commented):
@SoulKyu Re-reviewed at 4e22deb. The four findings from the previous round all landed correctly: the drain is now capped at `healthCheckDrainLimit` with a 256 KiB stub proving the cap, `barrierStub` replaced the wall-clock budget in `TestFetchAllAlertsDetailedIsConcurrent`, `serving` orders the fetch ahead of the writer in `TestFanOutDoesNotHoldMutexAcrossHTTP`, and the renamed `TestTestConnectionUsesHealthFilterAndReusesConnection` now pins reuse via `RemoteAddr`.

I also re-verified the parts of the fan-out that a review can get wrong: `go.mod` declares `go 1.23.0`, so the `go func()` in `fanOut` captures per-iteration `i`/`nc` and the `results[i]` writes are to distinct slots published by `wg.Wait()`. `UpdateFromConfig` replaces `mc.clients` with freshly built `*Client` values rather than mutating the ones a snapshot may still hold, and each client owns its own `ProxyAuthManager`, so dropping `mc.mutex` around the I/O does not introduce shared-state writes. `GetHealthyClients` indexing `clients[i]` against the fan-out result is sound because `fanOut` preserves snapshot order. `go build -tags "nogui,webui" ./...`, `go vet ./...` and `go test -race -count=5 ./internal/alertmanager/...` are clean here, including at `GOMAXPROCS=1`.

One non-blocking follow-up, inline: the timing-based assertions the last round removed from the fetch test are still present in two other tests. Nothing here blocks merge.




internal/alertmanager/client_test.go
SoulKyu: @SoulKyu I'm not making a code change for this thread.

Already implemented at the current head 8e74312: barrierStub is armed over the four probed sources (client_test.go:264-271), the broken stub answers immediately and stays out of the barrier (272-277), the barrier is re-armed before the GetHealthyClients round (297), the wall-clock budget was dropped, and the partial-failure bound is now derived from the dead-source timeout as 2*deadTimeout (141). Verified with go test -race -count=3.




SoulKyu:
## 🧪 QA Report

Booted the docker-compose env (`make test` fails on this host — no `templ` binary — so I ran `make docker-build-all` + `docker compose up -d`; the Dockerfiles do the templ step internally). To actually exercise the multi-Alertmanager fan-out I added an untracked compose override giving the webui **1 healthy fake Alertmanager + 2 `socat` tarpits** that accept TCP and never answer, so each dead source costs its full 10 s client timeout. **Serial ⇒ ~20 s, concurrent ⇒ ~10 s** — the arithmetic that separates the two.

|---|---|---|
| `FetchAllAlertsDetailed` wall-clock ≈ max(latency), not sum | ✅ | `TestFetchAllAlertsDetailedIsConcurrent` (2 and 5 sources) passes; its `barrierStub` only answers once every peer is also in flight, so a serial regression deadlocks rather than merely being slow. Runtime: `Alert cache refresh` log lines land at 07:49:48 / 07:49:58 / 07:50:08 — exactly **10 s apart** with two sources each burning a 10 s timeout (the issue predicted ~21 s serially). |
| 1 hanging + 2 healthy → healthy alerts returned, dead in `failedSources`, bounded by the single timeout | ✅ | `TestFetchAllAlertsDetailedPartialFailure` passes. Runtime: `GET /health/alertmanager` with 1 healthy + 2 tarpits returned **10.003 s / 10.003 s / 10.003 s** over 3 runs (HTTP 206), body `{"healthy":false,"instances":{"Fake Alertmanager":true,"Tarpit A":false,"Tarpit B":false}}` — one timeout total, not two. |
| Partial-failure semantics unchanged; `alert_cache.go:236-238` behaviour + existing tests preserved | ✅ | `TestAlertCache_RefreshWithPartialFetchFailure` (4 subtests, incl. "Alerts from a failed source survive the cycle") passes under `-race`. Runtime logs: `failed to fetch alerts from Tarpit A, keeping its cached alerts untouched` for both tarpits, while the healthy source kept delivering — `fetched 25 → 26 → 27 alerts`, `0 newly resolved`. No mass-resolve. |
| `mc.mutex` not held across any HTTP call | ✅ | `TestFanOutDoesNotHoldMutexAcrossHTTP` passes — it waits for the fan-out to be *inside* the handler before racing `UpdateFromConfig` for the write lock. Code: `snapshotClients()` copies under `RLock` and releases before `fanOut()`. |
| `FetchAllSilences`, `TestAllConnections`, `GetHealthStatus`, `GetHealthyClients` concurrent, return shapes unchanged | ✅ | Silences at runtime: `GET /api/v1/dashboard/silences` = **10.005 s / 10.004 s** with 2 dead sources (serial ≥ 20 s), body still carries `failedSources` + `silences` with per-silence `source`. `TestGetHealthStatusConcurrent` covers both health fan-outs (barrier stub + correct name→client mapping). `TestAllConnections` shares the same proven `testConnections()` helper — note it has **no caller anywhere in the tree**, so runtime coverage is via the identical helper path. |
| `TestConnection` drains/closes body on success | ✅ | `defer resp.Body.Close()` + bounded `io.CopyN`. `TestTestConnectionUsesHealthFilterAndReusesConnection` (same client port on probe 2 ⇒ connection returned to keep-alive pool) and `TestTestConnectionBoundsDrainOfIgnoredFilter` (new connection when the payload exceeds the 4 KiB ceiling) both pass. Runtime on the fake AM: full payload **556 B → 13 B** (`[]`) with the health filter. |
| `go build -tags "nogui,webui" ./...`, `go vet ./...`, `go test -race ./internal/alertmanager/... ./internal/webui/services/...` clean | ✅ | All three clean on go1.23.10 — **54 tests pass** across the 2 packages under `-race`. The `%w`→`%v` fix in `internal/webui/router.go` is what makes `go vet` pass. |

**Exploratory:** Logged into `/dashboard` through the real form in headless Chromium with two Alertmanagers dead — alerts from the healthy source render and keep arriving live (**18 → 19 rows over 14 s**), `GET /api/v1/dashboard/stream` holds open at 200, and `/api/v1/dashboard/data` answers in **7.9 ms** (cache untouched by the dead sources); the only console 401s are pre-login hits on `/api/impersonate/status` and `/api/v1/profile/timezone`, unrelated to this diff. Biggest regression risk here is the *changed health-probe URL*, so I probed a real upstream `prom/alertmanager:latest` with the new filter: **HTTP 200, `[]`** — the `alertname="…"` filter syntax is accepted, so the health check does not go false-negative against a genuine Alertmanager. Also confirmed the goroutine closure over `i`/`nc` is safe: `go.mod` declares `go 1.23.0`, so loop vars are per-iteration.

**Verdict: PASS**

SoulKyu:

internal/alertmanager/client_test.go
```diff
@@ -0,0 +1,307 @@
...
+	if len(failed) != 1 {
+		t.Fatalf("unexpected failures: %v", failed)
+	}
+	// Bounded by the single timeout, not timeout + healthy latencies. The bound
+	// is derived from the timeout so the intent stays "did not serialise" rather
+	// than a hand-tuned number that a slow runner can trip.
+	if elapsed > 2*deadTimeout {
+		t.Fatalf("fan-out took %s, want it bounded by the dead source timeout %s", elapsed, deadTimeout)
+	}
```
SoulKyu: **The `2*deadTimeout` bound cannot detect serialisation — a fully sequential `fanOut` passes this test in 0.61 s.**

Problem: the comment above the assertion says the bound expresses "did not serialise", but the arithmetic does not support that. The only source with a shortened timeout is `dead` at `deadTimeout` = 500 ms; the two healthy stubs sleep 50 ms each (lines 106-107). A sequential fan-out therefore costs 500 + 50 + 50 ≈ 600 ms, comfortably inside the 1000 ms bound, so the assertion passes either way.

Evidence: I replaced the `sync.WaitGroup` loop in `fanOut` with a plain `for i, nc := range clients { ... }` sequential loop in this worktree and ran `go test -race -run TestFetchAllAlertsDetailedPartialFailure -count=1 ./internal/alertmanager/` — `--- PASS ... (0.61s)`. The same mutation fails `TestFetchAllAlertsDetailedIsConcurrent` (both subtests) and `TestGetHealthStatusConcurrent`, so the barrier stubs do catch it and this test genuinely does not.

Why it matters: with the barrier now covering the fetch and health fan-outs, this test is the only place that pins the *other* half of issue #80 — that a dead source's timeout does not stack onto the healthy latencies. As written it contributes no signal, only residual flake surface: the single way it can fail is a runner slow enough to push a correct ~500 ms fan-out past 1 s, which is precisely the failure mode the barrier rewrite was introduced to eliminate. An assertion whose comment claims a property it cannot observe is worse than no assertion, because the next reader takes the comment at face value.

Suggested change: give the test a second dead source so serialising actually stacks two timeouts. Serial then costs 500 + 500 + 50 + 50 ≈ 1100 ms and trips the existing 1 s bound, while concurrent stays ≈ 500 ms.

```go
deadA := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) { <-hang }))
deadB := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) { <-hang }))
defer func() { close(hang); deadA.Close(); deadB.Close() }()

urls := map[string]string{
    "healthy-a": alertmanagerStub(t, "healthy-a", 50*time.Millisecond).URL,
    "healthy-b": alertmanagerStub(t, "healthy-b", 50*time.Millisecond).URL,
    "dead-a":    deadA.URL,
    "dead-b":    deadB.URL,
}
...
for _, nc := range mc.snapshotClients() {
    if strings.HasPrefix(nc.name, "dead") {
        nc.client.HTTPClient.Timeout = deadTimeout
    }
}
```

Both tarpits can share the one `hang` channel, and `strings` is already imported (line 7). The assertions below then become `len(failed) != 2` with both `dead-a` and `dead-b` present, and the "dead source contributed alerts" loop keys off `strings.HasPrefix(a.Source, "dead")`. If you want more headroom against a slow runner, a third dead source pushes the serial cost to ≈ 1600 ms against the same 1 s bound.

If you would rather keep exactly one dead source, delete lines 138-143 instead. The three assertions above them — two healthy alerts, `dead` present in `failedSources`, nothing else failed — are the real content of this test, and dropping the bound removes the flake surface without losing any coverage.


SoulKyu (commented):
@SoulKyu Re-reviewed the fan-out at `b93d931`, focused on what the last two rounds changed.

The production side of `internal/alertmanager/client.go` holds up on a fresh pass. `snapshotClients()` copies under `RLock` and releases before any I/O; `UpdateFromConfig`, `UpdateClient` and `AddClient` all *replace* map entries with freshly built `*Client` values rather than mutating one a snapshot may still hold, and I grepped the tree for in-place writes to `BaseURL`/`Token`/`Username`/`Password`/`Headers`/`HTTPClient` outside the constructors to confirm it — there are none, so dropping `mc.mutex` around the HTTP work introduces no shared-state writes. `go.mod` declares `go 1.23.0`, so the `go func()` in `fanOut` captures per-iteration `i`/`nc`, and the `results[i]` writes land in distinct slots published by `wg.Wait()`. `GetHealthyClients` indexing `clients[i]` is sound because `fanOut` preserves snapshot order, and `result.err == nil` is exactly the old `IsHealthy()` semantics.

The bounded drain landed well: `io.CopyN(io.Discard, resp.Body, healthCheckDrainLimit)` plus `TestTestConnectionBoundsDrainOfIgnoredFilter` pins the ceiling from the outside, via connection reuse, which is the right observable for it. `barrierStub` is also doing its job — I reverted `fanOut` to a plain sequential loop locally and both `TestFetchAllAlertsDetailedIsConcurrent` subtests and `TestGetHealthStatusConcurrent` failed deterministically on the client timeout.

`go build -tags "nogui,webui" ./...`, `go vet ./...` and `go test -race -count=3 ./internal/alertmanager/...` are all clean here.

One non-blocking follow-up, inline: the same sequential mutation that kills the barrier tests leaves `TestFetchAllAlertsDetailedPartialFailure` green, so its remaining wall-clock bound is carrying a claim it cannot check. Nothing here blocks merge — this is a solid piece of work, and the barrier rewrite in particular is a much better guard than the timing budgets it replaced.



internal/alertmanager/client_test.go
SoulKyu: @SoulKyu I'm not making a code change for this thread.

No change this round: the suggested fix is already on the branch at head 283b769 — client_test.go:98-124 stands up deadA and deadB on a shared hang channel, applies deadTimeout via strings.HasPrefix(nc.name, "dead"), and asserts len(failed) != 2 with both dead-a and dead-b present. Verified by mutation: replacing fanOut's WaitGroup loop with a sequential loop fails the test at 1.105s against the 1s bound.




SoulKyu:
## 🧪 QA Report

Re-ran QA at `283b769` (head moved since the last report). Booted the compose env with `make docker-build-all` + `docker compose up -d` (`make test` still needs a local `templ`, which this host lacks — the Dockerfiles do the templ step internally). To make the fan-out arithmetic observable I ran the webui with **1 healthy fake Alertmanager + 2 `socat` tarpits** (accept TCP, never answer) via an untracked compose override, so each dead source costs a full 10 s client timeout: **serial ⇒ ~20 s, concurrent ⇒ ~10 s**.

|---|---|---|
| `FetchAllAlertsDetailed` wall-clock ≈ max(latency), not sum | ✅ | `TestFetchAllAlertsDetailedIsConcurrent` (2 and 5 sources) passes — its `barrierStub` only answers once every peer is in flight. Runtime: `Alert cache refresh complete` at 09:51:48 / :58 / 09:52:08 / :18 / :28 — a steady **10 s** cadence with two 10 s-timeout sources; the issue predicted ~21 s serially. Upstream confirms it: the fake AM logs `GET /api/v2/alerts` at 09:52:28→09:53:38, every 10 s. |
| 1 hanging + 2 healthy → healthy alerts returned, dead in `failedSources`, bounded by the single timeout | ✅ | `TestFetchAllAlertsDetailedPartialFailure` passes. Runtime `GET /health/alertmanager` with 1 healthy + 2 tarpits: **10.002 / 10.001 / 10.002 s** over 3 runs (HTTP 206), body `{"healthy":false,"instances":{"Fake Alertmanager":true,"Tarpit A":false,"Tarpit B":false}}` — one timeout total, not two. |
| Partial-failure semantics unchanged (`alert_cache.go:236-238`, existing coverage still green) | ✅ | `TestAlertCache_RefreshWithPartialFetchFailure` (4 subtests incl. "Alerts from a failed source survive the cycle") passes under `-race`. Runtime logs: `failed to fetch alerts from Tarpit A/B, keeping its cached alerts untouched` every cycle while the healthy source keeps delivering — `22 → 23 active alerts, 0 newly resolved`. No mass-resolve. |
| `mc.mutex` not held across any HTTP call | ✅ | `TestFanOutDoesNotHoldMutexAcrossHTTP` passes. Runtime cross-check: fired `/api/v1/dashboard/silences` and `/health/alertmanager` **simultaneously** — each 10.004 s, **combined wall 10.03 s**. A lock held across the I/O would have serialised them to ~20 s. |
| `FetchAllSilences`, `TestAllConnections`, `GetHealthStatus`, `GetHealthyClients` concurrent, shapes unchanged | ✅ | Silences at runtime: `GET /api/v1/dashboard/silences` = **10.007 / 10.005 s** with 2 dead sources (serial ≥ 20 s); body still carries `failedSources` + `silences[].source`. `TestGetHealthStatusConcurrent` covers both health fan-outs. Note `TestAllConnections` still has **no caller in the tree** — it shares the proven `testConnections()` helper, so runtime coverage is via that identical path. |
| `TestConnection` drains/closes the body on the success path | ✅ | `defer resp.Body.Close()` + bounded `io.CopyN`. Runtime: the AM access log shows the probe as `GET /api/v2/alerts?filter=alertname%3D"__notificator_health_check__"`, and the same request drops the payload from **20179 B → 3 B** (`[]`). `TestTestConnectionUsesHealthFilterAndReusesConnection` (same client port on probe 2 ⇒ keep-alive) and `TestTestConnectionBoundsDrainOfIgnoredFilter` both pass. |
| `go build -tags "nogui,webui" ./...`, `go vet ./...`, `go test -race ./internal/alertmanager/... ./internal/webui/services/...` clean | ✅ | All three clean on go1.23.10 — **54 tests** across the 2 packages under `-race`. The `%w`→`%v` fix in `internal/webui/router.go:52` is what makes `go vet` pass. |

**Mutation check (the two new test commits):** I reverted `fanOut` to a plain sequential loop in my working tree and re-ran the package — `TestFetchAllAlertsDetailedIsConcurrent` (both subtests), `TestGetHealthStatusConcurrent` **and** `TestFetchAllAlertsDetailedPartialFailure` (`fan-out took 1.102s, want it bounded by the dead source timeout 500ms`) all fail. So `283b769` does close the gap the last review flagged: the partial-failure wall-clock bound now actually bites. Tree restored, nothing pushed.

**Exploratory:** Logged into `/dashboard` through the real form in headless Chromium with both tarpits dead — 14 alerts from the healthy source render, `GET /api/v1/dashboard/stream` holds open at **200**, and `/api/v1/dashboard/data` answers in **8.2 ms** (cache untouched by the dead sources); `/silences` also renders. Only console errors are three pre-login 401s (`/api/v1/profile/timezone`, `/api/impersonate/status`) — zero after login, and unrelated to this diff. The real regression risk here is the *changed health-probe URL*, so I probed an actual `prom/alertmanager:latest` with the new filter: **HTTP 200, `[]`** — the syntax is accepted upstream, so the health check does not go false-negative against a genuine Alertmanager.

**Verdict: PASS**

PR #86, opened 2026-07-25
feat(silences): silence inventory — audit, extend and expire silences across every Alertmanager

### LINKED ISSUE #76: feat(silences): silence inventory — audit, extend and expire silences across every Alertmanager
## User & need

The on-call engineer who silenced something during last night's incident.

Notificator can **create** silences (dashboard → Silence, single or bulk) and **remove** the ones
attached to a currently-firing alert (`processUnsilenceAction` deletes the IDs in
`status.silencedBy`). Everything else about a silence's life is invisible:

- **No expiry visibility.** You silence `KafkaLagHigh` for 2h at 01:00 during an incident. At 03:00
  the silence expires and the alert storms back with no warning. Nothing in the UI ever told you
  when it would end.
- **No inventory.** A silence created for an alert that is no longer firing does not appear
  anywhere in Notificator — it only exists in Alertmanager's own UI. Those are exactly the
  dangerous ones: the stale `severity=critical` silence from three weeks ago that is quietly
  suppressing a real page.
- **N copies, N UIs.** `processSilenceAction` (`dashboard_handlers.go:1937`) creates the silence on
  **every** configured Alertmanager. One logical silence becomes N silences, and the only way to
  audit them is to open N Alertmanager UIs (with the right `X-Scope-OrgID` for Mimir/Cortex
  tenants). Notificator is the console the team lives in; it is the only place that can show them
  together.
- **The alert modal lies by omission.** `GetAlertDetails` sets `details.Silences = []` with a
  `// Note: Silences would need to be implemented` comment (`dashboard_handlers.go:1285`). Looking
  at a silenced alert, you cannot see who silenced it, why, or until when — only opaque
  `silencedBy` IDs.

Cost today: re-silencing the same alert over and over, surprise re-fires mid-incident, and no way
to answer "what are we currently hiding?" during a review.

## Proposed feature

**A silence inventory page (`/silences`) plus real silence detail in the alert modal.**

From the user's side of the screen:

- A new nav entry **Silences**. One table, all configured Alertmanagers, one row per silence:
  source Alertmanager, matchers, creator, comment, state (`active` / `pending` / `expired`),
  **ends in 47m** relative countdown, and **# alerts currently matched** (computed against the live
  alert cache).
- Default view: active + pending, sorted by soonest expiry. An **Expiring soon** badge on anything
  under 1h. Free-text filter over matchers/creator/comment, plus a source filter.
- **Suppressing nothing** marker on active silences that currently match zero cached alerts — the
  cleanup worklist.
- Per-row actions: **Extend +1h / +4h / +24h** and **Expire now**. No detour through the
  Alertmanager UI, no re-deriving matchers by hand.
- In the alert detail modal, the existing (always-empty) Silences area becomes real: for a silenced
  alert, each silence with creator, comment, end time + countdown, and the same Extend/Expire
  actions.

## Approach

Everything needed already exists and is currently **dead code** — no new proto, no new RPC, no DB
work. Silences stay on the "straight to Alertmanager" side of the two mental models documented in
`openwiki/dashboard.md#alert-actions`.

- `internal/alertmanager/client.go`: `MultiClient.FetchAllSilences()` (line 727) and
  `SilenceWithSource` (line 666) exist and have **zero callers**. Add a
  `FetchAllSilencesDetailed()` returning `map[string]error` per source, mirroring the
  `FetchAllAlertsDetailed()` convention (line 674) so one unreachable Alertmanager degrades that
  source instead of blanking the page.
- Extend = `Client.CreateSilence()` with the **same `ID`** and a new `EndsAt` (Alertmanager
  `POST /api/v2/silences` upserts on `id`); `CreateSilence` already marshals the whole
  `models.Silence`, so this is a field set, not new HTTP code. Expire = existing
  `DeleteSilenceFromAlertmanager()` (line 793).
- New `internal/webui/handlers/silence_handlers.go`:
  `GET /api/v1/dashboard/silences`, `POST /api/v1/dashboard/silences/:id/extend`,
  `DELETE /api/v1/dashboard/silences/:id` — registered in the existing `dashboard` group in
  `internal/webui/router.go`, plus a protected page route `GET /silences` next to `/statistics`
  (line 358). Both mutating endpoints take the source Alertmanager name.
- Matched-alert count: a small `silenceMatchesAlert(silence, alert)` helper (all matchers must
  hold; honour `isRegex` / `isEqual`) run over the in-memory `alertCache` — the same source
  `GetDashboardData` reads, so no extra Alertmanager traffic.
- Alert modal: fill `details.Silences` in `GetAlertDetails` (`dashboard_handlers.go:1285`) from the
  fetched silences intersected with `alert.Status.SilencedBy`, into the existing
  `webuimodels.Silence` type (`internal/webui/models/dashboard.go:245`) — the wire type is already
  there and unused.
- UI: new `templates/pages/Silences.templ` + a small Alpine component, following the
  `StatisticsDashboard` page/handler shape; modal section in
  `templates/components/alert_modal_shared.templ`. Regenerate with `make webui-templates`; never
  touch `*_templ.go`.
- One Go test for `silenceMatchesAlert` (equal / not-equal / regex / multi-matcher).

## Acceptance criteria

- [ ] `/silences` is reachable from the nav for any logged-in user and lists silences from **every**
      configured Alertmanager with a source column.
- [ ] Each row shows matchers, creator, comment, state, absolute end time **and** a relative
      countdown; active silences ending in <1h are visually flagged.
- [ ] Default sort is soonest-expiry-first; expired silences are hidden until the state filter
      includes them.
- [ ] Each active silence shows how many alerts it currently matches; zero-match silences are
      marked as suppressing nothing.
- [ ] Extend (+1h/+4h/+24h) updates the silence in place on the correct Alertmanager and the new end
      time is visible after refresh; the silence ID is unchanged.
- [ ] Expire now removes the silence from the correct Alertmanager only, and the row leaves the
      active view.
- [ ] If one Alertmanager is unreachable, the page still renders the reachable sources and names the
      failing one (no blanket error, no silent empty table).
- [ ] Opening a silenced alert's modal shows its silences with creator, comment and end time
      (`details.Silences` is no longer empty).
- [ ] No changes to `proto/` and no DB migration.
- [ ] `make webui-templates` regenerated; no hand-edited `*_templ.go`.

## Out of scope

- **Creating** silences from this page — creation stays in the dashboard flow (see #27 for matcher
  editing and #25 for duration limits; this issue deliberately does not touch either).
- Proactive notification when a silence is about to expire (needs a scheduler; natural follow-up
  once the inventory exists).
- Persisting silence history in the backend DB or attributing silences to Notificator users beyond
  Alertmanager's `createdBy`.
- Deduplicating the N per-Alertmanager copies into one logical silence — they are listed with their
  source, not merged.
- Editing matchers of an existing silence.

SoulKyu:
## Why

Notificator could create silences but never showed what happened to them afterwards. `processSilenceAction` fans a single logical silence out to **every** configured Alertmanager, and the only way to audit the result was to open N Alertmanager UIs. Nothing surfaced expiry times, stale silences suppressing nothing, or who silenced what — the alert modal literally hardcoded `details.Silences = []`.

## What changed

**Backend (`internal/alertmanager`, `internal/models`)**
- `MultiClient.FetchAllSilencesDetailed()` returns per-source errors, mirroring the existing `FetchAllAlertsDetailed()` convention. `FetchAllSilences()` now delegates to it. One unreachable Alertmanager degrades that source instead of blanking the page.
- `SilenceMatcher.UnmarshalJSON` defaults `isEqual` to `true`. Alertmanager before 0.22 omits the field, and decoding it as `false` would silently negate every matcher.

**API (`internal/webui/handlers/silence_handlers.go`, new)**
- `GET /api/v1/dashboard/silences` — all silences from all sources, sorted soonest-expiry-first, with the list of configured sources and a `failedSources` map.
- `POST /api/v1/dashboard/silences/:id/extend` — body `{source, duration}`. Re-posts the silence with the **same ID** and a later `EndsAt`; Alertmanager upserts on `id`, so the identity is preserved. Extends from the current end time, or from now if the silence already lapsed. Duration is validated and capped at 30 days.
- `DELETE /api/v1/dashboard/silences/:id?source=…` — expires the silence on the named Alertmanager only.
- `silenceMatchesAlert()` counts matched alerts against the in-memory alert cache (restricted to the silence's own source), so the "suppressing nothing" marker costs no extra Alertmanager traffic. Regexes are fully anchored, matching Alertmanager's own semantics; a missing label is compared as the empty string.

**Alert modal (`GetAlertDetails`)**
- `details.Silences` is now populated by resolving `status.silencedBy` against the alert's own Alertmanager — at most one request per silence ID, and an unresolvable ID is skipped rather than failing the modal.

**UI**
- New `/silences` page: source column, matchers, creator, comment, state badge, absolute end time + relative countdown, matched-alert count, per-row `+1h / +4h / +24h` and `Expire now`. Default view is active + pending; free-text filter over matchers/creator/comment plus a source filter; an **Expiring soon** badge under 1h and a **Suppressing nothing** marker on zero-match active silences. Unreachable Alertmanagers are named in a banner above the table.
- `PageNavigator` gains a third tab, and the dashboard user dropdown a matching link.

## Tests

`silence_handlers_test.go` covers `silenceMatchesAlert` (equal / not-equal / regex / negated regex / anchoring / multi-matcher / missing label / invalid regex / no matchers) and the `isEqual` unmarshal default.

`go build ./...`, `go test ./...` and `templ generate` (pinned v0.3.906) all pass. The two pre-existing lint findings in `internal/webui/router.go` (`gofmt`, `log.Fatalf %w`) are untouched.

## Scope

No `proto/` changes, no DB migration. Silence creation, expiry notifications, matcher editing and cross-Alertmanager deduplication remain out of scope.

Closes #76


SoulKyu (commented):
@SoulKyu — reviewed the whole silence inventory: the three new `/api/v1/dashboard/silences*` handlers, `FetchAllSilencesDetailed` per-source degradation, the `isEqual` unmarshal default, the alert-modal silence resolution and the new `/silences` Alpine page. `go build ./...` and `go test ./internal/webui/handlers/...` both pass locally on `a3669c4`, and the matcher table test is a genuinely good fixture matrix — the anchoring and missing-label cases are exactly the ones that bite.

Nothing here blocks merge. Five inline notes below; the two worth acting on before this meets a busy Alertmanager are the per-alert regex compilation in `silenceMatcherMatches`, and what Alertmanager actually does when you re-POST a silence that has already lapsed.




internal/webui/handlers/silence_handlers.go
```diff
@@ -0,0 +1,255 @@
...
+// maxSilenceExtension caps how far a single extend request can push the end time.
+const maxSilenceExtension = 30 * 24 * time.Hour
+
+// silenceMatcherMatches reports whether a single matcher holds for a label value.
+// An absent label is matched as the empty string, which is how Alertmanager evaluates it.
+func silenceMatcherMatches(matcher models.SilenceMatcher, value string) bool {
+	matched := false
+	if matcher.IsRegex {
+		// Alertmanager anchors silence regexes on both ends.
+		re, err := regexp.Compile("^(?:" + matcher.Value + ")$")
```
SoulKyu: **Compile each matcher once per silence, not once per alert — and skip expired silences entirely.**

`regexp.Compile` runs on the innermost loop of the whole endpoint. `countMatchedAlerts` (line 59) walks every cached alert of the source, `silenceMatchesAlert` (line 45) walks every matcher of the silence, and this line recompiles the pattern from source text on each of those iterations. `GetSilences` then repeats that for **every** silence `FetchAllSilencesDetailed()` returns — including expired ones, which Alertmanager keeps around for `--data.retention` (120h by default) and which the page never even shows a matched count for (`Silences.templ:140` hides the count behind `silence.status?.state !== 'expired'`).

Why it matters: the cost is `O(silences × alerts × matchers)` regex compilations inside a synchronous handler that runs on every page load and again after every extend/expire. A site with 300 retained silences, 1500 cached alerts and 2 regex matchers each does roughly 900k `regexp.Compile` calls to render one table — hundreds of milliseconds to seconds of pure CPU, repeated per user per refresh. Nothing caches the result.

Concrete fix — hoist the compilation out of the alert loop, and don't count what isn't displayed:

```go
type compiledMatcher struct {
	name    string
	value   string
	isEqual bool
	re      *regexp.Regexp // nil for plain equality matchers
}

func compileMatchers(silence models.Silence) ([]compiledMatcher, bool) {
	compiled := make([]compiledMatcher, 0, len(silence.Matchers))
	for _, m := range silence.Matchers {
		cm := compiledMatcher{name: m.Name, value: m.Value, isEqual: m.IsEqual}
		if m.IsRegex {
			re, err := regexp.Compile("^(?:" + m.Value + ")$")
			if err != nil {
				return nil, false // invalid regex: silence matches nothing
			}
			cm.re = re
		}
		compiled = append(compiled, cm)
	}
	return compiled, true
}
```

`countMatchedAlerts` then compiles once up front and reuses the slice across all alerts. And in `GetSilences` (line 136):

```go
matched := 0
if item.Silence.Status.State != "expired" {
	matched = countMatchedAlerts(item.Silence, item.Source, alerts)
}
```

The exported behaviour and the existing table test stay identical; keep `silenceMatchesAlert` as the tested entry point and have it delegate to the compiled form.


internal/webui/handlers/silence_handlers.go
```diff
@@ -0,0 +1,255 @@
...
+		c.JSON(http.StatusBadGateway, webuimodels.ErrorResponse("Failed to fetch silence: "+err.Error()))
+		return
+	}
+
+	// Extend from the current end time, or from now if the silence already lapsed.
+	base := silence.EndsAt
+	if now := time.Now(); base.Before(now) {
+		base = now
+	}
+	silence.EndsAt = base.Add(duration)
```
SoulKyu: **Extending a silence that has already lapsed does not extend it — Alertmanager creates a different silence.**

The `base = now` branch is written as if re-POSTing an expired silence revives it under the same ID, and the comment on line 157 plus the PR description say so outright ("Alertmanager upserts on `id`, so the identity is preserved"). That only holds for `active` and `pending` silences. Alertmanager's `Silences.Set` calls `canUpdate(prev, sil, now)`, whose `types.SilenceStateExpired` case returns `false` unconditionally; control then falls through to `sil.Id = uid()` and `if sil.StartsAt.Before(now) { sil.StartsAt = now }`. The result is a brand-new silence with a new ID and a new start time, while the original stays expired.

Why it matters: this is reachable from the UI, not just from curl. `Silences.templ:147` hides the action buttons on `state === 'expired'`, but the page only refetches on demand and `this.now` ticks every 30s — a silence that lapses after page load still renders its `+1h` button. The operator clicks "extend", gets a 200, and ends up with a duplicate silence plus an expired original in the audit trail. Worse, line 208 returns `toWebuiSilence(*updated, ...)` built from the object fetched at line 189, so the response carries the new ID next to the *stale* `status.state` (`"expired"`) and the old `startsAt` — the API is describing a silence that does not exist in that shape anywhere.

Concrete fix — make the lapsed case explicit rather than silently recreating:

```go
if silence.Status.State == "expired" {
	c.JSON(http.StatusConflict, webuimodels.ErrorResponse("Silence has already expired and cannot be extended; recreate it instead"))
	return
}
silence.EndsAt = silence.EndsAt.Add(duration)
```

If recreating a lapsed silence is actually the behaviour you want, keep the `base = now` branch but (a) drop the "same ID" claim from the comment on line 157 and from the PR body, (b) re-fetch the created silence (or clear `Status`/`StartsAt`) before returning it so the response is not stale, and (c) surface it in the UI — e.g. a distinct message so the operator knows a new silence was created rather than the old one moved.


internal/webui/handlers/silence_handlers.go
```diff
@@ -0,0 +1,255 @@
...
+	failed := make(map[string]string, len(failedSources))
+	for name, err := range failedSources {
+		failed[name] = err.Error()
+	}
+
+	c.JSON(http.StatusOK, webuimodels.SuccessResponse(gin.H{
+		"silences":      silences,
+		"sources":       alertmanagerClient.GetClientNames(),
+		"failedSources": failed,
+	}))
```
SoulKyu: **Two ordering issues in the `GetSilences` response.**

**1. The sort buries the silences the page is actually about.** Sorting ascending on `EndsAt` across the whole set means expired silences — whose end times are all in the past — occupy the front of the list. With the default `states: ['active', 'pending']` filter this is invisible, but as soon as the operator enables the `expired` toggle (`Silences.templ:85`) the table opens with the oldest expired entries and pushes everything actionable below the fold. Alertmanager's 120h default retention makes that the common case, not an edge case. The PR describes the ordering as "soonest-expiry-first", which is only the useful reading for live silences.

Suggested change — group by liveness first, and order expired most-recent-first:

```go
sort.SliceStable(silences, func(i, j int) bool {
	iExpired := silences[i].Status.State == "expired"
	jExpired := silences[j].Status.State == "expired"
	if iExpired != jExpired {
		return !iExpired
	}
	if iExpired {
		return silences[i].EndsAt.After(silences[j].EndsAt)
	}
	return silences[i].EndsAt.Before(silences[j].EndsAt)
})
```

**2. `sources` comes back in random order.** `alertmanagerClient.GetClientNames()` (line 151, `internal/alertmanager/client.go:816`) ranges over `mc.clients`, a map, so Go's randomised iteration hands back a different order on every request. The "All sources" `<select>` at `Silences.templ:79` is rendered straight from it, so the dropdown reshuffles on every refresh and after every extend/expire — a small thing that reads as a rendering bug to whoever hits it. Add `sort.Strings(sources)` before putting it in the payload (`sort` is already imported here):

```go
sources := alertmanagerClient.GetClientNames()
sort.Strings(sources)
```


internal/alertmanager/client.go
```diff
@@ -746,8 +749,16 @@ func (mc *MultiClient) FetchAllSilences() ([]SilenceWithSource, error) {
...
+}
+
+func (mc *MultiClient) FetchAllSilences() ([]SilenceWithSource, error) {
+	allSilences, failedSources := mc.FetchAllSilencesDetailed()
+
+	if len(failedSources) > 0 && len(allSilences) == 0 { // If all clients failed, return the first error
+		for name, err := range failedSources {
+			return nil, fmt.Errorf("failed to fetch silences from %s: %w", name, err)
+		}
 	}
```
SoulKyu: **`FetchAllSilences` is now dead code, and the loop that replaced the error selection reads like a bug.**

`rg 'FetchAllSilences\b'` over the tree finds only this definition — the sole consumer of the old API, the silence inventory, calls `FetchAllSilencesDetailed` directly (`internal/webui/handlers/silence_handlers.go:127`). Nothing else in the repo, including the desktop code paths, calls the wrapper.

Separately, `for name, err := range failedSources { return ... }` unconditionally returns on the first iteration, so it is not a loop — it is "pick an arbitrary map entry" written in a shape that makes the next reader stop and check whether a `break` went missing. The old code's `errors[0]` was at least honestly indexed; map iteration order is randomised, so which Alertmanager gets named in the error now varies run to run.

Suggested change: delete `FetchAllSilences` entirely — `FetchAllSilencesDetailed` is strictly more informative and the compiler will flag anything that still needs it. If you want to keep it as a stable external surface, at least make the intent explicit and deterministic:

```go
func (mc *MultiClient) FetchAllSilences() ([]SilenceWithSource, error) {
	allSilences, failedSources := mc.FetchAllSilencesDetailed()
	if len(allSilences) == 0 && len(failedSources) > 0 {
		names := make([]string, 0, len(failedSources))
		for name := range failedSources {
			names = append(names, name)
		}
		sort.Strings(names)
		return nil, fmt.Errorf("failed to fetch silences from %s: %w", names[0], failedSources[names[0]])
	}
	return allSilences, nil
}
```


internal/webui/handlers/silence_handlers_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			silence := models.Silence{Matchers: tt.matchers}
+			if got := silenceMatchesAlert(silence, labels); got != tt.want {
+				t.Errorf("silenceMatchesAlert() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
```
SoulKyu, line 101: **Add a case for the cross-source guard in `countMatchedAlerts` — it is the one untested rule the UI actually leans on.**

The table above is thorough on `silenceMatchesAlert` (equal / not-equal / regex anchoring / negated regex / missing label / invalid regex / no matchers — good coverage). What it never exercises is the `alert.Source != source` guard at `silence_handlers.go:62-64`, which is the rule that stops a silence living on one Alertmanager from claiming identically-labelled alerts scraped from another. That guard is load-bearing twice over: it decides the `matchedAlerts` number, and it decides whether the "Suppressing nothing" marker appears (`Silences.templ:143`). In a multi-Alertmanager setup — the entire premise of this feature — identical `alertname`/`severity` label sets across sources are the norm, so a regression there would silently mislabel stale silences as busy ones and vice versa, with no test failing.

Suggested addition to this file:

```go
func TestCountMatchedAlertsIsScopedToSource(t *testing.T) {
	silence := models.Silence{Matchers: []models.SilenceMatcher{
		{Name: "alertname", Value: "KafkaLagHigh", IsEqual: true},
	}}
	alerts := []*webuimodels.DashboardAlert{
		{Source: "prod-am", Labels: map[string]string{"alertname": "KafkaLagHigh"}},
		{Source: "staging-am", Labels: map[string]string{"alertname": "KafkaLagHigh"}},
		{Source: "prod-am", Labels: map[string]string{"alertname": "DiskFull"}},
	}

	if got := countMatchedAlerts(silence, "prod-am", alerts); got != 1 {
		t.Errorf("countMatchedAlerts(prod-am) = %d, want 1", got)
	}
	if got := countMatchedAlerts(silence, "unknown-am", alerts); got != 0 {
		t.Errorf("countMatchedAlerts(unknown-am) = %d, want 0", got)
	}
}
```

That needs `webuimodels "notificator/internal/webui/models"` added to the import block. It also pins the behaviour if the compiled-matcher refactor suggested on `silence_handlers.go:28` lands.


SoulKyu:
## 🧪 QA Report

Env: `make docker-build-all` + `docker compose up -d`, fake Alertmanager, real browser (Playwright) + `curl`. Acceptance criteria from #76.

|---|---|---|
| `/silences` reachable from nav for a logged-in user, lists silences from **every** configured Alertmanager with a source column | ✅ | Logged in as a fresh user, `PageNavigator` shows `Alerts / Silences / Statistics`; clicking the nav link from `/dashboard` lands on `/silences` (2 `a[href="/silences"]` links found: navigator + user dropdown). Table renders a `SOURCE` column; with 2 Alertmanagers configured the header reads `2 alertmanager(s)`. |
| Row shows matchers, creator, comment, state, absolute end time **and** relative countdown; active silences ending <1h visually flagged | ✅ | Row rendered as `["Fake Alertmanager","alertname=DiskSpaceLow","oncall-alice","QA: expiring soon…","Active Expiring soon","7/25/2026, 8:35:12 AM ends in 29m","1 alert(s)","+1h +4h +24h Expire now"]`. Matcher operators render correctly for all three kinds: `alertname=DiskSpaceLow`, `severity=~crit.*`, `team!=network`. |
| Default sort soonest-expiry-first; expired hidden until the state filter includes them | ✅ | API returns silences ordered `05:05Z → 06:35Z → 08:04Z → 12:05Z → 14:05Z → 18:05Z`. Default view = 5 rows (all `active`); clicking **Expired** → 6 rows, the extra one being `oncall-carol / Expired / ended 1h 1m ago`. Free-text filter (`oncall-bob`, `NoSuchAlertEver`) and the source `<select>` both narrow correctly. |
| Each active silence shows how many alerts it matches; zero-match silences marked as suppressing nothing | ✅ | `alertname=DiskSpaceLow` → `1 alert(s)`; `severity=~crit.*` → `14 alert(s)`; `alertname=NoSuchAlertEver` → `0 alert(s) / Suppressing nothing`. Regex anchoring verified: `severity=~crit` matched **0** while `severity=~crit.*` matched 14. Negated matcher verified: `team!=network AND alertname=NetworkLatencyHigh` → 0, correct. A matcher posted **without** `isEqual` (pre-0.22 Alertmanager shape) decoded as equality, not negation. |
| Extend (+1h/+4h/+24h) updates in place on the correct Alertmanager, new end time visible after refresh, silence ID unchanged | ✅ | Clicked **+4h** in the browser on `oncall-alice`: `endsAt 06:35:12.366Z → 10:35:12.366Z` (delta exactly 240 min), `id 16c359e7-…f0ef` unchanged, row re-rendered as `ends in 4h 28m` and the *Expiring soon* badge cleared. `+1h` and `+24h` also confirmed (`12:05Z → 13:05Z → next-day 13:05Z`, same ID). Guardrails hold: `9999h` → 400 `Duration must be between 0 and 30 days`; unknown source → 502 `alertmanager 'Nope' not found`. |
| Expire now removes it from the correct Alertmanager only, row leaves the active view | ✅ | Clicked **Expire now** on `oncall-erin`, accepted the `Expire this silence on Fake Alertmanager now?` confirm. Row gone from the table, silence absent from `GET /api/v1/dashboard/silences`, and the other 5 silences untouched. `DELETE` without `?source=` → 400 `Source alertmanager is required`. |
| One Alertmanager unreachable → page still renders the reachable sources and names the failing one | ✅ | Restarted webui with a second AM at `http://127.0.0.1:19999`. Banner: *"Alertmanager `Dead Alertmanager` is unreachable — its silences are not listed"* + `connection refused` detail; the 4 `Fake Alertmanager` rows still render, `failedSources` map populated, no blanket error, no empty table. |
| Silenced alert's modal shows its silences with creator, comment and end time | ✅ | `GET /api/v1/dashboard/alert/<fp>` now returns a populated `silences` array. In the UI, modal → **Details** tab renders: `oncall-dave / Fake Alertmanager / "QA: regex severity critical, no isEqual field" / severity=~crit.* / Ends 7/25/2026, 8:05:12 PM (in 11h 55m)`. |
| No `proto/` changes, no DB migration | ✅ | `git diff --name-only main...HEAD` has no `proto/` or migration path. |
| `make webui-templates` regenerated, no hand-edited `*_templ.go` | ✅ | Ran `templ generate` (pinned v0.3.906) on the checked-out branch → `git status` clean, zero diff. `go test ./internal/webui/handlers/... ./internal/models/...` → 14 passed. |

**Exploratory:** Adjacent flows survive — creating a silence through the existing dashboard bulk-action still works and the new silence appears in the inventory immediately with its 5 matchers and a correct matched count; `/dashboard` and `/statistics` render unchanged; the 3-tab navigator round-trips; **zero JS console/page errors on `/silences`** across every interaction. Two non-blocking notes: (1) *pre-existing on `main`, not caused by this PR* — `Client.CreateSilence` posts `"id": ""` and the fake Alertmanager's `.get('id', uuid4())` keeps the empty string, so dashboard-created silences land with a blank ID in this test env; the new page degrades cleanly rather than breaking (Extend surfaces `Silence ID is required` in the error line, Expire 404s), and a real Alertmanager assigns the ID server-side; (2) the `pending` state badge could not be exercised because the fake Alertmanager hardcodes `state: "active"` on POST — it is the same `silence.Status.State` passthrough as `active`/`expired`, both of which were verified.

**Verdict: PASS**

internal/webui/handlers/silence_handlers.go
```diff
@@ -0,0 +1,318 @@
...
+
+	silence.EndsAt = silence.EndsAt.Add(duration)
+
+	updated, err := alertmanagerClient.CreateSilenceOnAlertmanager(request.Source, *silence)
+	if err != nil {
+		c.JSON(http.StatusBadGateway, webuimodels.ErrorResponse("Failed to extend silence: "+err.Error()))
+		return
+	}
+
+	c.JSON(http.StatusOK, webuimodels.SuccessResponse(toWebuiSilence(*updated, request.Source, 0)))
```
SoulKyu, line 312: **The lapsed-silence guard still has a window: Alertmanager can create a *new* silence here and this handler reports success.**

The 409 at line 258 is decided from the state read by the GET at line 250, but the decision that matters is taken inside Alertmanager on its own clock: `Silences.Set` calls `canUpdate(prev, sil, now)`, whose `types.SilenceStateExpired` case returns `false` unconditionally, and control then falls through to `sil.Id = uid()` plus `if sil.StartsAt.Before(now) { sil.StartsAt = now }`. If the silence lapses between our GET and that `Set`, we get a brand-new silence under a new ID with a new start time, the original stays expired — and this code returns HTTP 200 with `success: true`.

Why it matters: the page is *designed* to push operators at silences that are about to lapse. `isExpiringSoon()` badges anything under an hour (`Silences.templ:132`, `246-252`) and the `+1h / +4h / +24h` buttons sit on that same row, so "extend a silence seconds before it ends" is the intended workflow, not an exotic path. When it goes wrong the operator gets a green result, the audit trail gains a duplicate silence with a different ID, and the expired original stays behind — exactly the outcome commit 7591935 set out to prevent. The response is misleading on top of that: `updated` is the struct fetched *before* the POST, so its `status.state` and `updatedAt` describe a shape that no longer exists anywhere.

Evidence that the fix is nearly free: `Client.CreateSilence` (`internal/alertmanager/client.go:540-543`) copies the request struct and overwrites only `ID` with the `silenceID` Alertmanager echoed back. So a mismatch is already sitting in `updated` — no extra request needed to detect it.

Suggested change:

```go
updated, err := alertmanagerClient.CreateSilenceOnAlertmanager(request.Source, *silence)
if err != nil {
	c.JSON(http.StatusBadGateway, webuimodels.ErrorResponse("Failed to extend silence: "+err.Error()))
	return
}

// Alertmanager only reuses the ID when it accepted the update in place. A different ID
// means the silence lapsed mid-flight and this is a new silence, not an extension.
if updated.ID != silenceID {
	c.JSON(http.StatusConflict, webuimodels.ErrorResponse(
		"Silence lapsed while being extended; Alertmanager created a new silence "+updated.ID+" instead and the original stayed expired"))
	return
}
```

Optionally, re-read the silence before responding (`FetchSilenceFromAlertmanager(request.Source, silenceID)`, falling back to the local copy on error) so the payload carries Alertmanager's own `status`/`updatedAt` instead of the pre-POST snapshot. Keep the state check at line 258 either way — it is what catches the 30s-stale UI; the ID comparison closes the part only Alertmanager can see.


internal/webui/handlers/silence_handlers.go
```diff
@@ -0,0 +1,318 @@
...
+	if duration <= 0 || duration > maxSilenceExtension {
+		c.JSON(http.StatusBadRequest, webuimodels.ErrorResponse("Duration must be between 0 and 30 days"))
+		return
+	}
+
+	silence, err := alertmanagerClient.FetchSilenceFromAlertmanager(request.Source, silenceID)
+	if err != nil {
+		c.JSON(http.StatusBadGateway, webuimodels.ErrorResponse("Failed to fetch silence: "+err.Error()))
+		return
+	}
```
SoulKyu, line 278: **Client-side mistakes are reported as 502 Bad Gateway, which misleads the operator and pollutes 5xx alerting.**

Two failures reach this branch without any Alertmanager ever misbehaving:

- an unknown `source`: `MultiClient.FetchSilenceFromAlertmanager` returns `alertmanager '%s' not found` (`internal/alertmanager/client.go:786`) purely from a map lookup, before a request is made;
- an unknown silence ID: `Client.FetchSilence` returns `silence with ID %s not found` on an upstream 404 (`internal/alertmanager/client.go:428-430`).

Both become `502` here, and `ExpireSilence` (line 293) collapses the same two cases the same way.

Why it matters: `source` is whatever the row carried, and rows outlive a config change that renames or drops an Alertmanager — after such a change every extend/expire on a stale row surfaces as "Failed to fetch silence: alertmanager 'old-am' not found" behind a gateway error, reading as "the Alertmanager is broken" when the real answer is "reload the page". Second, anything watching webui 5xx rates (this is an SRE tool) starts alarming on user input. 502 should stay reserved for a genuine upstream fault, which is what the banner and the on-call graphs both assume.

Suggested change — validate the source up front in both `ExtendSilence` and `ExpireSilence`. It is deterministic and needs no error-string matching (`GetClientNames` is already used by `GetSilences`, and `slices` is available on go 1.23):

```go
if !slices.Contains(alertmanagerClient.GetClientNames(), request.Source) {
	c.JSON(http.StatusBadRequest, webuimodels.ErrorResponse("Unknown alertmanager source: "+request.Source))
	return
}
```

For the missing-silence case, the cheapest honest route is a sentinel in the client — `var ErrSilenceNotFound = errors.New("silence not found")` wrapped at `client.go:429` — matched here with `errors.Is` to return `404`, leaving `502` for transport and non-404 upstream statuses.


internal/webui/handlers/silence_handlers_test.go
```diff
@@ -0,0 +1,142 @@
...
+		t.Error("IsEqual should default to true when absent")
+	}
+
+	if err := json.Unmarshal([]byte(`{"name":"severity","value":"critical","isEqual":false}`), &matcher); err != nil {
+		t.Fatalf("unmarshal failed: %v", err)
+	}
+	if matcher.IsEqual {
+		t.Error("IsEqual should stay false when explicitly false")
+	}
+}
```
SoulKyu, line 164: **Nothing pins the new ordering rule — the one behaviour in this PR most likely to be "simplified" back into a bug.**

The table above is genuinely good on matching, and `TestCountMatchedAlertsIsScopedToSource` closes the cross-source guard. The gap is the comparator added in 7591935 at `silence_handlers.go:187-197`: three branches (live before expired, live ascending on `EndsAt`, expired *descending*) with no test. It is exactly the shape a later reader collapses to `EndsAt.Before(...)` because that reads "sorted by expiry" — which is what the previous review round flagged, since it buries every actionable silence under 120h of retained expired ones as soon as the `expired` toggle is on (`Silences.templ:85`). Today it is untestable from here because the logic is inlined in a handler that needs a live `*alertmanager.MultiClient`.

Suggested change — lift the comparator out of `GetSilences` (body moved verbatim, `GetSilences` then calls `sortSilences(silences)`):

```go
// silence_handlers.go
func sortSilences(silences []webuimodels.Silence) {
	sort.SliceStable(silences, func(i, j int) bool {
		iExpired := silences[i].Status.State == silenceStateExpired
		jExpired := silences[j].Status.State == silenceStateExpired
		if iExpired != jExpired {
			return !iExpired
		}
		if iExpired {
			return silences[i].EndsAt.After(silences[j].EndsAt)
		}
		return silences[i].EndsAt.Before(silences[j].EndsAt)
	})
}
```

and add here (needs `time` in the import block):

```go
func TestSortSilencesPutsLiveFirstAndExpiredNewestFirst(t *testing.T) {
	base := time.Now()
	silences := []webuimodels.Silence{
		{ID: "expired-old", EndsAt: base.Add(-48 * time.Hour), Status: webuimodels.SilenceStatus{State: "expired"}},
		{ID: "active-late", EndsAt: base.Add(6 * time.Hour), Status: webuimodels.SilenceStatus{State: "active"}},
		{ID: "expired-recent", EndsAt: base.Add(-1 * time.Hour), Status: webuimodels.SilenceStatus{State: "expired"}},
		{ID: "pending-soon", EndsAt: base.Add(1 * time.Hour), Status: webuimodels.SilenceStatus{State: "pending"}},
	}

	sortSilences(silences)

	want := []string{"pending-soon", "active-late", "expired-recent", "expired-old"}
	for i, id := range want {
		if silences[i].ID != id {
			t.Fatalf("position %d = %s, want %s", i, silences[i].ID, id)
		}
	}
}
```

That covers the mixed live/expired case, the pending-vs-active tie the UI's default filter shows, and the reversed expired order in one assertion.


SoulKyu (commented):
@SoulKyu — re-reviewed at 7591935 (silence inventory across every Alertmanager).

Everything from the previous round is properly closed: matchers are compiled once per silence and expired ones are skipped entirely, the lapsed-extend case is an explicit 409 instead of a silent duplicate, the ordering now groups live silences first and expired newest-first, `sources` is sorted, `FetchAllSilences` was deleted rather than left as a delegating shim, and `countMatchedAlerts` has a cross-source test. I also checked the generated templ output for the new `<script>` blocks — no entity escaping leaked into the JS — and `go build ./...` plus `go test ./internal/webui/handlers/...` pass on this head.

Three non-blocking follow-ups, all inline: a residual race in `ExtendSilence` that can still create a duplicate silence and report success, HTTP status mapping for client-side errors in the extend/expire handlers, and the new ordering rule having no test. None of them block merge.



internal/webui/handlers/silence_handlers.go
SoulKyu: @SoulKyu I'm not making a code change for this thread.

Already implemented on this branch by commit 7591935: compileMatchers/compiledMatcher hoist regex compilation out of the alert loop (silence_handlers.go:25-111) and GetSilences skips countMatchedAlerts for expired silences (silence_handlers.go:177-182). No further change needed this round.




internal/webui/handlers/silence_handlers.go
SoulKyu: @SoulKyu I'm not making a code change for this thread.

Already implemented on this branch by commit 7591935: the base=now branch is gone and ExtendSilence returns 409 for an expired silence (silence_handlers.go:282-285), with the doc comment corrected to state that Alertmanager's canUpdate mints a new silence instead of reviving one.




internal/webui/handlers/silence_handlers.go
SoulKyu: @SoulKyu I'm not making a code change for this thread.

Already implemented on this branch by commit 7591935: the comparator groups live before expired and orders expired descending, and sort.Strings(sources) stabilises the dropdown (silence_handlers.go:186-206). This round only extracted that comparator into sortSilences and added a test for it.




internal/alertmanager/client.go
SoulKyu: @SoulKyu I'm not making a code change for this thread.

Already implemented on this branch by commit 7591935: FetchAllSilences was deleted from internal/alertmanager/client.go along with the arbitrary-map-entry loop; a repo-wide grep for FetchAllSilences now matches only FetchAllSilencesDetailed and its single caller.




internal/webui/handlers/silence_handlers_test.go
SoulKyu: @SoulKyu I'm not making a code change for this thread.

Already implemented on this branch by commit 7591935: TestCountMatchedAlertsIsScopedToSource exists at silence_handlers_test.go:105-124, asserting 1 match for prod-am, 1 for staging-am and 0 for an unknown source, with the webuimodels import in place.




internal/webui/handlers/silence_handlers.go
SoulKyu: @SoulKyu This has been fixed in a47a3d8.

ExtendSilence now compares the ID returned by CreateSilenceOnAlertmanager against the requested silenceID and answers 409 when they differ, catching the window where the silence lapsed between our GET and Alertmanager's Set (silence_handlers.go:295-302). It then re-reads the silence so the 200 payload carries Alertmanager's own status/updatedAt instead of the pre-POST snapshot, falling back to the local copy on a read error.




internal/webui/handlers/silence_handlers.go
SoulKyu: @SoulKyu This has been fixed in a47a3d8.

Both ExtendSilence and ExpireSilence validate the source against GetClientNames with slices.Contains and return 400 for an unknown Alertmanager (silence_handlers.go:269-272, 334-337). Added ErrSilenceNotFound in internal/alertmanager/client.go, wrapped at both 404 sites in FetchSilence and DeleteSilence, and matched via errors.Is in the new silenceErrorStatus helper so a stale silence ID returns 404, leaving 502 for genuine upstream faults.




internal/webui/handlers/silence_handlers_test.go
SoulKyu: @SoulKyu This has been fixed in a47a3d8.

Lifted the comparator out of GetSilences into sortSilences (silence_handlers.go:161-176, called at line 189) and added TestSortSilencesPutsLiveFirstAndExpiredNewestFirst to silence_handlers_test.go, which pins live-before-expired, live ascending on EndsAt and expired descending in one assertion. go test ./... passes.




SoulKyu:
**Looper fixer round complete** — a47a3d8

- ⏸️ Review comment on `internal/webui/handlers/silence_handlers.go` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/86#discussion_r3649614885)
  - Already implemented on this branch by commit 7591935: compileMatchers/compiledMatcher hoist regex compilation out of the alert loop (silence_handlers.go:25-111) and GetSilences skips countMatchedAlerts for expired silences (silence_handlers.go:177-182). No further change needed this round.
- ⏸️ Review comment on `internal/webui/handlers/silence_handlers.go` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/86#discussion_r3649614886)
  - Already implemented on this branch by commit 7591935: the base=now branch is gone and ExtendSilence returns 409 for an expired silence (silence_handlers.go:282-285), with the doc comment corrected to state that Alertmanager's canUpdate mints a new silence instead of reviving one.
- ⏸️ Review comment on `internal/webui/handlers/silence_handlers.go` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/86#discussion_r3649614887)
  - Already implemented on this branch by commit 7591935: the comparator groups live before expired and orders expired descending, and sort.Strings(sources) stabilises the dropdown (silence_handlers.go:186-206). This round only extracted that comparator into sortSilences and added a test for it.
- ⏸️ Review comment on `internal/alertmanager/client.go` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/86#discussion_r3649614888)
  - Already implemented on this branch by commit 7591935: FetchAllSilences was deleted from internal/alertmanager/client.go along with the arbitrary-map-entry loop; a repo-wide grep for FetchAllSilences now matches only FetchAllSilencesDetailed and its single caller.
- ⏸️ Review comment on `internal/webui/handlers/silence_handlers_test.go:101` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/86#discussion_r3649614889)
  - Already implemented on this branch by commit 7591935: TestCountMatchedAlertsIsScopedToSource exists at silence_handlers_test.go:105-124, asserting 1 match for prod-am, 1 for staging-am and 0 for an unknown source, with the webuimodels import in place.
- ✅ Review comment on `internal/webui/handlers/silence_handlers.go:312` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/86#discussion_r3649634053)
  - ExtendSilence now compares the ID returned by CreateSilenceOnAlertmanager against the requested silenceID and answers 409 when they differ, catching the window where the silence lapsed between our GET and Alertmanager's Set (silence_handlers.go:295-302). It then re-reads the silence so the 200 payload carries Alertmanager's own status/updatedAt instead of the pre-POST snapshot, falling back to the local copy on a read error.
- ✅ Review comment on `internal/webui/handlers/silence_handlers.go:278` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/86#discussion_r3649634054)
  - Both ExtendSilence and ExpireSilence validate the source against GetClientNames with slices.Contains and return 400 for an unknown Alertmanager (silence_handlers.go:269-272, 334-337). Added ErrSilenceNotFound in internal/alertmanager/client.go, wrapped at both 404 sites in FetchSilence and DeleteSilence, and matched via errors.Is in the new silenceErrorStatus helper so a stale silence ID returns 404, leaving 502 for genuine upstream faults.
- ✅ Review comment on `internal/webui/handlers/silence_handlers_test.go:164` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/86#discussion_r3649634056)
  - Lifted the comparator out of GetSilences into sortSilences (silence_handlers.go:161-176, called at line 189) and added TestSortSilencesPutsLiveFirstAndExpiredNewestFirst to silence_handlers_test.go, which pins live-before-expired, live ascending on EndsAt and expired descending in one assertion. go test ./... passes.


SoulKyu:
## 🧪 QA Report

Env: `make docker-build-all` + `docker compose up -d` (fake Alertmanager, backend, webui on :8081). All UI criteria driven with a real headless Chromium (Playwright); API criteria with `curl`. Seeded 4 extra silences on the fake Alertmanager (soon-expiring / zero-match / already-expired / wide regex) on top of the built-in one.

|---|---|---|
| `/silences` reachable from the nav for a logged-in user, lists silences from every configured Alertmanager with a source column | ✅ | Logged in as `qaagent`, clicked the new **Silences** tab in `nav.page-navigator` on `/dashboard` → landed on `/silences`; table rendered with a `SOURCE` column (`Fake Alertmanager`). Anonymous `GET /silences` → `302 → /login`. |
| Row shows matchers, creator, comment, state, absolute end time **and** relative countdown; active silences <1h flagged | ✅ | Row: `Fake Alertmanager \| alertname=DiskSpaceLow \| oncall@example.com \| QA soon-expiring silence \| Active + Expiring soon \| 7/25/2026, 9:04:02 AM / ends in 28m \| 2 alert(s)`. Negated/regex matchers render as `alertname=~.*`. |
| Default sort soonest-expiry-first; expired hidden until the state filter includes them | ✅ | Default view = 4 rows ordered `ends in 28m → 1h 57m → 4h 58m → 1d 23h`. The `QA already expired` silence was absent; clicking the **Expired** state button added it as a 5th row (`ended 11m ago`), pushed to the bottom. |
| Matched-alert count per active silence; zero-match marked as suppressing nothing | ✅ | `alertname=NoSuchAlertXYZ` → `0 alert(s)` + amber **Suppressing nothing**; `alertname=~.*` → `23 alert(s)`; `alertname=DiskSpaceLow` → `2 alert(s)`. Counts match the alert cache. |
| Extend (+1h/+4h/+24h) updates in place on the correct Alertmanager, new end time visible after refresh, ID unchanged | ✅ | Clicked **+1h** on `QA suppressing nothing`. Alertmanager before: `9d21c5c4-…  endsAt 11:34:02Z` → after: `9d21c5c4-…  endsAt 12:34:02Z` (**same ID**, delta exactly 60 min). UI reloaded itself and showed `2:34:02 PM / ends in 5h 58m`. |
| Expire now removes the silence from the correct Alertmanager only; row leaves the active view | ✅ | Clicked **Expire now** on `QA long regex silence`, accepted the confirm dialog → row count for it went 1 → 0, Alertmanager silence count 5 → 4, silence gone upstream. Per-source targeting proven separately: with two sources configured, `DELETE /silences/9d21c5c4-…?source=Dead%20Alertmanager` returned `502` and the silence **survived** on `Fake Alertmanager` — the delete is not fanned out. |
| One unreachable Alertmanager → reachable sources still render, failing one named, no blanket error | ✅ | Restarted webui with a second source `Dead Alertmanager → http://alertmanager-dead:9093`. Page rendered 3 rows plus an amber banner: *"Alertmanager **Dead Alertmanager** is unreachable — its silences are not listed"* with the dial error. Header read `3 of 4 silences • 2 alertmanager(s)`; source dropdown listed both. |
| Silenced alert's modal shows its silences with creator, comment and end time | ✅ | Opened a silenced `DiskSpaceLow` alert → **Details** tab → *Silences* section shows both silences with creator, source, comment, matcher chips and `Ends 7/25/2026, 10:33:02 AM (in 1h 55m)`. `GET /api/v1/dashboard/alert/<fp>` returns both silence objects (`details.Silences` no longer `[]`). |
| No `proto/` changes and no DB migration | ✅ | `git diff main --name-only` touches only `internal/**` — nothing under `proto/`, no migration/schema file. |
| `make webui-templates` regenerated; no hand-edited `*_templ.go` | ✅ | Re-ran `templ generate` with the pinned `v0.3.906` in a clean container: `updates=125`, `git status --porcelain` → clean. Generated files are byte-identical to the committed ones. |

**Exploratory:** the three-tab `PageNavigator` did not break the adjacent pages — `/statistics` still loads through it (tabs `Alerts / Silences / Statistics`), and the existing dashboard silence-creation flow still works end to end: selected a `HighCPUUsage` alert → **Silence** modal → 1 hour + reason → the new silence appears in the inventory with its 5 matchers, creator `55clhqxl…`, `1 alert(s)` matched and an **Expiring soon** badge. No JS/page errors on `/silences`, `/dashboard` or the modal. `go test ./internal/webui/handlers/... ./internal/models/...` → 16 passed.

**Verdict: PASS**

PR #94, opened 2026-07-25
perf(webui): count resolved alerts instead of fetching them for the Resolved counter

### LINKED ISSUE #92: perf(dashboard): every classic-mode dashboard request fetches up to 1000 full resolved-alert blobs over gRPC just to render the Resolved counter
## Problem

`buildDashboardMetadata` (`internal/webui/handlers/dashboard_handlers.go:708-727`) contains this block for the **default** display modes (`classic` and `acknowledge`): it calls `alertCache.GetResolvedAlertsWithLimit(...)` / `GetResolvedAlerts()`, which does a gRPC `GetResolvedAlerts` round-trip to the backend (`internal/webui/services/alert_cache.go:551-580`). The backend runs `SELECT *` over `resolved_alerts` (`internal/backend/database/gorm_db.go:517-533`) — every row carrying the full `alert_data`, `comments` and `acknowledgments` JSONB blobs — ships them over gRPC, and the webui JSON-unmarshals each blob into a `DashboardAlert`.

All of that work feeds exactly one thing: `counters.Resolved++` (lines 719-726). The alerts themselves are discarded.

This runs on **every** `GetDashboardData` request (page load, filter/sort/page/mode change — `dashboard_handlers.go:205`) and on every incremental poll via `getDashboardMetadata` → `buildDashboardMetadata` (`dashboard_handlers.go:1046-1075`, called from `processIncremental:1206`). In the SSE-fallback polling mode that is every 5-60 s **per browser** (`dashboard_core.templ`, base interval 5000 ms). The frontend always sends `resolvedAlertsLimit` (default 100, user-configurable up to 1000 in the settings modal — `modal_components.templ:90`), so each request moves 100-1000 blob rows to produce a single integer.

Bonus defect: the counter is **wrong** anyway — it counts at most `resolvedAlertsLimit` rows, so with 2500 resolved alerts in retention the tile shows 100.

## Impact

- Every dashboard paint in the default view blocks on a backend round-trip + N JSONB decodes that contribute one number. This sits on the critical path of first paint alongside the hidden-alerts fetch (#78).
- Backend/DB load scales as (dashboard interactions + poll rate) × row limit × blob size — the dominant per-request cost once #78's cache lands.
- Users with a large `resolvedAlertsLimit` (1000) pay ~10× the default cost per request without knowing.

## Proposed approach

Use a count instead of the rows. The RPC response already carries `total_count` populated from `GetResolvedAlertsCount()` (`internal/backend/services/services.go:1113`), so the webui can get the true count with a `limit=1` call — or cache it in `AlertCache` with a short TTL, invalidated at the end of a refresh cycle that resolved alerts (the webui is the sole writer of `resolved_alerts`, via `storeResolvedAlertInBackend`).

Caveat to preserve: today the count is taken **after** `applyDashboardFilters` (line 717), so with active search/severity/team filters the resolved counter is filtered. Cheapest faithful behavior: use the cached/total count when no filters are active (the overwhelmingly common case), and keep the current fetch+filter path only when filters are set. This also fixes the counter cap in the unfiltered case.

## Acceptance criteria

- [ ] A classic-mode `GET /api/v1/dashboard/data` with no active filters issues **no** `GetResolvedAlerts` row-fetch (verifiable via backend logs) and still shows a correct Resolved counter.
- [ ] The Resolved tile shows the true retention-window count, not `min(count, resolvedAlertsLimit)`, when no filters are active.
- [ ] Filtered views (search/severity/team) keep their current filtered-resolved-count semantics.
- [ ] Incremental polling responses (`/api/v1/dashboard/incremental`) get the same treatment.
- [ ] `go build ./...` passes.

SoulKyu:
## Problem

In classic and acknowledge display modes, `buildDashboardMetadata` fetched up to `resolvedAlertsLimit` (100 by default, up to 1000) resolved alerts over gRPC — each row carrying its full `alert_data`, `comments` and `acknowledgments` JSONB blobs, JSON-unmarshalled webui-side — only to increment `counters.Resolved` and throw the alerts away.

This ran on every `/api/v1/dashboard/data` request (page load, filter/sort/page/mode change) and on every `/api/v1/dashboard/incremental` poll (every 5-60s per browser in SSE-fallback mode).

The counter was wrong anyway: capped at `resolvedAlertsLimit`, so 2500 resolved alerts in retention displayed as 100.

## Change

- `BackendClient.GetResolvedAlertsCount()` — `GetResolvedAlerts` with `limit=1`, reads `total_count` from the response. No blobs on the wire.
- `AlertCache.GetResolvedAlertsCount()` — 30s TTL cache over that call, invalidated by `storeResolvedAlertInBackend` and `RemoveAllResolvedAlerts` (the webui is the sole writer of `resolved_alerts`). Falls back to the last known value on backend error.
- `filtersAffectResolvedCount()` — mirrors every rejection in `applyDashboardFilters` (search, alertmanagers, severities, statuses, teams, alert names, acknowledged, hasComments, per-filter hidden alerts/rules, plus global hidden entries for the session).
- `buildDashboardMetadata` uses the cached count when no filter can exclude a resolved alert; otherwise it keeps the existing fetch-and-filter path unchanged.
- `HiddenAlertsService.HasHiddenEntries(sessionID)` — cheap "does this session hide anything at all" check.

Both `GetDashboardData` and the incremental path go through `buildDashboardMetadata`, so both get the treatment.

## Effect

- Unfiltered classic/acknowledge requests: zero resolved-alert row fetches, one cheap `SELECT count(*)` at most every 30s.
- The Resolved tile now shows the true retention-window count instead of `min(count, resolvedAlertsLimit)`.
- Filtered views keep their current filtered-resolved-count semantics.

## Validation

- `go build ./...` — passes
- `go vet ./internal/webui/...` — clean
- `go test ./internal/webui/...` — 86 passed, including a new table test pinning `filtersAffectResolvedCount` against every filter that `applyDashboardFilters` acts on (a filter it fails to recognize would silently show an unfiltered counter in a filtered view).

Closes #92

### DIFF
--- a/internal/backend/services/services.go
+++ b/internal/backend/services/services.go
@@ -1162,11 +1162,17 @@ func (s *AlertServiceGorm) GetResolvedAlerts(ctx context.Context, req *alertpb.G
 		}, nil
 	}
 
-	// Get total count
+	// Get total count. Falling back to len(resolvedAlerts) here would be
+	// indistinguishable from a genuine count for callers that only read
+	// total_count (a limit=1 count request would silently get 1), so a count
+	// failure fails the whole response instead.
 	totalCount, err := s.db.GetResolvedAlertsCount()
 	if err != nil {
 		log.Printf("Error getting resolved alerts count: %v", err)
-		totalCount = int64(len(resolvedAlerts))
+		return &alertpb.GetResolvedAlertsResponse{
+			Success: false,
+			Message: "Failed to count resolved alerts",
+		}, nil
 	}
 
 	// Convert to protobuf messages
--- a/internal/backend/services/services.go
+++ b/internal/backend/services/services.go
@@ -1164,15 +1164,13 @@ func (s *AlertServiceGorm) GetResolvedAlerts(ctx context.Context, req *alertpb.G
 
 	// Get total count. Falling back to len(resolvedAlerts) here would be
 	// indistinguishable from a genuine count for callers that only read
-	// total_count (a limit=1 count request would silently get 1), so a count
-	// failure fails the whole response instead.
+	// total_count (a limit=1 count request would silently get 1), so a failed
+	// count reports -1 instead: the rows we did fetch stay usable, and the only
+	// reader of total_count rejects the negative value.
 	totalCount, err := s.db.GetResolvedAlertsCount()
 	if err != nil {
 		log.Printf("Error getting resolved alerts count: %v", err)
-		return &alertpb.GetResolvedAlertsResponse{
-			Success: false,
-			Message: "Failed to count resolved alerts",
-		}, nil
+		totalCount = -1
 	}
 
 	// Convert to protobuf messages
--- a/internal/webui/client/backend_client.go
+++ b/internal/webui/client/backend_client.go
@@ -553,6 +553,11 @@ func (c *BackendClient) GetResolvedAlertsCount() (int, error) {
 		return 0, fmt.Errorf("failed to get resolved alerts count: %s", resp.Message)
 	}
 
+	// The backend sends -1 when the rows are good but the count query failed.
+	if resp.TotalCount < 0 {
+		return 0, fmt.Errorf("backend could not count resolved alerts")
+	}
+
 	return int(resp.TotalCount), nil
 }
 
--- a/internal/webui/client/backend_client.go
+++ b/internal/webui/client/backend_client.go
@@ -533,6 +533,29 @@ func (c *BackendClient) GetResolvedAlerts(limit, offset int) ([]*alertpb.Resolve
 	return resp.ResolvedAlerts, nil
 }
 
+// GetResolvedAlertsCount returns the total number of resolved alerts kept in the
+// retention window. It asks for a single row and reads total_count from the
+// response, so no alert blobs travel over the wire.
+func (c *BackendClient) GetResolvedAlertsCount() (int, error) {
+	if c.alertClient == nil {
+		return 0, fmt.Errorf("not connected to backend")
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	resp, err := c.alertClient.GetResolvedAlerts(ctx, &alertpb.GetResolvedAlertsRequest{Limit: 1})
+	if err != nil {
+		return 0, err
+	}
+
+	if !resp.Success {
+		return 0, fmt.Errorf("failed to get resolved alerts count: %s", resp.Message)
+	}
+
+	return int(resp.TotalCount), nil
+}
+
 // GetResolvedAlert retrieves a specific resolved alert by fingerprint
 func (c *BackendClient) GetResolvedAlert(fingerprint string) (*alertpb.ResolvedAlertInfo, error) {
 	if c.alertClient == nil {
--- a/internal/webui/handlers/dashboard_handlers.go
+++ b/internal/webui/handlers/dashboard_handlers.go
@@ -369,6 +369,27 @@ func getAcknowledgedAlerts() []*webuimodels.DashboardAlert {
 	return acknowledgedAlerts
 }
 
+// filtersAffectResolvedCount reports whether applyDashboardFilters could drop a
+// resolved alert for these filters. When it cannot, the Resolved counter is just
+// the backend's total and no resolved alert needs to be fetched at all.
+// Mirrors every rejection in applyDashboardFilters.
+func filtersAffectResolvedCount(filters webuimodels.DashboardFilters, sessionID string) bool {
+	if filters.Search != "" ||
+		len(filters.Alertmanagers) > 0 ||
+		len(filters.Severities) > 0 ||
+		len(filters.Statuses) > 0 ||
+		len(filters.Teams) > 0 ||
+		len(filters.AlertNames) > 0 ||
+		filters.Acknowledged != nil ||
+		filters.HasComments != nil ||
+		len(filters.FilterHiddenAlerts) > 0 ||
+		len(filters.FilterHiddenRules) > 0 {
+		return true
+	}
+
+	return hiddenAlertsService != nil && hiddenAlertsService.HasHiddenEntries(sessionID)
+}
+
 func applyDashboardFilters(alerts []*webuimodels.DashboardAlert, filters webuimodels.DashboardFilters, sessionID string) []*webuimodels.DashboardAlert {
 	var filtered []*webuimodels.DashboardAlert
 
@@ -708,20 +729,27 @@ func buildDashboardMetadata(allAlerts, filteredAlerts []*webuimodels.DashboardAl
 	// Always include resolved alerts in statistics, even if not displayed (for Classic view)
 	// Only do this if we're not already in DisplayModeResolved or DisplayModeFull to avoid double counting
 	if filters.DisplayMode == webuimodels.DisplayModeClassic || filters.DisplayMode == webuimodels.DisplayModeAcknowledge {
-		var resolvedAlerts []*webuimodels.DashboardAlert
-		if filters.ResolvedAlertsLimit > 0 {
-			resolvedAlerts = alertCache.GetResolvedAlertsWithLimit(filters.ResolvedAlertsLimit)
+		if !filtersAffectResolvedCount(filters, sessionID) {
+			// Nothing can exclude a resolved alert, so the backend count is the
+			// answer - and it is the true retention-window count, not one capped
+			// at ResolvedAlertsLimit. No alert blobs travel over gRPC for this.
+			counters.Resolved = alertCache.GetResolvedAlertsCount()
 		} else {
-			resolvedAlerts = alertCache.GetResolvedAlerts()
-		}
-		filteredResolvedAlerts := applyDashboardFilters(resolvedAlerts, filters, sessionID)
+			var resolvedAlerts []*webuimodels.DashboardAlert
+			if filters.ResolvedAlertsLimit > 0 {
+				resolvedAlerts = alertCache.GetResolvedAlertsWithLimit(filters.ResolvedAlertsLimit)
+			} else {
+				resolvedAlerts = alertCache.GetResolvedAlerts()
+			}
+			filteredResolvedAlerts := applyDashboardFilters(resolvedAlerts, filters, sessionID)
 
-		for _, alert := range filteredResolvedAlerts {
-			// Only count resolved alerts in the Resolved counter for Classic/Acknowledge views
-			// Exclude them from Critical, Warning, Info, Acknowledged, WithComments counters
-			// since these views don't display resolved alerts in their main lists
-			if alert.Status.State == "resolved" {
-				counters.Resolved++
+			for _, alert := range filteredResolvedAlerts {
+				// Only count resolved alerts in the Resolved counter for Classic/Acknowledge views
+				// Exclude them from Critical, Warning, Info, Acknowledged, WithComments counters
+				// since these views don't display resolved alerts in their main lists
+				if alert.Status.State == "resolved" {
+					counters.Resolved++
+				}
 			}
 		}
 	}
--- a/internal/webui/handlers/dashboard_handlers.go
+++ b/internal/webui/handlers/dashboard_handlers.go
@@ -735,6 +735,12 @@ func buildDashboardMetadata(allAlerts, filteredAlerts []*webuimodels.DashboardAl
 			// at ResolvedAlertsLimit. No alert blobs travel over gRPC for this.
 			counters.Resolved = alertCache.GetResolvedAlertsCount()
 		} else {
+			// Same contract as the fast path: this tile is the resolved store, not
+			// resolved rows that happened to survive into filteredAlerts. A no-op on
+			// today's inputs, but it stops the two branches from disagreeing if an
+			// alert ever reaches the active cache already carrying a resolved state.
+			counters.Resolved = 0
+
 			var resolvedAlerts []*webuimodels.DashboardAlert
 			if filters.ResolvedAlertsLimit > 0 {
 				resolvedAlerts = alertCache.GetResolvedAlertsWithLimit(filters.ResolvedAlertsLimit)
--- a/internal/webui/handlers/dashboard_resolved_count_test.go
+++ b/internal/webui/handlers/dashboard_resolved_count_test.go
@@ -1,10 +1,37 @@
 package handlers
 
 import (
+	"reflect"
 	"testing"
 
+	alertpb "notificator/internal/backend/proto/alert"
 	webuimodels "notificator/internal/webui/models"
+	"notificator/internal/webui/services"
 )
+
+// stubHiddenBackend feeds HiddenAlertsService without a backend connection. Only
+// the two read calls LoadUserData makes are ever exercised.
+type stubHiddenBackend struct {
+	rules []*alertpb.UserHiddenRule
+}
+
+func (s stubHiddenBackend) GetUserHiddenAlerts(string, ...string) ([]*alertpb.UserHiddenAlert, error) {
+	return nil, nil
+}
+
+func (s stubHiddenBackend) GetUserHiddenRules(string, ...string) ([]*alertpb.UserHiddenRule, error) {
+	return s.rules, nil
+}
+
+func (s stubHiddenBackend) HideAlert(string, string, string, string, string, ...string) error {
+	return nil
+}
+func (s stubHiddenBackend) UnhideAlert(string, string, ...string) error { return nil }
+func (s stubHiddenBackend) SaveHiddenRule(string, *alertpb.UserHiddenRule, ...string) (*alertpb.UserHiddenRule, error) {
+	return nil, nil
+}
+func (s stubHiddenBackend) RemoveHiddenRule(string, string, ...string) error { return nil }
+func (s stubHiddenBackend) ClearAllHiddenAlerts(string, ...string) error     { return nil }
 
 // filtersAffectResolvedCount gates the cheap backend-count path against the
 // expensive fetch-and-filter path: a filter it fails to recognize would silently
@@ -42,5 +69,56 @@ func TestFiltersAffectResolvedCount(t *testing.T) {
 				t.Errorf("filtersAffectResolvedCount() = %v, want %v", got, tt.want)
 			}
 		})
+	}
+}
+
+// The table above is hand-written per field, so it stays green when
+// DashboardFilters grows a field that applyDashboardFilters rejects on and
+// filtersAffectResolvedCount ignores. This forces the decision at compile-test
+// time instead.
+func TestFiltersAffectResolvedCountCoversEveryFilterField(t *testing.T) {
+	known := map[string]bool{
+		"Search": true, "Alertmanagers": true, "Severities": true,
+		"Statuses": true, "Teams": true, "AlertNames": true,
+		"Acknowledged": true, "HasComments": true,
+		"FilterHiddenAlerts": true, "FilterHiddenRules": true,
+		// Not filtering predicates in applyDashboardFilters:
+		"DisplayMode": true, "ViewMode": true, "ResolvedAlertsLimit": true,
+	}
+
+	typ := reflect.TypeOf(webuimodels.DashboardFilters{})
+	for i := 0; i < typ.NumField(); i++ {
+		if name := typ.Field(i).Name; !known[name] {
+			t.Fatalf("DashboardFilters.%s is new: decide whether it can exclude a resolved alert, then update filtersAffectResolvedCount and this list", name)
+		}
+	}
+}
+
+// The global hidden-alerts branch is the one filtersAffectResolvedCount case
+// that does not come from the filters struct: a session hiding anything at all
+// must take the fetch-and-filter path.
+func TestFiltersAffectResolvedCountGlobalHiddenRules(t *testing.T) {
+	previous := hiddenAlertsService
+	t.Cleanup(func() { hiddenAlertsService = previous })
+
+	const sessionID = "session-with-hidden-rule"
+
+	hiddenAlertsService = services.NewHiddenAlertsService(stubHiddenBackend{})
+	if filtersAffectResolvedCount(webuimodels.DashboardFilters{}, sessionID) {
+		t.Error("no hidden entries: want false")
+	}
+
+	hiddenAlertsService = services.NewHiddenAlertsService(stubHiddenBackend{
+		rules: []*alertpb.UserHiddenRule{{Id: "r1", LabelKey: "team", LabelValue: "sre", IsEnabled: true}},
+	})
+	if !filtersAffectResolvedCount(webuimodels.DashboardFilters{}, sessionID) {
+		t.Error("one enabled hidden rule: want true")
+	}
+
+	hiddenAlertsService = services.NewHiddenAlertsService(stubHiddenBackend{
+		rules: []*alertpb.UserHiddenRule{{Id: "r1", LabelKey: "team", LabelValue: "sre", IsEnabled: false}},
+	})
+	if filtersAffectResolvedCount(webuimodels.DashboardFilters{}, sessionID) {
+		t.Error("only a disabled hidden rule: want false")
 	}
 }
--- a/internal/webui/handlers/dashboard_resolved_count_test.go
+++ b/internal/webui/handlers/dashboard_resolved_count_test.go
@@ -1,0 +1,46 @@
+package handlers
+
+import (
+	"testing"
+
+	webuimodels "notificator/internal/webui/models"
+)
+
+// filtersAffectResolvedCount gates the cheap backend-count path against the
+// expensive fetch-and-filter path: a filter it fails to recognize would silently
+// report an unfiltered Resolved counter in a filtered view.
+func TestFiltersAffectResolvedCount(t *testing.T) {
+	acknowledged := true
+	hasComments := false
+
+	tests := []struct {
+		name    string
+		filters webuimodels.DashboardFilters
+		want    bool
+	}{
+		{"no filters", webuimodels.DashboardFilters{}, false},
+		{"resolved limit alone does not filter", webuimodels.DashboardFilters{ResolvedAlertsLimit: 100}, false},
+		{"search", webuimodels.DashboardFilters{Search: "disk"}, true},
+		{"alertmanagers", webuimodels.DashboardFilters{Alertmanagers: []string{"prod"}}, true},
+		{"severities", webuimodels.DashboardFilters{Severities: []string{"critical"}}, true},
+		{"statuses", webuimodels.DashboardFilters{Statuses: []string{"firing"}}, true},
+		{"teams", webuimodels.DashboardFilters{Teams: []string{"sre"}}, true},
+		{"alert names", webuimodels.DashboardFilters{AlertNames: []string{"HighLoad"}}, true},
+		{"acknowledged", webuimodels.DashboardFilters{Acknowledged: &acknowledged}, true},
+		{"has comments", webuimodels.DashboardFilters{HasComments: &hasComments}, true},
+		{"filter hidden alerts", webuimodels.DashboardFilters{
+			FilterHiddenAlerts: []webuimodels.FilterHiddenAlert{{Fingerprint: "abc"}},
+		}, true},
+		{"filter hidden rules", webuimodels.DashboardFilters{
+			FilterHiddenRules: []webuimodels.FilterHiddenRule{{LabelKey: "team"}},
+		}, true},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := filtersAffectResolvedCount(tt.filters, ""); got != tt.want {
+				t.Errorf("filtersAffectResolvedCount() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
--- a/internal/webui/services/alert_cache.go
+++ b/internal/webui/services/alert_cache.go
@@ -65,6 +65,7 @@ type AlertCache struct {
 	resolvedCountMu      sync.RWMutex
 	resolvedCount        int
 	resolvedCountFetched time.Time
+	resolvedCountGen     uint64 // bumped by every invalidation, so an in-flight fetch can tell its result is stale
 
 	// Configuration
 	refreshInterval       time.Duration
@@ -603,6 +604,7 @@ func (ac *AlertCache) GetResolvedAlertsWithLimit(limit int) []*webuimodels.Dashb
 func (ac *AlertCache) GetResolvedAlertsCount() int {
 	ac.resolvedCountMu.RLock()
 	count := ac.resolvedCount
+	gen := ac.resolvedCountGen
 	fresh := !ac.resolvedCountFetched.IsZero() && time.Since(ac.resolvedCountFetched) < resolvedCountTTL
 	ac.resolvedCountMu.RUnlock()
 
@@ -622,19 +624,33 @@ func (ac *AlertCache) GetResolvedAlertsCount() int {
 		return count
 	}
 
-	ac.resolvedCountMu.Lock()
-	ac.resolvedCount = fetched
-	ac.resolvedCountFetched = time.Now()
-	ac.resolvedCountMu.Unlock()
+	ac.publishResolvedCount(gen, fetched)
 
 	return fetched
 }
 
+// publishResolvedCount caches fetched unless an invalidation landed while the
+// fetch was in flight. Without the generation check, a count read before
+// RemoveAllResolvedAlerts would be published after it and served as fresh for a
+// full resolvedCountTTL. Same pattern as HiddenAlertsService.LoadUserData.
+func (ac *AlertCache) publishResolvedCount(gen uint64, fetched int) {
+	ac.resolvedCountMu.Lock()
+	defer ac.resolvedCountMu.Unlock()
+
+	if ac.resolvedCountGen != gen {
+		return
+	}
+
+	ac.resolvedCount = fetched
+	ac.resolvedCountFetched = time.Now()
+}
+
 // InvalidateResolvedAlertsCount forces the next GetResolvedAlertsCount call to
-// re-query the backend.
+// re-query the backend, and marks any in-flight fetch as pre-mutation.
 func (ac *AlertCache) InvalidateResolvedAlertsCount() {
 	ac.resolvedCountMu.Lock()
 	ac.resolvedCountFetched = time.Time{}
+	ac.resolvedCountGen++
 	ac.resolvedCountMu.Unlock()
 }
 
--- a/internal/webui/services/alert_cache.go
+++ b/internal/webui/services/alert_cache.go
@@ -42,6 +42,12 @@ type alertFetcher interface {
 // so a large diff cannot stampede the backend with thousands of gRPC calls.
 const maxBackendWorkers = 8
 
+// resolvedCountTTL bounds how long a resolved-alerts count is served without a
+// backend round-trip. The webui is the sole writer of resolved_alerts and
+// invalidates on every write, so the TTL only covers retention expiry and other
+// webui replicas.
+const resolvedCountTTL = 30 * time.Second
+
 type AlertCache struct {
 	mu                 sync.RWMutex
 	alerts             map[string]*webuimodels.DashboardAlert // fingerprint -> alert
@@ -54,6 +60,11 @@ type AlertCache struct {
 	// Color caching - keyed by userID then fingerprint
 	colorsMutex  sync.RWMutex
 	cachedColors map[string]map[string]*AlertColorResult // userID -> fingerprint -> color result
+
+	// Resolved-alerts count caching (dashboard counters only need the number)
+	resolvedCountMu      sync.RWMutex
+	resolvedCount        int
+	resolvedCountFetched time.Time
 
 	// Configuration
 	refreshInterval       time.Duration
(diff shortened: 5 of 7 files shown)


SoulKyu:
## 🧪 QA Report

Env: `make test` docker-compose stack (postgres + backend + fake alertmanager + webui on :8081), fresh build of this branch. GORM logs every SQL statement at `Info`, so backend container logs are the ground truth for what the DB actually did. Seeded 7 resolved alerts (5 `critical`/`sre`, 2 `warning`/`dba`) directly in `resolved_alerts`.

|---|---|---|
| Unfiltered classic `GET /api/v1/dashboard/data` issues no `GetResolvedAlerts` row-fetch, counter still correct | ✅ | Request at 12:50:57 produced only `SELECT * FROM "resolved_alerts" ... LIMIT 1` + `SELECT count(*) FROM "resolved_alerts"` in backend logs — no `LIMIT 100` fetch. `counters.resolved = 7` (DB count = 7). Two follow-up requests inside the 30 s TTL issued **zero** `resolved_alerts` queries. |
| Resolved tile shows the true retention count, not `min(count, resolvedAlertsLimit)` | ✅ | `...&resolvedAlertsLimit=2` with 7 rows in retention → `counters.resolved = 7` (pre-PR this returned 2). Browser (Playwright, real login) rendered the tile as `Resolved 7`. |
| Filtered views keep filtered-resolved-count semantics | ✅ | `severities=critical` → 5, `teams=dba` → 2, `search=QASeedAlert3` → 1; each of those three requests logged the old `... ORDER BY resolved_at DESC LIMIT 100` fetch, confirming the fetch-and-filter path is still taken. Browser: typing `QASeedAlert3` into `#dashboard-search` flipped the tile to `Resolved 1`. |
| Incremental polling gets the same treatment | ✅ | `POST /api/v1/dashboard/incremental?displayMode=classic` → `resolved = 7` with only `LIMIT 1` + `count(*)` in the logs. Same call with `severities=critical` → `resolved = 5` and the `LIMIT 100` fetch reappears — filtered semantics preserved on the incremental path too. |
| `go build ./...` passes | ✅ | `go build ./...` clean; `go test ./internal/webui/...` → `ok handlers`, `ok services`. |

**Exploratory:** the riskiest new branch is `HasHiddenEntries` — hiding one resolved alert (`POST /api/v1/dashboard/hidden-alerts` on `qafp1`) correctly dropped the fast path: the counter went 7 → 6 and the `LIMIT 100` fetch came back in the logs; unhiding restored 7 and the cheap path. `displayMode=acknowledge` behaves identically to `classic`. Resolved-mode listing (`displayMode=resolved`) still returns all 7 rows. No regression found; the browser's 401s on the notification endpoints predate this change.

Note (not a blocker): the "cheap" path still ships one row — `GetResolvedAlertsCount()` calls `GetResolvedAlerts` with `limit=1`, so a single `alert_data` blob crosses the wire at most once per 30 s TTL. That is exactly what #92 proposed ("get the true count with a `limit=1` call"), and it replaces 100–1000 blobs per request.

**Verdict: PASS**

internal/webui/services/alert_cache.go
```diff
@@ -585,6 +596,48 @@ func (ac *AlertCache) GetResolvedAlertsWithLimit(limit int) []*webuimodels.Dashb
...
+	fetched, err := ac.backendClient.GetResolvedAlertsCount()
+	if err != nil {
+		log.Printf("Error fetching resolved alerts count from backend: %v", err)
+		return count
+	}
+
+	ac.resolvedCountMu.Lock()
+	ac.resolvedCount = fetched
+	ac.resolvedCountFetched = time.Now()
+	ac.resolvedCountMu.Unlock()
```
SoulKyu: **An invalidation that lands while this fetch is in flight is silently overwritten, and the stale count is then served as fresh for a full 30s.**

Problem: the read of the cache, the backend call, and the publish are three separate critical sections with no epoch check between them. `InvalidateResolvedAlertsCount` only zeroes `resolvedCountFetched`; it has no way to tell the in-flight fetch that its result predates a mutation.

Failure scenario:
1. T0 — a dashboard request misses the TTL and calls `ac.backendClient.GetResolvedAlertsCount()`; the backend runs `count(*)` and gets 2500.
2. T0+50ms — the operator hits `DELETE /dashboard/remove-resolved-alerts`; `AlertCache.RemoveAllResolvedAlerts` succeeds and calls `InvalidateResolvedAlertsCount()`, setting `resolvedCountFetched` to the zero time.
3. T0+80ms — the in-flight call returns 2500 and unconditionally writes `resolvedCount = 2500` and `resolvedCountFetched = time.Now()`.
4. The Resolved tile keeps showing 2500 for the next 30s even though the table is now empty, and every browser polling `/api/v1/dashboard/incremental` sees the same wrong value. The same window applies to `storeResolvedAlertInBackend`, where the counter just lags by one for 30s.

Why it matters: the invalidation calls added in this PR are the only thing making the 30s TTL acceptable for the single-writer case, and this window quietly defeats them for exactly the most user-visible mutation (clear-all), right after a deliberate user action.

Suggested change: use the generation-counter pattern this codebase already applies to the same class of bug in `HiddenAlertsService.LoadUserData` (it snapshots `s.generation[sessionID]` before the fetch and drops the snapshot if it changed). Concretely, add a `resolvedCountGen uint64` field next to `resolvedCount`, bump it in `InvalidateResolvedAlertsCount`, capture it in the same `RLock` block that reads `count`/`fresh`, and only publish when it is unchanged:

```go
ac.resolvedCountMu.Lock()
if ac.resolvedCountGen == gen {
	ac.resolvedCount = fetched
	ac.resolvedCountFetched = time.Now()
}
ac.resolvedCountMu.Unlock()
```

Returning `fetched` to the caller either way is fine — the point is not to cache a pre-mutation value for another TTL.


internal/webui/client/backend_client.go
```diff
@@ -533,6 +533,29 @@ func (c *BackendClient) GetResolvedAlerts(limit, offset int) ([]*alertpb.Resolve
...
+	resp, err := c.alertClient.GetResolvedAlerts(ctx, &alertpb.GetResolvedAlertsRequest{Limit: 1})
+	if err != nil {
+		return 0, err
+	}
+
+	if !resp.Success {
+		return 0, fmt.Errorf("failed to get resolved alerts count: %s", resp.Message)
+	}
+
+	return int(resp.TotalCount), nil
```
SoulKyu, line 561: **`TotalCount` is not trustworthy on its own: when the backend's count query fails it returns `Success: true` with `TotalCount` set to the number of rows in the page — which is 1 for this `Limit: 1` request.**

Problem: `AlertServiceGorm.GetResolvedAlerts` (internal/backend/services/services.go, the `// Get total count` block) does `totalCount, err := s.db.GetResolvedAlertsCount()` and on error logs it and falls back to `totalCount = int64(len(resolvedAlerts))`, then still returns `Success: true`. Until this PR nobody read `TotalCount` from this RPC on the webui side, so the fallback was dormant. Now it is the only value this method returns, and it is indistinguishable from a genuine count of 1.

Failure scenario: `resolved_alerts` grows large and a `statement_timeout` (or any transient DB error) kills the `SELECT count(*) WHERE expires_at > now()` while the `ORDER BY resolved_at DESC LIMIT 1` index scan still succeeds — the more rows there are, the more likely this asymmetry is. The backend answers `Success: true, TotalCount: 1`; `GetResolvedAlertsCount` returns `1, nil`; `AlertCache.GetResolvedAlertsCount` treats it as a good value, overwrites the previously correct count and caches it for 30s. The Resolved tile reads `1` with nothing in the webui logs to explain it, because the error path here is never taken.

Why it matters: this is a silent wrong-number-with-no-signal, and it caches, so it survives well past the transient failure. It also defeats the `return count` fallback right below the caller, which is designed to keep the last known good value on error.

Suggested change: fix the contract at the source in `AlertServiceGorm.GetResolvedAlerts` — drop the `totalCount = int64(len(resolvedAlerts))` fallback and return `Success: false` with a message like `"Failed to count resolved alerts"` when `s.db.GetResolvedAlertsCount()` errors. The existing `if !resp.Success` check in this new method then surfaces it correctly, and `AlertCache.GetResolvedAlertsCount` keeps serving the last known value. If you would rather not change the shared RPC's behavior for existing callers, the alternative is a dedicated count RPC so the count error is never conflated with a successful page fetch.


internal/webui/handlers/dashboard_resolved_count_test.go
```diff
@@ -0,0 +1,46 @@
...
+		{"severities", webuimodels.DashboardFilters{Severities: []string{"critical"}}, true},
+		{"statuses", webuimodels.DashboardFilters{Statuses: []string{"firing"}}, true},
+		{"teams", webuimodels.DashboardFilters{Teams: []string{"sre"}}, true},
+		{"alert names", webuimodels.DashboardFilters{AlertNames: []string{"HighLoad"}}, true},
+		{"acknowledged", webuimodels.DashboardFilters{Acknowledged: &acknowledged}, true},
+		{"has comments", webuimodels.DashboardFilters{HasComments: &hasComments}, true},
+		{"filter hidden alerts", webuimodels.DashboardFilters{
+			FilterHiddenAlerts: []webuimodels.FilterHiddenAlert{{Fingerprint: "abc"}},
+		}, true},
+		{"filter hidden rules", webuimodels.DashboardFilters{
```
SoulKyu, line 61: **This table pins the filters that exist today, but it cannot fail for the regression the test's own doc comment describes.**

Problem: the comment above the test says "a filter it fails to recognize would silently report an unfiltered Resolved counter in a filtered view" — that is drift between `applyDashboardFilters` and `filtersAffectResolvedCount`. But every case here is hand-written against a known field, so adding a new field to `webuimodels.DashboardFilters` plus a new `continue` in `applyDashboardFilters` and forgetting `filtersAffectResolvedCount` leaves all 12 subtests green. The test locks in what is already correct and is blind to the failure mode it was written for.

Failure scenario: someone adds `Silenced *bool` to `DashboardFilters` with a matching rejection in `applyDashboardFilters` but not in `filtersAffectResolvedCount`. `go test ./internal/webui/...` stays green, and every classic-mode request with only `silenced=false` set takes the fast path and shows the full retention-window count instead of the filtered one — a wrong counter in a filtered view, exactly what this test claims to guard.

Why it matters: the mirror is hand-maintained across two files with no compiler link between them, and the fast path is silently wrong rather than loudly broken, so a test that actually fails on drift is the only cheap protection.

Suggested change: add a reflection-based field-coverage guard next to this test so a new `DashboardFilters` field forces a decision:

```go
// Fails when DashboardFilters gains a field, forcing an explicit decision about
// whether it can exclude a resolved alert in applyDashboardFilters.
func TestFiltersAffectResolvedCountCoversEveryFilterField(t *testing.T) {
	known := map[string]bool{
		"Search": true, "Alertmanagers": true, "Severities": true,
		"Statuses": true, "Teams": true, "AlertNames": true,
		"Acknowledged": true, "HasComments": true,
		"FilterHiddenAlerts": true, "FilterHiddenRules": true,
		// Not filtering predicates in applyDashboardFilters:
		"DisplayMode": true, "ViewMode": true, "ResolvedAlertsLimit": true,
	}
	typ := reflect.TypeOf(webuimodels.DashboardFilters{})
	for i := 0; i < typ.NumField(); i++ {
		if name := typ.Field(i).Name; !known[name] {
			t.Fatalf("DashboardFilters.%s is new: decide whether it can exclude a resolved alert and update filtersAffectResolvedCount and this list", name)
		}
	}
}
```

While you are here, the global-hidden branch (`hiddenAlertsService.HasHiddenEntries`) is also uncovered — every case runs with the package-level `hiddenAlertsService` nil, so the `return hiddenAlertsService != nil && ...` line is only ever exercised on its short-circuit. A case that sets the package-level service to an instance with one enabled rule for a session and asserts `true` would close that gap.


SoulKyu (commented):
Solid perf fix — the hot path really does drop from "fetch up to 1000 JSONB blobs and unmarshal them" to "one cached SELECT count(*) every 30s", and I verified the semantics line up: `GormDB.GetResolvedAlerts` and `GormDB.GetResolvedAlertsCount` both scope to `expires_at > now()`, so the cheap count covers exactly the same retention window the old fetch-and-filter path walked. I also checked that `filtersAffectResolvedCount` mirrors every rejection in `applyDashboardFilters` (search, alertmanagers, severities, statuses, teams, alert names, acknowledged, hasComments, per-filter hidden alerts/rules, plus global hidden via the new `HasHiddenEntries`) — nothing is missing today. The `counters.Resolved = ...` assignment is safe because in classic/acknowledge mode `filteredAlerts` comes from `getStandardAlerts()`/`getAcknowledgedAlerts()`, which never contain resolved alerts, so the loop-accumulated value is always 0 there. `go build ./...`, `go vet ./internal/webui/...` and `go test ./internal/webui/...` all pass locally on this head.

Three non-blocking items below: a lost-invalidation window in the new count cache, an error-masking edge in the `total_count` contract the new client method now depends on, and a drift guard for the hand-maintained filter mirror.



internal/webui/client/backend_client.go
SoulKyu: @SoulKyu This has been fixed in 674b7b8.

Fixed the contract at the source as suggested: AlertServiceGorm.GetResolvedAlerts (internal/backend/services/services.go:1165) no longer falls back to len(resolvedAlerts) and returns Success: false with "Failed to count resolved alerts" when s.db.GetResolvedAlertsCount() errors, so AlertCache.GetResolvedAlertsCount keeps serving its last known value. Tradeoff noted in a code comment: a count-only failure now fails the page fetch too.




internal/webui/handlers/dashboard_resolved_count_test.go
SoulKyu: @SoulKyu This has been fixed in 674b7b8.

Added TestFiltersAffectResolvedCountCoversEveryFilterField (reflection guard over DashboardFilters, verified failing when a Silenced *bool field is added) and TestFiltersAffectResolvedCountGlobalHiddenRules, which sets the package-level hiddenAlertsService via a stub backend and asserts false/true/false for no rules, one enabled rule, and one disabled rule.




SoulKyu:
## 🧪 QA Report

Re-test after the hardening commit `674b7b8` (previous report was on `46d3830`). Env: `make test` docker-compose stack (postgres + backend + fake alertmanager + webui on :8081), fresh build of this head. GORM logs every statement at `Info`, so backend container logs are the ground truth for what the DB actually did. Seeded resolved alerts directly in `resolved_alerts` (7, then 150 for the cap test).

|---|---|---|
| Unfiltered classic `GET /api/v1/dashboard/data` issues no `GetResolvedAlerts` row-fetch, counter still correct | ✅ | Only `SELECT * FROM "resolved_alerts" … LIMIT 1` + `SELECT count(*) FROM "resolved_alerts"` in the backend logs — no `LIMIT 100`. `counters.resolved = 7` (DB count = 7). Two follow-up requests inside the 30 s TTL issued **zero** `resolved_alerts` queries. |
| Resolved tile shows the true retention count, not `min(count, resolvedAlertsLimit)` | ✅ | 151 rows in retention, default `resolvedAlertsLimit=100`: browser (Playwright, real login) rendered the tile as `Resolved 151`. Also `&resolvedAlertsLimit=2` with 7 rows → `resolved = 7`. Pre-PR both would cap. |
| Filtered views keep filtered-resolved-count semantics | ✅ | `severities=critical` → 5, `teams=dba` → 2, `search=QASeedAlert3` → 1; each logged the old `ORDER BY resolved_at DESC LIMIT 100` fetch, so the fetch-and-filter path is still taken. Browser: typing `QASeedAlert3` into `#dashboard-search` with 150 seeded → tile `11` (`…3`, `…30-39`), clearing it → back to `151`. |
| Incremental polling gets the same treatment | ✅ | `POST /api/v1/dashboard/incremental?displayMode=classic` → `resolved = 7`, logs show only `LIMIT 1` + `count(*)`. Same call with `severities=critical` → `5` and the `LIMIT 100` fetch reappears. |
| `go build ./...` passes | ✅ | `go build ./...` clean, `go vet ./internal/webui/...` clean, `go test ./internal/webui/... ./internal/backend/...` → all `ok`. The three new tests run and pass (`TestFiltersAffectResolvedCountCoversEveryFilterField`, `…GlobalHiddenRules`, `TestAlertCache_ResolvedCountInvalidationDuringFetch`). |

### Hardening added by `674b7b8`

| Check | Result | Evidence |
|---|---|---|
| Invalidation is not lost — clear-all publishes 0, not the pre-mutation count | ✅ | Warm cache to `8`, `DELETE /api/v1/dashboard/remove-resolved-alerts` (as an impersonation-allowed user), DB rows → 0, next request → `resolved = 0` with a fresh `count(*)`. |
| Generation guard: a count fetch in flight when an invalidation lands must not be published | ✅ | 5 rounds of racing a TTL-expired `GET /dashboard/data` against `DELETE /remove-resolved-alerts` (jittered 5–45 ms offsets): the in-flight request saw `7`, and the post-clear request returned `0` every round — never a stale `7` served as fresh. Without the guard the stale value would be pinned for the full 30 s TTL. |
| A failing `GetResolvedAlerts` no longer yields a bogus count; the webui keeps its last known value | ⚠️ ✅ | Made `resolved_alerts` unreachable while the rest of the DB stayed up (`ALTER TABLE … RENAME`), TTL expired: backend logged `Error fetching resolved alerts: relation "resolved_alerts" does not exist`, returned `Success: false`, and the dashboard request stayed healthy (`success: true`) serving the last known `resolved = 8` — not `0`, not `1`. **Caveat:** a count-only failure (the exact case the diff changes, where `len(resolvedAlerts)` would have been returned as `1`) is not inducible black-box — the same table backs both the row query and the count query, and the row query returns first. Both failure branches return the identical `Success: false` shape the webui fallback handles, which is what I exercised. |
| Storing a newly resolved alert invalidates the count | ✅ | Real alerts resolving in the fake-alertmanager stream moved the counter `8 → 9` on the very next request rather than staying stale for the TTL. |

**Exploratory:** the `HasHiddenEntries` gate still behaves — hiding one resolved alert (`POST /api/v1/dashboard/hidden-alerts` on `qafp1`) dropped the fast path (counter `9 → 8`, `LIMIT 100` fetch back in the logs), and `DELETE /hidden-alerts` restored `9` on the cheap path. `displayMode=acknowledge` behaves identically to `classic`; `displayMode=resolved` still lists rows in the browser. The two 401s the page throws (`/api/v1/profile/timezone`, `/api/impersonate/status`) predate this change. No regression found.

Non-blocker, my env not the diff: `make test` needs `templ` on `PATH`, and generating with `templ@latest` produces code the go.mod-pinned `templ v0.3.906` runtime cannot compile (`undefined: templ.ResolveAttributeValue`) — regenerating with `templ@v0.3.906` fixes it.

**Verdict: PASS**

SoulKyu:

internal/backend/services/services.go
```diff
@@ -1162,11 +1162,17 @@ func (s *AlertServiceGorm) GetResolvedAlerts(ctx context.Context, req *alertpb.G
...
+	// total_count (a limit=1 count request would silently get 1), so a count
+	// failure fails the whole response instead.
 	totalCount, err := s.db.GetResolvedAlertsCount()
 	if err != nil {
 		log.Printf("Error getting resolved alerts count: %v", err)
-		totalCount = int64(len(resolvedAlerts))
+		return &alertpb.GetResolvedAlertsResponse{
+			Success: false,
+			Message: "Failed to count resolved alerts",
+		}, nil
```
SoulKyu: **A count-only failure now silently blanks the resolved-alerts list, not just the counter.**

Problem: `BackendClient.GetResolvedAlerts` turns `!resp.Success` into an error (internal/webui/client/backend_client.go:529-531), and `AlertCache.GetResolvedAlertsWithPagination` turns any error from that call into `return []*webuimodels.DashboardAlert{}` (internal/webui/services/alert_cache.go:663-667). So this new early return does not merely fail the count — it empties the alert list for every caller of the RPC.

Failure scenario: `resolved_alerts` grows large and a `statement_timeout` (or any transient DB error) kills `SELECT count(*) WHERE expires_at > now()` while the `ORDER BY resolved_at DESC LIMIT 100` scan on the same table still succeeds — the asymmetry gets *more* likely the more rows there are, which is exactly the deployment this PR is optimising for. Before this change: the resolved view rendered its rows and only the total was wrong. After: `GetDashboardData` in `DisplayModeResolved`, `DisplayModeFull` and `DisplayModeHidden` (internal/webui/handlers/dashboard_handlers.go:139-167) each receive an empty slice, and the UI renders an empty resolved list that is indistinguishable from a genuinely empty retention window. The only trace is a log line on the backend.

Why it matters: the previous round asked to stop conflating a count failure with a good count, and this does fix that — `AlertCache.GetResolvedAlertsCount` now correctly keeps its last known value. But it pays for that by turning a wrong number into silently missing data across three display modes, for a failure that has nothing to do with the rows that were fetched successfully. The comment above reads as "the count is unavailable, so the response fails"; the user-visible effect is "the resolved alerts disappeared".

Suggested change: keep the page successful and signal the count failure out of band. `TotalCount` is `int32`, so a sentinel needs no proto change and no new RPC. Here:

```go
totalCount, err := s.db.GetResolvedAlertsCount()
if err != nil {
	log.Printf("Error getting resolved alerts count: %v", err)
	// -1 means "the rows are good, the count is not". The only reader of
	// total_count rejects it; the display path keeps its rows.
	totalCount = -1
}
```

and in `BackendClient.GetResolvedAlertsCount` (internal/webui/client/backend_client.go:553), right before the return:

```go
if resp.TotalCount < 0 {
	return 0, fmt.Errorf("backend could not count resolved alerts")
}
```

`AlertCache.GetResolvedAlertsCount` then hits its existing `return count` fallback and serves the last known value exactly as it does today, while the resolved list stops depending on a query it never needed.


internal/webui/services/alert_cache_test.go
```diff
@@ -1015,3 +1015,25 @@ func TestAlertCache_MutateAlert(t *testing.T) {
 		t.Errorf("accessor returned a cache-resident pointer: CommentCount = %d, want 1", again.CommentCount)
 	}
 }
+
+func TestAlertCache_ResolvedCountInvalidationDuringFetch(t *testing.T) {
```
SoulKyu, line 1019: **The count getter on `AlertCache` is never called by any test — its TTL gate and its last-known-value fallback are both uncovered.**

Problem: this test drives `publishResolvedCount` directly, which pins the generation check and nothing else. The getter itself (internal/webui/services/alert_cache.go:604-628) has no caller anywhere in the suite — the only references in that package are its own definition and the client call inside it. That leaves two documented behaviours untested: the freshness gate at the top of the getter, which is the perf win this PR exists for, and the doc comment's promise that on backend failure the last known value is returned.

Failure scenario: someone tidies the freshness condition and drops the zero-timestamp guard. Tests stay green, but a cache that has never fetched has a zero fetch timestamp, so the elapsed-time check reads as decades and every request takes the backend path. The counter stays correct while the entire optimisation is silently dead, and the invalidation bookkeeping around it becomes decorative. The mirror case is just as quiet: regressing the disconnected-backend branch from returning the cached value to returning zero blanks the Resolved tile whenever the backend blips, and again nothing fails.

Why it matters: both branches are pure cache logic whose only observable effects are a number on a tile and a query that does or does not happen. A regression in either surfaces as "the tile went to zero" or "the query is back" in production rather than in CI — and the invalidation test sitting right here reads as if the count path is covered when it is not.

Suggested change: neither branch touches the backend client before returning, so the nil construction already used on line 1020 exercises both. No interface extraction needed. Add a second test beside this one that:

- seeds the cached count to a known value and stamps the fetch time to now, then asserts the getter returns that value — this covers the cache-hit path;
- calls the invalidation method and asserts the getter still returns the same value rather than zero — this covers the fallback, since invalidation only clears the timestamp and leaves the cached number in place, and the nil backend short-circuits before any fetch.

Sketch:

```go
func TestResolvedCountCacheAndFallback(t *testing.T) {
	cache := NewAlertCache(nil, nil, 90, 10*time.Second)

	cache.resolvedCount = 42
	cache.resolvedCountFetched = time.Now()
	if got := cache.GetResolvedAlertsCount(); got != 42 {
		t.Errorf("cache hit: got %d, want 42", got)
	}

	cache.InvalidateResolvedAlertsCount()
	if got := cache.GetResolvedAlertsCount(); got != 42 {
		t.Errorf("stale, no backend: got %d, want 42", got)
	}
}
```

I traced both assertions against the current code: the first exits at the freshness check, the second at the nil-backend guard. Both pass today and both fail under the regressions above.


internal/webui/handlers/dashboard_handlers.go
```diff
@@ -708,20 +729,27 @@ func buildDashboardMetadata(allAlerts, filteredAlerts []*webuimodels.DashboardAl
...
 	// Only do this if we're not already in DisplayModeResolved or DisplayModeFull to avoid double counting
 	if filters.DisplayMode == webuimodels.DisplayModeClassic || filters.DisplayMode == webuimodels.DisplayModeAcknowledge {
-		var resolvedAlerts []*webuimodels.DashboardAlert
-		if filters.ResolvedAlertsLimit > 0 {
-			resolvedAlerts = alertCache.GetResolvedAlertsWithLimit(filters.ResolvedAlertsLimit)
+		if !filtersAffectResolvedCount(filters, sessionID) {
+			// Nothing can exclude a resolved alert, so the backend count is the
+			// answer - and it is the true retention-window count, not one capped
+			// at ResolvedAlertsLimit. No alert blobs travel over gRPC for this.
+			counters.Resolved = alertCache.GetResolvedAlertsCount()
```
SoulKyu, line 736: **NIT — the fast path assigns `counters.Resolved` while the slow path adds to it, so the two branches are not equivalent by construction.**

Problem: the counting loop earlier in this function (internal/webui/handlers/dashboard_handlers.go:684-715) already does `counters.Resolved++` for every alert in `filteredAlerts` whose `Status.State == "resolved"`. This new fast path discards that with `=`; the `else` branch below still increments on top of it. For the same inputs the two branches differ by the number of resolved alerts already present in `filteredAlerts`, even though they are meant to compute the same tile.

Why it is only a nit today: in `DisplayModeClassic` the input comes from `getStandardAlerts()` (dashboard_handlers.go:346-357), which drops anything with `IsResolved`, so the pre-count is always 0. `DisplayModeAcknowledge` is the interesting one — `getAcknowledgedAlerts()` (dashboard_handlers.go:359-369) filters on `IsAcknowledged` alone and does *not* exclude resolved alerts; it relies entirely on `refreshAlerts` calling `delete(ac.alerts, fingerprint)` the instant an alert resolves (internal/webui/services/alert_cache.go:273). Since `transformStatus` passes unrecognised states straight through (alert_cache.go:18-25), an upstream alert arriving already carrying `status.state: "resolved"` would sit in the active cache as acknowledged-and-resolved, and the two branches would then disagree. The Alertmanager v2 alerts endpoint does not emit that state, so this is latent coupling rather than a live bug — but the correctness of `=` currently rests on a `delete` three files away, which is not obvious to the next reader.

Suggested change: make the branches agree explicitly. The fast path's contract — "in Classic/Acknowledge the Resolved tile is the retention-store total, full stop" — is the clearer of the two, so align the slow path to it rather than the reverse:

```go
} else {
	// Same contract as the fast path: this tile is the resolved store, not
	// resolved rows that happened to survive into filteredAlerts.
	counters.Resolved = 0

	var resolvedAlerts []*webuimodels.DashboardAlert
```

That is a no-op on today's inputs and removes the dependency on `getAcknowledgedAlerts` never seeing a resolved alert.


SoulKyu (commented):
Second pass on this after the fixer round — the two previously flagged items are genuinely closed: `AlertServiceGorm.GetResolvedAlerts` no longer fabricates a count from `len(resolvedAlerts)`, and `publishResolvedCount`'s generation check correctly drops a pre-mutation count when an invalidation lands mid-flight (I walked the interleavings, including two concurrent fetches straddling an invalidation, and they all resolve the right way). I also re-derived the filter mirror against `applyDashboardFilters`: all ten rejection predicates are covered, the reflection guard sees all 13 `DashboardFilters` fields, and the `DisplayModeHidden` inversion is unreachable from this call site. `go build ./...` and `go test ./internal/webui/...` pass locally on this head, and I confirmed nothing else consumes `counters.Resolved` beyond the display span in `NewDashboard.templ`, so the now-uncapped number can't create phantom pagination.

Three things left, none of them blockers:

1. The backend-side hardening has a wider blast radius than the code comment suggests — a count-only failure now empties the resolved-alerts list in three display modes, not just the counter.
2. `AlertCache.GetResolvedAlertsCount` is never actually called by a test, so the TTL gate and the last-known-value fallback — the two behaviours this PR exists for — are uncovered, and both are testable with the nil-backend construction already in the file.
3. A small consistency nit between the fast and slow branches in `buildDashboardMetadata`.

Details inline. Nice work on the shape of this overall — moving the counter off the row fetch is clearly the right call, and the generation-counter pattern reused from `HiddenAlertsService` is exactly the right precedent to follow.




internal/backend/services/services.go
SoulKyu: @SoulKyu This has been fixed in 0e19032.

internal/backend/services/services.go:1165 now sets totalCount = -1 on count failure and keeps Success: true, so the fetched rows still reach the caller; internal/webui/client/backend_client.go:552 rejects a negative TotalCount with an error, so AlertCache.GetResolvedAlertsCount hits its existing last-known-value fallback while the resolved list is unaffected. Confirmed backend_client.go:556 is the only reader of that total_count.




internal/webui/services/alert_cache_test.go
SoulKyu: @SoulKyu This has been fixed in 0e19032.

Added TestAlertCache_ResolvedCountCacheAndFallback in internal/webui/services/alert_cache_test.go: it seeds resolvedCount/resolvedCountFetched and asserts the getter serves the cached value, then invalidates and asserts it still returns that value rather than 0 with a nil backend. Both assertions pass and fail under the described regressions.




internal/webui/handlers/dashboard_handlers.go
SoulKyu: @SoulKyu This has been fixed in 0e19032.

internal/webui/handlers/dashboard_handlers.go:738 now resets counters.Resolved = 0 at the top of the else branch before summing filtered resolved alerts, so both branches compute the resolved-store total rather than adding to resolved rows already counted in filteredAlerts. No-op on today's inputs; removes the dependency on getAcknowledgedAlerts never seeing a resolved alert.




SoulKyu:
**Looper fixer round complete** — 0e19032

- ✅ Review comment on `internal/backend/services/services.go` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/94#discussion_r3650277087)
  - internal/backend/services/services.go:1165 now sets totalCount = -1 on count failure and keeps Success: true, so the fetched rows still reach the caller; internal/webui/client/backend_client.go:552 rejects a negative TotalCount with an error, so AlertCache.GetResolvedAlertsCount hits its existing last-known-value fallback while the resolved list is unaffected. Confirmed backend_client.go:556 is the only reader of that total_count.
- ✅ Review comment on `internal/webui/services/alert_cache_test.go:1019` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/94#discussion_r3650277089)
  - Added TestAlertCache_ResolvedCountCacheAndFallback in internal/webui/services/alert_cache_test.go: it seeds resolvedCount/resolvedCountFetched and asserts the getter serves the cached value, then invalidates and asserts it still returns that value rather than 0 with a nil backend. Both assertions pass and fail under the described regressions.
- ✅ Review comment on `internal/webui/handlers/dashboard_handlers.go:736` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/94#discussion_r3650277091)
  - internal/webui/handlers/dashboard_handlers.go:738 now resets counters.Resolved = 0 at the top of the else branch before summing filtered resolved alerts, so both branches compute the resolved-store total rather than adding to resolved rows already counted in filteredAlerts. No-op on today's inputs; removes the dependency on getAcknowledgedAlerts never seeing a resolved alert.


SoulKyu:
## 🧪 QA Report

Re-test after the rebase onto `main` (previous reports: `46d3830`, `674b7b8`). The rebase brought in exactly one upstream commit — `27dd195 chore: gitignore Python bytecode from factory-tui` — and `git diff 674b7b8 f148c49` is `.gitignore | 4 ++++` only, so **the PR's code diff is byte-identical to the head that already passed**. Re-verified end-to-end anyway against a freshly built stack.

Env: `docker compose` stack built from this head (postgres + backend + fake alertmanager + webui on :8081). GORM logs every statement at `Info`, so backend container logs are the ground truth for what the DB actually did. Resolved alerts seeded directly into `resolved_alerts`.

|---|---|---|
| Unfiltered classic `GET /api/v1/dashboard/data` issues **no** `GetResolvedAlerts` row-fetch, counter still correct | ✅ | With 7 rows seeded, the request logged only `SELECT * FROM "resolved_alerts" … LIMIT 1` + `SELECT count(*) FROM "resolved_alerts"` — no `LIMIT 100`. `counters.resolved = 7` (DB count 7). Two follow-up requests inside the 30 s TTL issued **0** `resolved_alerts` queries. |
| Resolved tile shows the true retention count, not `min(count, resolvedAlertsLimit)` | ✅ | 151 rows in retention, default `resolvedAlertsLimit=100`: browser (Playwright, real login as `qaagent`) rendered the tile as **151**. Also `&resolvedAlertsLimit=2` with 7 rows → `resolved = 7`. Pre-PR both would cap. |
| Filtered views keep their current filtered-resolved-count semantics | ✅ | `severities=critical` → 5, `teams=dba` → 2, `search=QASeedAlert3` → 1; each logged the old `ORDER BY resolved_at DESC LIMIT 100` fetch, so the fetch-and-filter path is still taken. Browser: typing `ExpAlert3` into `#dashboard-search` flipped the tile `6 → 1`, clearing it → back to `6`. |
| Incremental polling gets the same treatment | ✅ | `POST /api/v1/dashboard/incremental?displayMode=classic` → `resolved = 7`, logs show only `LIMIT 1` + `count(*)`. Same call with `severities=critical` → `5` and the `LIMIT 100` fetch reappears. |
| `go build ./...` passes | ✅ | `go build ./...` clean, `go vet ./internal/webui/...` clean, `go test ./internal/webui/... ./internal/backend/...` → all `ok`. The four PR tests run and pass: `TestFiltersAffectResolvedCount`, `…CoversEveryFilterField`, `…GlobalHiddenRules`, `TestAlertCache_ResolvedCountInvalidationDuringFetch`. |

### Hardening from `f508c71`+`f148c49` re-verified

| Check | Result | Evidence |
|---|---|---|
| Clear-all invalidates immediately, no stale value pinned for the TTL | ✅ | Warmed cache to `152`, `DELETE /api/v1/dashboard/remove-resolved-alerts` (user in `NOTIFICATOR_ADMIN_IMPERSONATION_ALLOWED_USERS`), DB → 0 rows, very next request → `resolved = 0`. |
| Generation guard: an in-flight count fetch must not publish over an invalidation that landed after it | ✅ | 4 rounds racing a TTL-expired `GET /dashboard/data` against `DELETE /remove-resolved-alerts` (12/24/36/48 ms offsets): in-flight saw `9`, post-clear returned `0` every round. Without the guard the stale `9` would be served as fresh for the full 30 s TTL. |
| A failing backend count yields no bogus number; webui keeps its last known value | ✅ | Warmed to `8`, then `ALTER TABLE resolved_alerts RENAME …` with the rest of the DB up, TTL expired: backend logged `relation "resolved_alerts" does not exist` and returned `Success: false`; the dashboard request stayed healthy (`success: true`) serving the last-known `8` — not `0`, not `1`. *Caveat (unchanged from the last report): the count-only failure — the exact branch `services.go` changes, where `len(resolvedAlerts)` would have been returned as `1` — is not inducible black-box, since the same table backs both the row query and the count query and the row query errors first. Both branches return the identical `Success: false` shape the webui fallback handles, which is what I exercised.* |
| `HasHiddenEntries` gate | ✅ | Hiding one resolved alert (`POST /dashboard/hidden-alerts` on `qafp1`) dropped the fast path — the `LIMIT 100` fetch reappeared in the logs; `DELETE /hidden-alerts` restored the cheap path and the counter returned to `151`. |

**Exploratory:** `displayMode=resolved` still lists every row (6/6 with correct names and `critical 3 / warning 3` counters) — the `services.go` change did not disturb the list path. `displayMode=acknowledge` behaves identically to `classic`. Newly resolved alerts from the live fake-alertmanager stream moved the counter `151 → 152` on the next request rather than going stale, confirming the `storeResolvedAlertInBackend` invalidation. The two 401s the page throws (`/api/v1/profile/timezone`, `/api/impersonate/status`) predate this change. No regression found.

Non-blocking observation (pre-existing, explicitly in scope per the issue's "keep current filtered semantics"): the *filtered* path is still capped at `resolvedAlertsLimit`, so with 151 rows in retention `search=QASeedAlert3` returns `0` at `limit=100` (that alert ranks 145th by `resolved_at`) but the correct `1` at `limit=1000`. Identical to pre-PR behaviour, but now more visible as an asymmetry, since the unfiltered counter is uncapped: applying any filter can make the tile drop from `151` to a ≤100-row windowed count. Worth a follow-up issue if that reads as confusing to users.

**Verdict: PASS**

SoulKyu:
## 🧪 QA Report

Re-test after `0e19032 fix(webui): keep the resolved list alive when only the count query fails` (previous reports: `46d3830`, `674b7b8`, `f148c49`). Env: `docker compose` stack built from this head (postgres + backend + fake alertmanager + webui on :8081). GORM logs every statement at `Info`, so backend container logs are ground truth for what the DB actually did. Resolved alerts seeded directly into `resolved_alerts` (7, then 151).

**This head makes the previously-untestable branch testable, and I got it.** The last two reports carried a caveat: a *count-only* failure — the exact case `services.go` changes — was "not inducible black-box". It is: a view over the real table with a `WHERE` predicate that inspects `current_query()` and raises only for the `count(*)` statement discriminates the two queries perfectly (row query returns rows, count query errors). Results below, including a build of the parent commit to prove the counterfactual.

|---|---|---|
| Unfiltered classic `GET /api/v1/dashboard/data` issues **no** `GetResolvedAlerts` row-fetch, counter still correct | ✅ | 7 rows seeded: the request logged only `SELECT * FROM "resolved_alerts" … LIMIT 1` + `SELECT count(*) FROM "resolved_alerts"` — no `LIMIT 100`. `counters.resolved = 7` (DB count 7). Two follow-up requests inside the 30 s TTL issued **0** `resolved_alerts` queries. |
| Resolved tile shows the true retention count, not `min(count, resolvedAlertsLimit)` | ✅ | 151 rows, default `resolvedAlertsLimit=100`: browser (Playwright, real login as `qaagent`) rendered the tile as **Resolved 151** — later **152** after one more row landed. Also `&resolvedAlertsLimit=2` with 7 rows → `resolved = 7`. Pre-PR both would cap. |
| Filtered views keep their current filtered-resolved-count semantics | ✅ | `severities=critical` → 5, `teams=dba` → 2, `search=QASeedAlert3` → 1 (7-row seed); each logged the old `ORDER BY resolved_at DESC LIMIT 100` fetch, so the fetch-and-filter path is still taken. Browser: typing `QASeedAlert3` into `#dashboard-search` with 151 seeded flipped the tile `151 → 11` (`…3`, `…30-39`), clearing it → back to `151`. |
| Incremental polling gets the same treatment | ✅ | `POST /api/v1/dashboard/incremental?displayMode=classic` → `resolved = 7`, logs show only `LIMIT 1` + `count(*)`, zero `LIMIT 100`. Same call with `severities=critical` → `5` and the `LIMIT 100` fetch reappears. |
| `go build ./...` passes | ✅ | `go build ./...` clean, `go vet ./internal/webui/...` clean, `go test ./internal/webui/... ./internal/backend/...` → **102 passed / 17 packages**. The new `TestAlertCache_ResolvedCountCacheAndFallback` and the four earlier PR tests all run and pass. |

### The fix in `0e19032`, exercised for real

| Check | Result | Evidence |
|---|---|---|
| A count-only DB failure no longer blanks the resolved list | ✅ | Gate installed (rows OK, `count(*)` raises), TTL expired. Backend logged `Error getting resolved alerts count: … QA injected failure` — and `GET /dashboard/data?displayMode=resolved` still returned **all 7 alerts** (`QASeedAlert1…7`), `success: true`. |
| …and the counter falls back to the last known value, never to `-1`, `0` or `1` | ✅ | Same failure state, webui logged `Error fetching resolved alerts count from backend: backend could not count resolved alerts` (the `TotalCount < 0` rejection in `backend_client.go`) and served the last-known `resolved = 7` in `classic` **and** `acknowledge`. `full` / `hidden` modes unaffected. No negative number reached any response field. |
| Counterfactual: the parent commit really does blank the list | ✅ | Built `f148c49`'s backend as `notificator-backend:preqa` in a throwaway worktree and swapped it into the stack under the identical gate: `displayMode=resolved` returned **0 alerts of 7**, webui logging `Error fetching resolved alerts from backend: failed to get resolved alerts: Failed to count resolved alerts`. Swapped the PR backend back → 7 alerts again. The regression is real and this commit closes it. |
| Slow-path `counters.Resolved = 0` reset is a no-op on today's inputs | ✅ | Every filtered count matched the DB exactly (5 / 2 / 1 on the 7-row seed, 11 on the 151-row seed, 6 for `acknowledge&severities=critical`) — no double counting, no undercount. |

**Exploratory:** `HasHiddenEntries` gate still works — hiding `qafp1` dropped the fast path (counter `151 → 150`, the `LIMIT 1000` fetch back in the logs), unhiding restored the cheap path and `151`. `displayMode=resolved` lists normally (page of 50 with the counter at 151). After a new row landed, the counter moved `151 → 152` on the first request past the TTL and matched `SELECT count(*)` exactly, with zero `Error fetching resolved alerts` lines in the webui log. Clear-all invalidation was not re-run on this head (`remove-resolved-alerts` needs a user in `NOTIFICATOR_ADMIN_IMPERSONATION_ALLOWED_USERS`, and that path is untouched by this commit — verified ✅ on the two previous heads). No regression found.

**Verdict: PASS**

PR #99, opened 2026-07-25
fix(webui): surface Sentry test and settings-save outcomes in the modal

### LINKED ISSUE #97: fix(webui): Sentry token "Test" button and settings-save failures report only to the browser console
## Problem

In the dashboard Settings modal, the Sentry integration's **Test** button exists solely to tell the user whether their token works — and it never tells them anything. `testSentryConnection()` (`internal/webui/templates/scripts/dashboard_settings.templ:801-836`) resolves the check into `console.log('Connection Successful')` / `console.error('Connection Failed')` and nothing else. The button UI (`internal/webui/templates/components/modal_components.templ:466-479`) only toggles a spinner: after "Testing…" finishes, a valid token and an invalid token look pixel-identical.

The surrounding save paths have the same hole:

- `saveSentryToken()` failure branch is literally empty (`dashboard_settings.templ:792-793`): the backend rejects the token, the form keeps the value, no message appears — while on success the field clears, so failure is easily *mistaken* for success by a user who glances away.
- "💾 Save All Settings" (`modal_components.templ:636-648` → `saveSettings()`, `dashboard_settings.templ:441-487`): any failure (color prefs, notification prefs, …) lands in `catch { console.error(...) }`. The modal shows no success confirmation and no error either way — the user cannot distinguish "saved" from "silently dropped" without devtools.

Concrete failure path: user pastes an expired Sentry token → clicks **Test** → spinner → nothing → assumes "no news is good news" → clicks Save → opens an alert's Sentry tab later and gets an unexplained failure, with no breadcrumb pointing back at the token.

## Impact

Anyone configuring the Sentry integration or saving dashboard settings. The settings modal is exactly the place where users are performing deliberate, verify-me actions; every outcome signal currently goes to the developer console, which SRE end-users will not have open. The Test button is functionally dead UI.

## Proposed approach

Add a small result surface to the Sentry section and the modal footer — the codebase already has the pattern: `annotationButtonsError` + inline display used by `saveAnnotationButtonConfigs()` (`dashboard_settings.templ:1084-1095`) and the toast in `StatisticsDashboard.templ:4585`. Concretely: a `sentryTestResult` field rendered as green "Connection successful" / red "Connection failed: …" text under the token input, an error message for the empty `saveSentryToken()` else-branch, and a success/error line (or toast) for `saveSettings()`. Edit `.templ` sources, regenerate with `make webui-templates`.

## Acceptance criteria

- [ ] Clicking Test with a valid token shows a visible success indicator in the modal; with an invalid token, a visible failure message.
- [ ] A rejected token save shows a visible error and does not clear the input; only success clears it.
- [ ] "Save All Settings" shows a visible confirmation on success and a visible error naming the failed step on failure.
- [ ] No feedback path relies on the browser console.
- [ ] No edits to `*_templ.go`; regenerated via `make webui-templates`.

SoulKyu:
## Problem

Every outcome signal in the dashboard Settings modal went to the browser console:

- **Test** button → `console.log('Connection Successful')` / `console.error('Connection Failed')`. A valid and an expired token looked pixel-identical once the spinner stopped.
- `saveSentryToken()` failure branch was empty. Success clears the token field, failure leaves it — so a rejected save reads as success to anyone not watching closely.
- `saveSettings()` ("💾 Save All Settings") funnelled every failure into `catch { console.error(...) }`, with no confirmation on success either.

## Changes

- `sentryConfig.testResult` = `{success, message}`, set by `testSentryConnection()` and `saveSentryToken()`. The green/red markup under the token input already existed in the Sentry tab — it was never populated. Backend messages (`Connection failed: ...`, `Invalid Sentry token`) are shown verbatim.
- `settingsSaveResult` drives a new line in the modal footer: green `All settings saved`, red with the thrown error, which already names the failed step (`Failed to save color preferences`, `Failed to save notification preferences`, …).
- Both are reset when the modal is reopened, so a stale outcome can't be mistaken for a fresh one.
- Failed token save keeps the value in the input; only success clears it.

## Validation

- `make webui-templates` (no hand edits to `*_templ.go`)
- `go build ./...`, `go test ./internal/webui/...`
- `node --check` on the extracted Alpine script

Closes #97


internal/webui/templates/scripts/dashboard_settings.templ
```diff
@@ -477,10 +487,10 @@ templ DashboardSettings() {
 							throw new Error('Failed to save annotation button configs');
 						}
 
-						console.log('All settings saved successfully');
-						
+						this.settingsSaveResult = { success: true, message: 'All settings saved' };
+
 					} catch (error) {
-						console.error('Error saving settings:', error);
+						this.settingsSaveResult = { success: false, message: error.message || 'Failed to save settings' };
```
SoulKyu, line 504: **The failure line names the step but throws away the reason, which the sub-savers already have.**

The message rendered here comes from the `throw new Error('Failed to save color preferences')` guards on lines 469, 475, 481 and 487, so the footer can only ever say *which* step failed. The sub-savers already hold the backend's explanation and discard it:

- `saveColorPreferences()` line 188 reads `result.error` and sends it to `console.error` before `return false`.
- `saveAnnotationButtonConfigs()` line 1100 computes `errorMsg = result.error || 'Failed to save annotation button configs'`, stores it in `this.annotationButtonsError`, and still returns a bare `false`.
- `saveNotificationPreferences()` line 944 logs a fixed string and returns `false`, so a 401 from an expired session and a 500 from the backend are indistinguishable.

That matters because the actionable half of the signal stays in the console — exactly where this PR is trying to move things out of. "Failed to save color preferences" gives the user no way to tell whether retrying helps, whether they need to re-authenticate, or whether their input was rejected.

Suggested change, keeping the boolean contract the standalone callers rely on (`dashboard_utilities.templ` lines 207 and 216 branch on the return value): add a `lastSaveError: null` field to the component, set it in each sub-saver's failure branch (`this.lastSaveError = result.error || error.message`), reset it to `null` at the top of `saveSettings()` alongside `settingsSaveResult`, and build the thrown message from both:

```js
const fail = (step) => {
	throw new Error(this.lastSaveError ? step + ': ' + this.lastSaveError : step);
};
// ...
if (!colorResult) fail('Failed to save color preferences');
```

While you are in there: `saveAnnotationButtonConfigs()` raises a blocking `alert()` on lines 1103 and 1109. Now that a footer line exists, that step is the only one that reports failure twice — dropping the `alert()` calls in favour of the new footer would make the whole modal behave consistently.


internal/webui/templates/scripts/dashboard_settings.templ
```diff
@@ -62,6 +64,13 @@ templ DashboardSettings() {
 
 				async init() {
 					console.log('Settings modal initializing...');
+					// Drop stale feedback so a reopened modal never shows the previous outcome
+					this.$watch('showSettings', (open) => {
+						if (open) {
+							this.settingsSaveResult = null;
+							this.sentryConfig.testResult = null;
+						}
+					});
```
SoulKyu, line 75: **A modal dismissed mid-save still ends with zero signal, and this watcher is what erases the evidence.**

`Cancel` is guarded with `:disabled="settingsSaving"` (`modal_components.templ` line 651), but the other two dismiss affordances are not:

- the backdrop, `modal_components.templ` line 7: `@click="showSettings = false"`
- the ✕ button, `modal_components.templ` line 15: `@click="showSettings = false"` with no `:disabled`

So this sequence is reachable: click 💾 Save All Settings → click ✕ (or the backdrop) while the spinner is running → one of the awaited saves rejects → `saveSettings()` sets `settingsSaveResult` to the failure object at line 493, but `showSettings` is already `false`, so nothing renders → the next time the modal opens, this watcher nulls it out before it is ever seen.

The net result is a settings save that failed server-side and reported nothing at all, which is the precise failure mode the PR set out to remove — and the user is left believing the change persisted.

Suggested change: mirror the `Cancel` guard on the two remaining exits, so the outcome cannot be dismissed before it is shown. In `modal_components.templ`, line 7 becomes `@click="if (!settingsSaving) showSettings = false"` and the ✕ on line 15 gains `:disabled="settingsSaving"`. If you would rather keep the modal dismissible at all times, the alternative is to make this watcher preserve an outcome that arrived while the modal was closed:

```js
this.$watch('showSettings', (open) => {
	if (open && !this.settingsSaving) {
		this.settingsSaveResult = null;
		this.sentryConfig.testResult = null;
	}
});
```

Blocking dismissal is the cleaner of the two, since it also stops the user from walking away from a half-applied multi-step save.


internal/webui/templates/components/modal_components.templ
```diff
@@ -653,6 +653,10 @@ templ SettingsModal() {
 							:class="{ 'opacity-50 cursor-not-allowed': settingsSaving }">
 						Cancel
 					</button>
+					<p x-show="settingsSaveResult"
+					   class="mt-3 text-sm sm:mt-0 sm:mr-auto sm:self-center"
+					   :class="settingsSaveResult && settingsSaveResult.success ? 'text-green-600 dark:text-green-400' : 'text-red-600 dark:text-red-400'"
+					   x-text="settingsSaveResult ? settingsSaveResult.message : ''"></p>
```
SoulKyu, line 663: **The new outcome line is not announced to assistive tech, so for screen-reader users the result is still as invisible as `console.log` was.**

This `<p>` and the Sentry result block on lines 507-510 both change their text content asynchronously, well after the user has clicked, with no `role` or `aria-live`. A screen reader gets no notification that "All settings saved" or the red failure text appeared — the user has to guess when the operation finished and go hunting for the element. Since the entire point of the change is to turn a console-only signal into a perceivable one, leaving the non-visual half of the audience on the old behaviour is a gap worth closing, and it is a two-attribute fix.

This works cleanly here precisely because the modal uses `x-show` rather than `x-if`: the element stays in the DOM, so the live region is registered before the text mutates and the announcement fires reliably.

Suggested change on this element:

```html
<p x-show="settingsSaveResult"
   role="status"
   aria-live="polite"
   class="mt-3 text-sm sm:mt-0 sm:mr-auto sm:self-center"
   :class="settingsSaveResult && settingsSaveResult.success ? 'text-green-600 dark:text-green-400' : 'text-red-600 dark:text-red-400'"
   x-text="settingsSaveResult ? settingsSaveResult.message : ''"></p>
```

And the same `role="status" aria-live="polite"` pair on the wrapper `<div x-show="sentryConfig.testResult" class="mt-2">` at line 507, so the Test / Save token outcomes announce too.


SoulKyu (commented):
Nice, focused fix — the Sentry result slot in `modal_components.templ` finally gets populated, the token now survives a rejected save, and the reopen watcher keeps stale outcomes from leaking into a fresh session. Generated files are in sync (`templ generate` produces no drift) and `go build ./...` is clean.

Three non-blocking follow-ups left inline, all about how much of the outcome actually reaches the user: the settings-save failure message drops the backend reason it already has, a modal dismissed mid-save still swallows its outcome entirely, and the two result lines are not exposed as live regions. None of these should hold up merge.




internal/webui/templates/scripts/dashboard_settings.templ
SoulKyu: @SoulKyu This has been fixed in cb280a8.

Added a lastSaveError field to settingsModalData and set it in the failure branches of saveColorPreferences, saveHiddenRules, saveNotificationPreferences (including an HTTP status fallback) and saveAnnotationButtonConfigs; saveSettings resets it and throws via a fail(step) helper that appends the backend reason. Removed both alert() calls in saveAnnotationButtonConfigs so the footer is the single failure channel. Boolean returns unchanged for the dashboard_utilities callers.




internal/webui/templates/scripts/dashboard_settings.templ
SoulKyu: @SoulKyu This has been fixed in cb280a8.

Took the blocking-dismissal option: modal_components.templ line 7 backdrop is now ="if (!settingsSaving) showSettings = false" and the close button gained :disabled="settingsSaving" plus a disabled style, matching the existing Cancel guard. No Escape-key handler closes this modal, so all three exits are now guarded and the watcher can no longer erase an unseen outcome.




internal/webui/templates/components/modal_components.templ
SoulKyu: @SoulKyu This has been fixed in cb280a8.

Added role="status" aria-live="polite" to the settingsSaveResult paragraph in the modal footer and to the sentryConfig.testResult wrapper div, so both asynchronous outcomes are announced to screen readers.




SoulKyu:
**Looper fixer round complete** — cb280a8

- ✅ Review comment on `internal/webui/templates/scripts/dashboard_settings.templ:504` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/99#discussion_r3650543948)
  - Added a lastSaveError field to settingsModalData and set it in the failure branches of saveColorPreferences, saveHiddenRules, saveNotificationPreferences (including an HTTP status fallback) and saveAnnotationButtonConfigs; saveSettings resets it and throws via a fail(step) helper that appends the backend reason. Removed both alert() calls in saveAnnotationButtonConfigs so the footer is the single failure channel. Boolean returns unchanged for the dashboard_utilities callers.
- ✅ Review comment on `internal/webui/templates/scripts/dashboard_settings.templ:75` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/99#discussion_r3650543950)
  - Took the blocking-dismissal option: modal_components.templ line 7 backdrop is now ="if (!settingsSaving) showSettings = false" and the close button gained :disabled="settingsSaving" plus a disabled style, matching the existing Cancel guard. No Escape-key handler closes this modal, so all three exits are now guarded and the watcher can no longer erase an unseen outcome.
- ✅ Review comment on `internal/webui/templates/components/modal_components.templ:663` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/99#discussion_r3650543952)
  - Added role="status" aria-live="polite" to the settingsSaveResult paragraph in the modal footer and to the sentryConfig.testResult wrapper div, so both asynchronous outcomes are announced to screen readers.


SoulKyu:
## 🧪 QA Report

Exercised against the running `make test` stack (postgres + backend + webui on :8081), driven with a real Chromium via Playwright. Default compose ships `NOTIFICATOR_SENTRY_ENABLED=false`, which leaves `sentryService == nil` and makes every Sentry call a 503 — so for this run I enabled Sentry on the webui container and pointed `NOTIFICATOR_SENTRY_BASE_URL` at a stub that answers `GET /api/0/` with 200 for one known token and 401 otherwise. That gives genuinely distinguishable valid/invalid tokens. No repo files were modified.

|---|---|---|
| Test with a valid token shows a visible success indicator; invalid token shows a visible failure message | ✅ | Typed `bad_token_qa` → clicked **Test** → red text under the input: *"Invalid token or connection failed"*. Replaced with the stub's valid token → **Test** → green *"Connection successful"*. Both rendered in the modal, screenshotted. |
| A rejected token save shows a visible error and does not clear the input; only success clears it | ✅ | Saved `bad_token_qa` → backend 400 → red *"Failed to save token: Invalid Sentry token"*, input still held `bad_token_qa`. Saved the valid token → green *"Sentry token saved successfully"*, input value became `""`. |
| "Save All Settings" shows a visible confirmation on success and a visible error naming the failed step on failure | ✅ | Success: green *"All settings saved"* in the footer. Failure (intercepted `POST /api/v1/notifications/preferences` → 500): red *"Failed to save notification preferences: HTTP 500 Internal Server Error"*. Second step checked too — `POST /api/v1/dashboard/color-preferences` returning `{success:false,error:"colour palette rejected by backend"}` → red *"Failed to save color preferences: colour palette rejected by backend"*, confirming the `fail(step)` + `lastSaveError` carry from cb280a8 works on both the HTTP-status and the `result.error` branch. |
| No feedback path relies on the browser console | ✅ | All four outcome paths (test / token-save success / token-save failure / save-all) render text in the modal; both slots carry `role="status" aria-live="polite"`. The two `alert()` calls in `saveAnnotationButtonConfigs` are gone, so the footer is the single failure channel. |
| No edits to `*_templ.go`; regenerated via `make webui-templates` | ✅ | Installed `templ` at the go.mod version (v0.3.906), deleted every `*_templ.go` and re-ran `templ generate` (127 files): `git status` on `*_templ.go` came back empty — the committed generated files are byte-identical to a fresh generation. |

**Exploratory:** Reopening the modal clears both stale outcomes (footer and Sentry slot back to empty) ✅; the new dismissal guard holds — a backdrop click while `settingsSaving` is true leaves the modal open, and the same click closes it normally once idle ✅; dashboard renders fine after all modal flows ✅. Separately surfaced (pre-existing, **not** caused by this frontend-only PR): re-saving a Sentry token after deleting one returns HTTP 500 — backend logs show a duplicate-key error at `internal/backend/database/sentry_db.go:122`. Ironically this PR is what makes that failure visible to the user at all; worth its own issue.

**Verdict: PASS**

PR #104, opened 2026-07-25
fix(webui): keep acknowledgement state in SSE alert updates

### LINKED ISSUE #102: fix(webui): SSE pushes alerts with acknowledgement and comment state stripped, flipping acked alerts back to un-acked every cycle
## Problem

The alert built from a fresh Alertmanager poll carries no collaboration state, but it is both the object compared against the cache *and* the object pushed to browsers over SSE.

`convertToDashboardAlert` (`internal/webui/services/alert_cache.go:291`) populates only what Alertmanager knows. It never sets `IsAcknowledged`, `AcknowledgedBy`, `AcknowledgedAt` or `CommentCount`, so those fields are always zero on the fresh object.

`hasAlertChanged` (`internal/webui/services/alert_cache.go:378`) then compares exactly those fields:

```go
if existing.IsAcknowledged != new.IsAcknowledged { return true }
if existing.CommentCount != new.CommentCount { return true }
```

For any alert that is acknowledged or has at least one comment, `existing` holds the real value and `new` holds zero, so this returns `true` on **every** refresh cycle, forever.

In the merge branch (`alert_cache.go:220`) the *stripped* object is what gets queued for SSE:

```go
if ac.hasAlertChanged(existingAlert, dashAlert) {
    updatedAlertsForSSE = append(updatedAlertsForSSE, dashAlert)  // <- zeroed ack/comment fields
}
ac.updateExistingAlert(existingAlert, dashAlert)
```

`updateExistingAlert` (`alert_cache.go:360`) is careful — it deliberately does not copy the collaboration fields, so the server-side cache stays correct. Only the SSE payload is wrong. It ships at `alert_cache.go:280`.

The browser then replaces the row wholesale (`internal/webui/templates/scripts/dashboard_data.templ:368`):

```js
this.alerts[existing.index] = updatedAlert;
```

and the table renders straight off those fields — `alert.isAcknowledged` drives the "Acknowledged" badge and the Acknowledge button (`internal/webui/templates/components/table_components.templ:83` and `:89`), `alert.commentCount` drives the comment icon and count (`table_components.templ:180-184`).

### Failure path

1. Alice acknowledges `HighMemoryUsage`. The badge appears; the cache is correct.
2. The next refresh cycle (10s by default) runs. `hasAlertChanged` is true because the fresh Alertmanager alert has `IsAcknowledged == false`.
3. Every connected browser receives an SSE `update` whose payload for that alert has `isAcknowledged: false`, `acknowledgedBy: ""`, `commentCount: 0`.
4. Alpine overwrites the row. The badge disappears, the Acknowledge button reappears, the comment count drops to `-`. It stays wrong until a full dashboard fetch repaints from the (correct) cache — and then flips back on the next tick.

If the user has an acknowledged filter or a display mode selected, `alertMatchesFilters(updatedAlert)` (`dashboard_data.templ:501`, called at `:371`) is evaluated against the stripped object, so the alert can be dropped from the visible list entirely.

## Impact

On any dashboard left open — which is the normal way this tool is used — acknowledgement badges and comment counts flicker off every 10 seconds. Two engineers looking at the same board disagree about whether an alert is owned, and someone re-acknowledges an alert that Alice already picked up. The collaboration features are the product; this makes them look broken.

Secondary cost: because `updateExistingAlert` bumps `existing.UpdatedAt = time.Now()` on the same always-true condition, every acknowledged or commented alert is permanently "recently updated". The poll path (`internal/webui/handlers/dashboard_handlers.go:1187`, which selects on `UpdatedAt > lastUpdate`) therefore re-sends all of them on every request, and sorting by "Updated" is meaningless for exactly the alerts someone is working on.

## Proposed approach

The comparison and the payload should both use cache-resident state, not the raw Alertmanager object.

- Restrict `hasAlertChanged` to fields the Alertmanager poll actually populates (`Status.State`, `EndsAt`, `Annotations`/`Summary`). `IsAcknowledged` and `CommentCount` change through `MutateAlert` and the `loadBackendData` path, not through a poll, so comparing them against a fresh alert is meaningless by construction.
- Push a snapshot of the merged cache entry rather than `dashAlert`. `updateExistingAlert` already produces the correct merged state on `existingAlert`; queue a copy of that (the same shallow-copy approach `GetAllAlerts` uses at `alert_cache.go:568`) so the SSE payload carries real ack and comment state.
- If acknowledgement changes should also trigger an SSE push, they need to be emitted from where they actually happen (`MutateAlert` call sites and `loadAcknowledgmentsEfficiently`), not inferred from a poll diff.

## Acceptance criteria

- With a dashboard open and an alert acknowledged, the badge and comment count remain stable across at least three refresh cycles with no full page reload.
- An SSE `update` payload for an acknowledged alert carries `isAcknowledged: true` and the correct `acknowledgedBy` and `commentCount`.
- An alert that is acknowledged and otherwise unchanged does **not** appear in `updatedAlerts` on a refresh cycle where nothing about it changed in Alertmanager.
- `UpdatedAt` no longer advances every cycle for acknowledged or commented alerts; sorting by "Updated" reflects real changes.
- A test in `internal/webui/services/alert_cache_test.go` seeds an acknowledged alert, runs a refresh with an unchanged Alertmanager payload, and asserts the alert is absent from `UpdatedAlerts`.

SoulKyu:
## Problem

`convertToDashboardAlert` builds an alert from what Alertmanager knows, so `IsAcknowledged`, `AcknowledgedBy` and `CommentCount` are always zero on it. `hasAlertChanged` compared exactly those fields against the cache entry, so every acknowledged or commented alert was "changed" on every poll — and the object queued for SSE was that stripped alert. Open dashboards lost the acknowledged badge and comment count once per refresh cycle, and `UpdatedAt` advanced forever on precisely the alerts someone was working on.

## Changes

`internal/webui/services/alert_cache.go`

- `hasPolledStateChanged` compares only what a poll populates: `Status.State`, `Summary`, `Annotations`. `EndsAt` is deliberately excluded — Alertmanager keeps pushing it forward for firing alerts, which would mark everything changed every cycle. `hasAlertChanged` is unchanged and still used by `UpdateAlert`, which does write collaboration fields.
- The refresh merge branch queues a shallow copy of the merged cache entry (same approach as `GetAllAlerts`) instead of `dashAlert`, so the SSE payload carries real ack and comment state.
- `updateExistingAlert` now copies `Summary` too. It never did; leaving it stale would make the new comparison report a change forever once a summary changed.
- Acknowledgements no longer reach browsers through the poll diff, so they are emitted where they happen: `MutateAlert` stamps `UpdatedAt` and pushes the merged alert to subscribers, and `loadAcknowledgmentsEfficiently` / `loadCommentCountsEfficiently` push only the alerts whose backend state actually moved rather than rewriting every alert every cycle.

No template or JS changes — the browser side was already rendering whatever the payload said.

## Validation

`TestAlertCache_RefreshKeepsAcknowledgementState` seeds an acknowledged alert, asserts the acknowledgement itself is pushed over SSE with `isAcknowledged: true`, then runs three refresh cycles with an unchanged Alertmanager payload and asserts no SSE update fires, the collaboration state survives and `UpdatedAt` does not move. A final cycle with a changed annotation asserts the push happens and carries both the new summary and the ack state.

`go build ./...`, `go vet ./internal/webui/...`, `go test ./... -count=1` and `go test ./internal/webui/... -race` all pass.

Closes #102

### DIFF
--- a/internal/webui/services/alert_cache.go
+++ b/internal/webui/services/alert_cache.go
@@ -40,6 +40,16 @@ type alertFetcher interface {
 	FetchAllAlertsDetailed() ([]alertmanager.AlertWithSource, map[string]error)
 }
 
+// collabLoader abstracts the backend calls that feed collaboration state into the
+// cache, so tests can exercise the dedup guards in loadAcknowledgmentsEfficiently
+// and loadCommentCountsEfficiently without a live backend.
+// *client.BackendClient satisfies it.
+type collabLoader interface {
+	IsConnected() bool
+	GetAllAcknowledgedAlerts(alertKeys []string) (map[string]*alertpb.Acknowledgment, error)
+	GetCommentCountsBatch(fingerprints []string) (map[string]int, error)
+}
+
 // maxBackendWorkers bounds concurrent backend calls spawned by a refresh cycle,
 // so a large diff cannot stampede the backend with thousands of gRPC calls.
 const maxBackendWorkers = 8
@@ -56,6 +66,7 @@ type AlertCache struct {
 	userHiddenAlerts   map[string]map[string]bool             // userID -> fingerprint -> hidden
 	alertmanagerClient alertFetcher
 	backendClient      *client.BackendClient
+	collabClient       collabLoader // same object as backendClient, narrowed so tests can fake it
 	colorService       *ColorService
 	backendSem         chan struct{} // semaphore bounding refresh-triggered backend calls
 
@@ -108,11 +119,19 @@ func NewAlertCache(amClient *alertmanager.MultiClient, backendClient *client.Bac
 		fetcher = amClient
 	}
 
+	// Same care for the collaboration seam: a typed-nil in the interface would
+	// defeat the nil check in loadBackendData.
+	var collab collabLoader
+	if backendClient != nil {
+		collab = backendClient
+	}
+
 	return &AlertCache{
 		alerts:                make(map[string]*webuimodels.DashboardAlert),
 		userHiddenAlerts:      make(map[string]map[string]bool),
 		alertmanagerClient:    fetcher,
 		backendClient:         backendClient,
+		collabClient:          collab,
 		colorService:          NewColorService(backendClient),
 		cachedColors:          make(map[string]map[string]*AlertColorResult),
 		backendSem:            make(chan struct{}, maxBackendWorkers),
@@ -220,7 +239,11 @@ func (ac *AlertCache) refreshAlerts() {
 		if existingAlert, exists := ac.alerts[fingerprint]; !exists {
 			ac.alerts[fingerprint] = dashAlert
 			ac.newAlerts = append(ac.newAlerts, fingerprint)
-			newAlertsForSSE = append(newAlertsForSSE, dashAlert)
+			// Snapshot, never the cache-resident pointer: the backend loaders and
+			// MutateAlert write to it under ac.mu while the SSE handler serialises
+			// the payload holding no lock.
+			alertCopy := *dashAlert
+			newAlertsForSSE = append(newAlertsForSSE, &alertCopy)
 
 			// Capture alert fired event for statistics
 			ac.runBounded(func() {
@@ -293,12 +316,10 @@ func (ac *AlertCache) refreshAlerts() {
 
 	log.Printf("Alert cache refresh complete: %d active alerts, %d newly resolved", activeCount, resolvedCount)
 
-	ac.loadBackendData()
-
-	// Refresh color cache for all active users after alerts are updated
-	go ac.RefreshAllCachedColors()
-
-	// Notify SSE subscribers if there are any changes
+	// Emit the poll diff before anything else can write to the cache. These are
+	// snapshots taken under ac.mu, so a collaboration write landing right after the
+	// unlock above must not be allowed to push its fresher payload first and then be
+	// overwritten by this older one — the dashboard replaces alert objects wholesale.
 	if len(newAlertsForSSE) > 0 || len(updatedAlertsForSSE) > 0 || len(removedFingerprints) > 0 {
 		update := &webuimodels.DashboardIncrementalUpdate{
 			NewAlerts:      newAlertsForSSE,
@@ -308,6 +329,11 @@ func (ac *AlertCache) refreshAlerts() {
 		}
 		ac.notifySubscribers(update)
 	}
+
+	ac.loadBackendData()
+
+	// Refresh color cache for all active users after alerts are updated
+	go ac.RefreshAllCachedColors()
 }
 
 func (ac *AlertCache) convertToDashboardAlert(alert models.Alert, source string) *webuimodels.DashboardAlert {
@@ -496,11 +522,11 @@ func (ac *AlertCache) MutateAlert(fingerprint string, fn func(*webuimodels.Dashb
 
 func (ac *AlertCache) loadBackendData() {
 	log.Printf("loadBackendData called - checking backend connection...")
-	if ac.backendClient == nil {
+	if ac.collabClient == nil {
 		log.Printf("Backend client is nil - skipping acknowledgment loading")
 		return
 	}
-	if !ac.backendClient.IsConnected() {
+	if !ac.collabClient.IsConnected() {
 		log.Printf("Backend client not connected - skipping acknowledgment loading")
 		return
 	}
@@ -528,7 +554,7 @@ func (ac *AlertCache) loadAcknowledgmentsEfficiently() {
 	}
 
 	// Step 2: call gRPC with NO lock held
-	acknowledgedAlerts, err := ac.backendClient.GetAllAcknowledgedAlerts(fingerprints)
+	acknowledgedAlerts, err := ac.collabClient.GetAllAcknowledgedAlerts(fingerprints)
 	if err != nil {
 		log.Printf("Failed to load acknowledged alerts from backend: %v", err)
 		return
@@ -624,7 +650,7 @@ func (ac *AlertCache) loadCommentCountsEfficiently() {
 	}
 
 	// Step 2: call gRPC with NO lock held
-	counts, err := ac.backendClient.GetCommentCountsBatch(fingerprints)
+	counts, err := ac.collabClient.GetCommentCountsBatch(fingerprints)
 	if err != nil {
 		log.Printf("Failed to load comment counts batch: %v", err)
 		// Reset all comment counts to 0 on error
--- a/internal/webui/services/alert_cache.go
+++ b/internal/webui/services/alert_cache.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"maps"
 	"strings"
 	"sync"
 	"time"
@@ -231,11 +232,19 @@ func (ac *AlertCache) refreshAlerts() {
 			})
 
 		} else {
-			// Check if alert changed before updating
-			if ac.hasAlertChanged(existingAlert, dashAlert) {
-				updatedAlertsForSSE = append(updatedAlertsForSSE, dashAlert)
+			// Only compare what a poll can actually tell us: the fresh alert has no
+			// collaboration state, so comparing it against the cache would report a
+			// change on every cycle for every acknowledged or commented alert.
+			changed := ac.hasPolledStateChanged(existingAlert, dashAlert)
+			ac.updateExistingAlert(existingAlert, dashAlert, changed)
+
+			if changed {
+				// Push the merged cache entry, not the stripped poll result, so the
+				// SSE payload keeps acknowledgement and comment state. Shallow copy
+				// for the same reason GetAllAlerts snapshots.
+				alertCopy := *existingAlert
+				updatedAlertsForSSE = append(updatedAlertsForSSE, &alertCopy)
 			}
-			ac.updateExistingAlert(existingAlert, dashAlert)
 		}
 	}
 
@@ -370,9 +379,8 @@ func (ac *AlertCache) convertToDashboardAlert(alert models.Alert, source string)
 // 		!old.EndsAt.Equal(new.EndsAt)
 // }
 
-func (ac *AlertCache) updateExistingAlert(existing, new *webuimodels.DashboardAlert) {
-	// Check if alert has meaningfully changed before updating UpdatedAt
-	if ac.hasAlertChanged(existing, new) {
+func (ac *AlertCache) updateExistingAlert(existing, new *webuimodels.DashboardAlert, changed bool) {
+	if changed {
 		existing.UpdatedAt = time.Now()
 	}
 	// Note: UpdatedAt is NOT updated if alert hasn't changed
@@ -383,6 +391,21 @@ func (ac *AlertCache) updateExistingAlert(existing, new *webuimodels.DashboardAl
 	existing.IsResolved = new.IsResolved
 
 	existing.Annotations = new.Annotations
+	// Summary is derived from the annotations, so it has to follow them; leaving
+	// it stale would make hasPolledStateChanged report a change forever.
+	existing.Summary = new.Summary
+}
+
+// hasPolledStateChanged compares only the fields an Alertmanager poll actually
+// populates. Acknowledgements and comment counts arrive through MutateAlert and
+// loadBackendData, never through a poll, so comparing them against a freshly
+// converted alert would always report a change.
+// ponytail: EndsAt is deliberately excluded — Alertmanager keeps pushing it
+// forward for firing alerts, which would mark every alert changed every cycle.
+func (ac *AlertCache) hasPolledStateChanged(existing, new *webuimodels.DashboardAlert) bool {
+	return existing.Status.State != new.Status.State ||
+		existing.Summary != new.Summary ||
+		!maps.Equal(existing.Annotations, new.Annotations)
 }
 
 // hasAlertChanged compares two alerts to determine if there are meaningful changes
@@ -448,15 +471,26 @@ func (ac *AlertCache) UpdateAlert(alert *webuimodels.DashboardAlert) {
 // fingerprint is not in the live cache. This is the only supported way to write
 // to a cached alert — the read accessors return snapshots, so writes through
 // their return values never reach the cache.
+//
+// A mutation is a real change by definition (acknowledgements, comments,
+// manual resolution), so it stamps UpdatedAt and pushes the merged alert to SSE
+// subscribers. The poll cycle cannot detect these — the alert it builds from
+// Alertmanager carries no collaboration state at all.
 func (ac *AlertCache) MutateAlert(fingerprint string, fn func(*webuimodels.DashboardAlert)) bool {
 	ac.mu.Lock()
-	defer ac.mu.Unlock()
 
 	alert, exists := ac.alerts[fingerprint]
 	if !exists {
+		ac.mu.Unlock()
 		return false
 	}
+	now := time.Now()
 	fn(alert)
+	alert.UpdatedAt = now
+	alertCopy := *alert
+	ac.mu.Unlock()
+
+	ac.notifyAlertsUpdated([]*webuimodels.DashboardAlert{&alertCopy}, now)
 	return true
 }
 
@@ -519,28 +553,57 @@ func (ac *AlertCache) loadAcknowledgmentsEfficiently() {
 // The snapshot is point-in-time: an acknowledgment written locally (the
 // acknowledge handler's MutateAlert) after the map was built is cleared here and
 // reappears on the next refresh.
+//
+// Only real transitions are pushed to SSE subscribers: a steady acknowledgment
+// re-observed on every refresh must not generate an update, or every poll cycle
+// would repaint the dashboard.
 func (ac *AlertCache) applyAcknowledgments(acknowledgedAlerts map[string]*alertpb.Acknowledgment) {
-	ac.mu.Lock()
-	defer ac.mu.Unlock()
+	now := time.Now()
+	var changed []*webuimodels.DashboardAlert
 
+	ac.mu.Lock()
 	for fingerprint, alert := range ac.alerts {
 		if acknowledgment, exists := acknowledgedAlerts[fingerprint]; exists {
+			acknowledgedAt := acknowledgment.CreatedAt.AsTime()
+			if alert.IsAcknowledged &&
+				alert.AcknowledgedBy == acknowledgment.Username &&
+				alert.AcknowledgedAt.Equal(acknowledgedAt) &&
+				alert.AcknowledgeReason == acknowledgment.Reason {
+				continue
+			}
+
 			alert.IsAcknowledged = true
 			alert.AcknowledgedBy = acknowledgment.Username
-			alert.AcknowledgedAt = acknowledgment.CreatedAt.AsTime()
+			alert.AcknowledgedAt = acknowledgedAt
 			alert.AcknowledgeReason = acknowledgment.Reason
+			alert.UpdatedAt = now
+
+			alertCopy := *alert
+			changed = append(changed, &alertCopy)
 
 			// Note: Comment counts are loaded separately by loadCommentCountsEfficiently()
 			// Note: We don't capture statistics here because this is loading historical
 			// acknowledgments. Statistics should only be captured when alerts are
 			// acknowledged in real-time to avoid negative MTTR calculations.
 		} else {
+			if !alert.IsAcknowledged && alert.AcknowledgedBy == "" &&
+				alert.AcknowledgedAt.IsZero() && alert.AcknowledgeReason == "" {
+				continue
+			}
+
 			alert.IsAcknowledged = false
 			alert.AcknowledgedBy = ""
 			alert.AcknowledgedAt = time.Time{}
 			alert.AcknowledgeReason = ""
+			alert.UpdatedAt = now
+
+			alertCopy := *alert
+			changed = append(changed, &alertCopy)
 		}
 	}
+	ac.mu.Unlock()
+
+	ac.notifyAlertsUpdated(changed, now)
 }
 
 func (ac *AlertCache) loadCommentCountsEfficiently() {
@@ -575,19 +638,29 @@ func (ac *AlertCache) loadCommentCountsEfficiently() {
 
 	// Step 3: write results back under Lock
 	alertsWithComments := 0
+	now := time.Now()
+	var changed []*webuimodels.DashboardAlert
+
 	ac.mu.Lock()
 	for fingerprint, alert := range ac.alerts {
-		if count, exists := counts[fingerprint]; exists {
-			alert.CommentCount = count
-			if count > 0 {
-				alertsWithComments++
-			}
-		} else {
-			alert.CommentCount = 0
+		count := counts[fingerprint] // missing fingerprint means no comments
+		if count > 0 {
+			alertsWithComments++
 		}
+		if alert.CommentCount == count {
+			continue
+		}
+
+		alert.CommentCount = count
+		alert.UpdatedAt = now
+
+		alertCopy := *alert
+		changed = append(changed, &alertCopy)
 	}
 	totalAlerts := len(ac.alerts)
 	ac.mu.Unlock()
+
+	ac.notifyAlertsUpdated(changed, now)
 
 	log.Printf("Successfully loaded comment counts for %d alerts (%d with comments) using batch query", totalAlerts, alertsWithComments)
 }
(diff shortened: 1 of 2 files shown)


internal/webui/services/alert_cache.go
```diff
@@ -230,11 +231,19 @@ func (ac *AlertCache) refreshAlerts() {
...
+			changed := ac.hasPolledStateChanged(existingAlert, dashAlert)
+			ac.updateExistingAlert(existingAlert, dashAlert, changed)
+
+			if changed {
+				// Push the merged cache entry, not the stripped poll result, so the
+				// SSE payload keeps acknowledgement and comment state. Shallow copy
+				// for the same reason GetAllAlerts snapshots.
+				alertCopy := *existingAlert
+				updatedAlertsForSSE = append(updatedAlertsForSSE, &alertCopy)
 			}
```
SoulKyu, line 270: **Stale snapshot can still overwrite a fresh acknowledgement — reorder the notify.**

The copy is taken correctly here (after `updateExistingAlert`, so it carries merged state), but it is not pushed until line 308. Between the two, `refreshAlerts` releases `ac.mu` (line 291), calls `ac.loadBackendData()` (line 295, which spawns `loadAcknowledgmentsEfficiently` and `loadCommentCountsEfficiently`), and starts `go ac.RefreshAllCachedColors()` (line 298). Any collaboration write that lands in that window mutates the cache entry and emits its own SSE update first; the refresh then emits this older copy second.

Concrete sequence: alert A's summary annotation changes in cycle N, so a copy with `IsAcknowledged: false` is taken here. A user clicks Acknowledge while `refreshAlerts` still holds `ac.mu`. `MutateAlert` blocks, acquires the lock right after line 291, writes the ack and pushes `{A, isAcknowledged: true}`. Line 308 then pushes this copy with `isAcknowledged: false`. Because the dashboard replaces the alert object wholesale, the badge disappears again.

What makes it worth fixing rather than tolerating: it does not self-heal. The next cycle's `hasPolledStateChanged` sees no polled change, so no push. And `loadAcknowledgmentsEfficiently`'s new guard (line 545) finds the cache already matching the backend, so no push there either. The alert stays visually un-acked until a real Alertmanager-side change or a page reload — the same user-visible symptom this PR set out to remove, just through a narrower door.

Suggested change: move the notify block (lines 300-309) up to immediately after `ac.mu.Unlock()` on line 291, ahead of `ac.loadBackendData()` and `go ac.RefreshAllCachedColors()`. The poll diff is then always emitted before any collaboration push that could supersede it, and the window shrinks to nothing. Keep it outside the lock as it is today, since `notifySubscribers` takes `subMutex`.


internal/webui/services/alert_cache.go
```diff
@@ -1101,6 +1162,18 @@ func (ac *AlertCache) notifySubscribers(update *webuimodels.DashboardIncremental
 	}
 }
 
+// notifyAlertsUpdated pushes already-snapshotted alerts to SSE subscribers.
+// Callers must pass copies, never cache-resident pointers.
```
SoulKyu, line 1227: **The documented copy invariant does not hold for the new-alert path in the same file.**

This helper states "Callers must pass copies, never cache-resident pointers", and all three new call sites (`MutateAlert`, `loadAcknowledgmentsEfficiently`, `loadCommentCountsEfficiently`) honour it. The sibling path in `refreshAlerts` does not: line 222 appends `dashAlert` — the exact pointer stored into `ac.alerts[fingerprint]` on line 220 — into `newAlertsForSSE`, and line 308 hands that slice to `notifySubscribers` directly rather than through this helper.

Why it matters more after this PR: the goroutines spawned by `loadBackendData` (line 295) and any concurrent `MutateAlert` from an HTTP handler now write `IsAcknowledged`, `AcknowledgedBy`, `CommentCount` and `UpdatedAt` into those same structs under `ac.mu`, while the SSE handler serialises them holding no lock at all. That is an unsynchronised read/write on the same fields. It will not surface under `-race` in the current suite because the tests never wire a real `backendClient`, so the loader goroutines never run. The pre-existing code had the same aliasing, but this change adds writers and now names the rule, so it is a good moment to close it.

Suggested change at line 222, mirroring what the existing-alert branch does at line 244:

```go
ac.alerts[fingerprint] = dashAlert
ac.newAlerts = append(ac.newAlerts, fingerprint)
alertCopy := *dashAlert
newAlertsForSSE = append(newAlertsForSSE, &alertCopy)
```

Cost is one struct copy per newly firing alert per cycle, and it makes the invariant written here true for every notify path in the file.


internal/webui/services/alert_cache.go
```diff
@@ -501,13 +535,28 @@ func (ac *AlertCache) loadAcknowledgmentsEfficiently() {
...
 	ac.mu.Lock()
 	for fingerprint, acknowledgment := range acknowledgedAlerts {
 		if alert, exists := ac.alerts[fingerprint]; exists {
+			acknowledgedAt := acknowledgment.CreatedAt.AsTime()
+			if alert.IsAcknowledged &&
+				alert.AcknowledgedBy == acknowledgment.Username &&
+				alert.AcknowledgedAt.Equal(acknowledgedAt) &&
+				alert.AcknowledgeReason == acknowledgment.Reason {
+				continue
+			}
```
SoulKyu, line 599: **The two new dedup guards are now load-bearing but have no test coverage, and cannot get any as written.**

Since `hasPolledStateChanged` deliberately stops reporting acknowledgement and comment-count differences, this guard and its twin in `loadCommentCountsEfficiently` (line 615) became the only mechanism that gets backend-side collaboration changes to open dashboards. `TestAlertCache_RefreshKeepsAcknowledgementState` is a good test, but it exercises the poll path and `MutateAlert` only — neither loader ever runs in it, because `ac.backendClient` is a concrete `*client.BackendClient` (alert_cache.go:57) and there is no seam to fake `GetAllAcknowledgedAlerts` or `GetCommentCountsBatch`.

Why that is risky: this guard is a four-field comparison including `AcknowledgedAt.Equal(acknowledgedAt)` and `AcknowledgeReason`, and both failure directions are silent. Drop a field and every cycle re-pushes every acknowledged alert (an SSE update per alert per poll, forever). Widen it and a re-acknowledgement by a different user never reaches browsers. Today the suite is green either way.

Suggested change: introduce a narrow seam the way `alertFetcher` (alert_cache.go:38) already does for the Alertmanager side, for example

```go
type collabLoader interface {
	IsConnected() bool
	GetAllAcknowledgedAlerts(alertKeys []string) (map[string]*alertpb.Acknowledgment, error)
	GetCommentCountsBatch(fingerprints []string) (map[string]int, error)
}
```

held in its own field (`*client.BackendClient` already satisfies it), then add one table-driven test over a fake covering: identical acknowledgement returned on two consecutive loads pushes exactly once; a changed `Reason` pushes; a changed `CreatedAt` pushes; comment count 0 to 2 pushes with `commentCount: 2` and ack state intact; unchanged count pushes nothing; and the `GetCommentCountsBatch` error branch (lines 593-601) zeroes counts. That last branch is worth pinning too, since it mutates every cached alert without notifying anyone.


SoulKyu (commented):
@SoulKyu The diagnosis here is exactly right, and splitting the comparison into `hasPolledStateChanged` (what a poll can observe) versus `hasAlertChanged` (what a full write carries) is the correct shape for this. Verified locally on `bec0fee`: `go build ./...`, `go vet ./internal/webui/...`, and `go test ./internal/webui/services/... -count=1 -race` all pass, and the new test does fail against the old comparison.

Three non-blocking follow-ups, all inline: a residual ordering window in `refreshAlerts` that can still push a stale ack snapshot, the `newAlertsForSSE` path that does not yet honour the copy invariant the new `notifyAlertsUpdated` helper documents, and the fact that the two new dedup guards in the backend loaders are currently untestable. None of these block merge.



internal/webui/services/alert_cache.go
SoulKyu: @SoulKyu This has been fixed in a6ffb6f.

Moved the notifySubscribers block in refreshAlerts to immediately after ac.mu.Unlock(), ahead of ac.loadBackendData() and go ac.RefreshAllCachedColors(), so the poll diff is always emitted before any collaboration push that could supersede it. It stays outside the lock as before.




internal/webui/services/alert_cache.go
SoulKyu: @SoulKyu This has been fixed in a6ffb6f.

The new-alert branch in refreshAlerts now appends a shallow copy of dashAlert to newAlertsForSSE instead of the cache-resident pointer, matching the existing-alert branch, so the notifyAlertsUpdated copy invariant holds on every notify path in the file.




internal/webui/services/alert_cache.go
SoulKyu: @SoulKyu This has been fixed in a6ffb6f.

Added a collabLoader interface (IsConnected/GetAllAcknowledgedAlerts/GetCommentCountsBatch) in its own collabClient field mirroring alertFetcher, and two tests over a fake: a table-driven ack test (identical pushes once; changed reason, CreatedAt and user each push) plus a comment-count test covering 0 to 2 with ack state intact, unchanged count silent, and the error branch zeroing counts without notifying.




SoulKyu:
**Looper fixer round complete** — a6ffb6f

- ✅ Review comment on `internal/webui/services/alert_cache.go:270` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/104#discussion_r3650808307)
  - Moved the notifySubscribers block in refreshAlerts to immediately after ac.mu.Unlock(), ahead of ac.loadBackendData() and go ac.RefreshAllCachedColors(), so the poll diff is always emitted before any collaboration push that could supersede it. It stays outside the lock as before.
- ✅ Review comment on `internal/webui/services/alert_cache.go:1193` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/104#discussion_r3650808310)
  - The new-alert branch in refreshAlerts now appends a shallow copy of dashAlert to newAlertsForSSE instead of the cache-resident pointer, matching the existing-alert branch, so the notifyAlertsUpdated copy invariant holds on every notify path in the file.
- ✅ Review comment on `internal/webui/services/alert_cache.go:577` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/104#discussion_r3650808312)
  - Added a collabLoader interface (IsConnected/GetAllAcknowledgedAlerts/GetCommentCountsBatch) in its own collabClient field mirroring alertFetcher, and two tests over a fake: a table-driven ack test (identical pushes once; changed reason, CreatedAt and user each push) plus a comment-count test covering 0 to 2 with ack state intact, unchanged count silent, and the error branch zeroing counts without notifying.


SoulKyu:

SoulKyu (commented):
@SoulKyu Thanks for the follow-up — this is a solid fix. I re-reviewed the full diff in `internal/webui/services/alert_cache.go` and `alert_cache_test.go` at the current PR head.

What I verified:
- `hasPolledStateChanged` correctly narrows the poll comparison to `Status.State`, `Summary`, and `Annotations`, so acknowledged/commented alerts no longer look "changed" on every refresh cycle just because the poll result carries no collaboration state.
- The refresh-cycle SSE push now snapshots (`alertCopy := *dashAlert` / `*existingAlert`) before queuing for `newAlertsForSSE`/`updatedAlertsForSSE`, and pushes the merged cache entry rather than the stripped poll result — so ack/comment state survives.
- `MutateAlert`, `loadAcknowledgmentsEfficiently`, and `loadCommentCountsEfficiently` now emit SSE updates directly where collaboration state actually changes, with dedup guards to avoid re-pushing unchanged ack/comment data.
- The reordering in `refreshAlerts` (poll-diff notify now happens before `loadBackendData`) addresses the ordering window flagged in the previous review round, and the new `collabLoader` interface plus its fake implementation finally makes the dedup guards testable.

I ran a full build, vet, and the alert cache test suite with the race detector locally — everything compiles cleanly and all tests pass, including the new tests covering acknowledgement state survival across refresh cycles and the dedup guards in the backend loaders.

I didn't find any blocking or non-blocking issues in the changed ranges — the previous round's follow-ups all look properly addressed. Nice work tightening this up.



SoulKyu:

SoulKyu (commented):
@SoulKyu Re-reviewed at the current head. This PR's actual content in `internal/webui/services/alert_cache.go` and `alert_cache_test.go` is byte-identical to the commit that already passed a full clean review round (the branch was rebased onto a newer `main`, which is why the head SHA changed without any code changes) — I diffed the two commits directly (`git diff <prev-reviewed-sha>..<current-head> -- internal/webui/services/`) and confirmed zero differences in these files.

I still did an independent full pass rather than just trusting that:
- `hasPolledStateChanged` correctly narrows the poll comparison to `Status.State`, `Summary`, and `Annotations` (via `maps.Equal`), so acknowledged/commented alerts stop looking "changed" on every refresh just because a poll result carries no collaboration state. `EndsAt` is still copied into the cache on every cycle in `updateExistingAlert`, it's just excluded from the comparison, which matches the stated rationale.
- The refresh cycle now snapshots (`alertCopy := *dashAlert` / `*existingAlert`) before queuing into `newAlertsForSSE`/`updatedAlertsForSSE`, so the SSE payload for updated alerts carries the merged cache state (with ack/comment data) instead of the stripped poll result.
- The poll-diff notify was moved before `loadBackendData()`, closing the ordering window from the previous round where a stale pre-backend-load snapshot could overwrite a fresher ack push in the browser.
- `MutateAlert` now stamps `UpdatedAt`, copies the alert, releases the lock, then notifies — avoiding notifying subscribers while holding `ac.mu`.
- `loadAcknowledgmentsEfficiently` / `loadCommentCountsEfficiently` now dedup against the existing cached state before pushing, and the `collabLoader` seam makes both dedup guards testable via `fakeCollabLoader`.

I ran `go build ./...`, `go vet ./internal/webui/...`, and `go test ./internal/webui/services/... -race -count=1` locally against `64096b0` — everything compiles and all tests pass, including the new tests covering acknowledgement-state survival across refresh cycles and the dedup guards.

No blocking or non-blocking issues in the changed ranges. Nice work carrying this through multiple review rounds to a clean state.



SoulKyu:

SoulKyu (commented):
@SoulKyu Re-reviewed at the current head (`d2f6007`). I diffed the changed files against the previously clean-reviewed head (`64096b0`) and confirmed the content of `internal/webui/services/alert_cache.go` and `alert_cache_test.go` is byte-identical — this is another rebase-only head move, no code delta.

I still did an independent full pass rather than relying on that diff alone:
- `hasPolledStateChanged` correctly narrows the poll comparison to `Status.State`, `Summary`, and `Annotations` (via `maps.Equal`), so an acknowledged/commented alert no longer looks "changed" on every refresh just because the fresh poll result carries no collaboration state. `EndsAt` stays copied into the cache every cycle but is excluded from the comparison, matching the documented rationale (Alertmanager keeps advancing it for firing alerts).
- The refresh cycle now snapshots (`alertCopy := *dashAlert` / `*existingAlert`) before queuing into `newAlertsForSSE`/`updatedAlertsForSSE`, so the SSE payload for both new and updated alerts carries the merged cache state instead of the collaboration-free poll result.
- The poll-diff notify was moved before `loadBackendData()`, closing the ordering window where a stale pre-backend-load snapshot could race a fresher ack push to the browser and win.
- `MutateAlert` releases `ac.mu` before calling `notifyAlertsUpdated`, and `loadAcknowledgmentsEfficiently`/`loadCommentCountsEfficiently` now dedup against the already-cached ack/comment state before pushing, backed by the new `collabLoader` seam and `fakeCollabLoader` in tests.

I ran `go build ./...`, `go vet ./internal/webui/...`, and `go test ./internal/webui/services/... -race -count=1` locally against `d2f6007` — everything compiles and all tests pass, including `TestAlertCache_RefreshKeepsAcknowledgementState` and the dedup-guard tests for the collaboration loaders.

No blocking or non-blocking issues in the changed ranges. Good, careful fix for a real user-visible regression — thanks for carrying it cleanly through the rebases.



SoulKyu:
🔀 Rebase (conflicts resolved): rebased onto `origin/main` (now includes #105, #113, #110).

**One conflict:** `internal/webui/services/alert_cache.go` — `loadAcknowledgmentsEfficiently`.
- `main` (#105) had extracted the ack-application into `applyAcknowledgments(map)` and inverted the loop to iterate `ac.alerts`, clearing ack fields for any alert the backend no longer reports (so a re-firing alert does not inherit the previous incident's ack).
- This PR iterated the backend map, added a no-op dedup guard, and pushed only changed alerts to SSE.

**Resolution:** kept `main`'s `applyAcknowledgments` signature and mirror-the-backend loop (both branches of the `exists` check preserved), and layered this PR's behaviour on top: the dedup guard short-circuits an unchanged ack, and the clear branch got a symmetric guard so already-clean alerts do not generate spurious updates. Both branches stamp `UpdatedAt`, snapshot the alert, and the collected set is pushed via `notifyAlertsUpdated` after the mutex is released. The "successfully updated" log stayed in the caller.

Second commit applied cleanly. `go build ./...`, `go vet ./...` pass; `go test ./internal/webui/...` green (99 tests), including `TestAlertCache_AcknowledgmentDoesNotSurviveResolution` from #105. No `*_templ.go` touched.

SoulKyu:
## 🧪 QA Report

Environment: `make test` (docker-compose: postgres + fake alertmanager + backend + webui), all 4 containers healthy, webui on `:8081`, 22 seeded alerts, 10s refresh interval. Criteria taken from #102.

|---|---|---|
| Badge and comment count stable across ≥3 refresh cycles with a dashboard open, no page reload | ✅ | Headless Chromium logged in, Acknowledged view, sampled every 11s for 5 cycles: the 4 acknowledged alerts stayed `ack=true by=qaagent cc=1`, no `Acknowledge Alert` button ever reappeared on them, ACK badge count constant at 6 |
| SSE `update` payload for an acknowledged alert carries `isAcknowledged: true` + correct `acknowledgedBy` / `commentCount` | ✅ | `curl -N /api/v1/dashboard/stream` while acking `HighMemoryUsage`: push at 10:34:36 with `isAcknowledged:true, commentCount:1`; the later real Alertmanager-side change (`active`→`resolved`, 10:35:25) still carried `isAcknowledged:true, acknowledgedBy:"qaagent", commentCount:1` |
| Acknowledged-but-unchanged alert absent from `updatedAlerts` on a quiet refresh cycle | ✅ | Two SSE captures totalling ~20 refresh cycles: the long-acked `HighCPUUsage` and `HighMemoryUsage` never appear in `updatedAlerts`, while unacked alerts (`CriticalCPUUsage`, …) churn every cycle. Each ack produces exactly two pushes — the ack itself, then one backend reconciliation (`userID`→`username`, DB timestamp, reason) — and then goes silent |
| `UpdatedAt` no longer advances every cycle for acknowledged alerts | ✅ | Polled `/api/v1/dashboard/data?displayMode=acknowledge` every 10s ×9: `updatedAt` frozen at `10:36:55` (HighCPUUsage) and `10:35:25` (HighMemoryUsage) throughout; HighMemoryUsage still read `10:35:25` twelve minutes later |
| Test in `alert_cache_test.go` asserting absence from `UpdatedAlerts` after an unchanged refresh | ✅ | `TestAlertCache_RefreshKeepsAcknowledgementState` seeds an acked alert, runs 3 refresh cycles and fails on any SSE update / any `UpdatedAt` move; `go test ./internal/webui/services/... -race -count=1` → 67 tests pass |

**Exploratory:** unacknowledge still pushes immediately (`isAcknowledged:false`, `commentCount:2`) and does not flip back over the following 4 cycles — the alert correctly leaves the Acknowledged view; free-text search still narrows the table (9 alerts → 5, all `HighCPUUsage`); the alert detail modal still opens. One pre-existing quirk unrelated to this diff: non-acked new alerts can slip into the Acknowledged view via SSE `newAlerts`, because `alertMatchesFilters` (`dashboard_data.templ:501`) has no `displayMode` check — same on `main`.

**Verdict: PASS**

SoulKyu (commented):
@SoulKyu Re-reviewed at the current PR head. This is another rebase-only head move: the PR-scoped diff (from `gh pr diff --patch`, the two commits this PR owns) is unchanged in content from the last clean-reviewed head. A raw two-commit-range diff also picks up unrelated main-branch fixes that landed separately (a re-firing-alert acknowledgement fix, an encryption-key requirement, and an admin-gated resolved-alert deletion restriction) which were pulled in by the rebase, not authored by this PR.

I still did an independent full pass on the PR's own diff rather than relying on that comparison alone:
- `hasPolledStateChanged` correctly narrows the poll comparison to `Status.State`, `Summary`, and `Annotations` (via `maps.Equal`), so an acknowledged/commented alert no longer looks "changed" on every refresh just because the fresh poll result carries no collaboration state. `EndsAt` is still copied into the cache every cycle in `updateExistingAlert` but stays out of the comparison, matching the documented rationale.
- The refresh cycle snapshots (`alertCopy := *dashAlert` / `*existingAlert`) before queuing into `newAlertsForSSE`/`updatedAlertsForSSE`, so the SSE payload for both new and updated alerts carries the merged cache state instead of the collaboration-free poll result.
- The poll-diff notify was moved before `loadBackendData()` in `refreshAlerts`, narrowing (though, as flagged in the first review round, not fully eliminating — a small window remains between the cache unlock and the subscriber notify) the ordering window where a stale pre-backend-load snapshot could race a fresher acknowledgement push to the browser.
- `MutateAlert` releases the cache lock before calling `notifyAlertsUpdated`, avoiding notifying subscribers while holding the lock.
- `loadAcknowledgmentsEfficiently`/`loadCommentCountsEfficiently` dedup against already-cached state before pushing, backed by the new `collabLoader` seam and its fake implementation in tests, with dedicated tests covering identical-push, changed-reason, changed-timestamp, different-user, and comment-count transitions.

I ran a full build, `go vet` on the webui package, and the alert cache test suite with the race detector locally against the current head — everything compiles and all tests pass, including the tests covering acknowledgement-state survival across refresh cycles and the dedup guards for the backend collaboration loaders.

No blocking or non-blocking issues in the changed ranges. Good, careful fix for a real user-visible regression — thanks for carrying it cleanly through the repeated rebases.



SoulKyu:
## 🧪 QA Report

Env: `make test` (full rebuild, docker-compose: postgres + fake alertmanager + backend + webui), webui on `:8081`, cache sync interval 10s. Browser flow driven with Playwright/Chromium (chrome-devtools MCP was not permitted in this run).

|---|---|---|
| Badge + comment count stay stable across ≥3 refresh cycles with no page reload | ✅ | Logged in as `qauser`, clicked the row's Acknowledge button, filled the reason modal, submitted, switched to the **Acknowledged** view. Sampled Alpine state + rendered DOM every 10s for 70s (7 cycles): `domAck=true`, `commentCount=1` at every sample, `updatedAt` frozen at `11:03:17`. No JS console errors. Screenshot shows the three ACK badges with their comment counts. |
| SSE `update` payload for an acknowledged alert carries `isAcknowledged: true` + correct `acknowledgedBy` / `commentCount` | ✅ | Raw `curl -N /api/v1/dashboard/stream` capture. Ack of `e809b5a5…` produced `{"isAcknowledged":true,"acknowledgedBy":"3ylfbtim…","commentCount":1}` immediately (MutateAlert push), then one follow-up `{"isAcknowledged":true,"acknowledgedBy":"qauser","commentCount":1}` when `applyAcknowledgments` normalised the username — a real transition, then silence. |
| An acknowledged, otherwise unchanged alert does **not** appear in `updatedAlerts` on a quiet cycle | ✅ | 65s SSE window (6+ refresh cycles) after the ack: 7 events captured (new alerts, firing→resolved transitions, heartbeats), **zero** carrying `e809b5a5…`. On `main` this alert would be in `updatedAlerts` every single cycle. |
| `UpdatedAt` no longer advances every cycle | ✅ | `GET /api/v1/dashboard/data` polled repeatedly over ~8 min: the three acknowledged alerts held `10:59:37.660`, `11:04:59.800`, `11:05:37.664` unchanged across every subsequent cycle. Each timestamp maps to a real event (ack normalisation, added comment, firing→resolved). |
| Test in `alert_cache_test.go` seeds an acked alert, refreshes with an unchanged payload, asserts absence from `UpdatedAlerts` | ✅ | `go test ./internal/webui/services/ -run TestAlertCache_RefreshKeepsAcknowledgementState -v` → PASS; full `TestAlertCache*` suite: 39 tests pass. |

**Extra collaboration check:** adding a comment on the already-acked alert (`POST /alert/:fp/comments`) pushed exactly one SSE update with `commentCount: 2` and `isAcknowledged: true` preserved, then went quiet — `loadCommentCountsEfficiently`'s new dedup guard behaves.

**Exploratory:** adjacent flows are intact — the poll diff still pushes real changes (firing→resolved, new alerts observed live over SSE), and the `silence` bulk action (another `MutateAlert` call site, now notifying outside the lock) pushed immediately and the alert reached `state=silenced` two cycles later. One pre-existing wart unrelated to this diff: the client-side `alertMatchesFilters` has no `acknowledge` branch (`dashboard_data.templ:586`), so un-acked alerts arriving via SSE `newAlerts` leak into the **Acknowledged** view (screenshot header reads "Showing 7 alerts" vs "1 to 3 of 3 results"). The JS is untouched by this PR and reproduces the same way on `main` — worth a separate issue, not a blocker here.

**Verdict: PASS**

PR #105, opened 2026-07-25
fix(backend): a re-firing alert no longer inherits the previous incident's acknowledgement

### LINKED ISSUE #101: fix(backend): a re-firing alert inherits the previous incident's acknowledgement and vanishes from the default dashboard
## Problem

An acknowledgment is stored keyed only by the alert fingerprint, and the fingerprint is a hash of the alert's labels alone — nothing that identifies *which firing* of that alert it belongs to:

- `internal/models/alert.go:140` — `GetFingerprint()` is `md5(sorted "k=v" label pairs)`. Two separate incidents of the same alert produce the same fingerprint.

Acknowledgment rows are never removed when the alert resolves:

- `internal/backend/database/gorm_db.go:470` — `DeleteAcknowledgment(alertKey, userID)` is the *only* code path that deletes from `acknowledgments`, and it only runs when a user explicitly un-acknowledges.
- `internal/backend/database/gorm_db.go:516` — `CreateResolvedAlert` snapshots the acks into the `resolved_alerts` row but leaves the live `acknowledgments` rows in place.
- `internal/backend/server.go:314` — the hourly retention job only calls `CleanupExpiredResolvedAlerts()`; it never touches `acknowledgments`.

And the WebUI cache re-applies whatever it finds, unconditionally:

- `internal/backend/database/gorm_db.go:484` — `GetAllAcknowledgedAlerts` returns the latest ack per `alert_key`, with no time bound and no link to the current firing.
- `internal/webui/services/alert_cache.go:492` — `loadAcknowledgmentsEfficiently` sets `IsAcknowledged = true`, `AcknowledgedBy`, `AcknowledgedAt` for every fingerprint the backend returns, and never clears them. (Contrast `loadCommentCountsEfficiently` at `alert_cache.go:544`, which does have an `else` branch resetting the value.)

### Failure path

1. `HighMemoryUsage{instance="db-1"}` fires. Alice acknowledges it — a row lands in `acknowledgments` with `alert_key = <fingerprint>`.
2. The alert resolves. `refreshAlerts` drops it from the cache (`alert_cache.go:261`) and snapshots it into `resolved_alerts`. The live ack row survives.
3. Days later the same alert fires again with identical labels, so `GetFingerprint()` returns the same hash. `refreshAlerts` inserts a fresh `DashboardAlert` with `IsAcknowledged = false`.
4. In the same refresh cycle, `loadBackendData()` → `loadAcknowledgmentsEfficiently()` finds Alice's old ack and flips the new alert to acknowledged, attributed to Alice, with her timestamp from days ago.
5. `getStandardAlerts()` (`internal/webui/handlers/dashboard_handlers.go:346`) filters out `alert.IsAcknowledged`. Classic is the default display mode (`dashboard_handlers.go:224`), so **the newly firing alert never appears on the dashboard anyone is looking at.**

## Impact

Every recurring alert — which is most alerts worth alerting on — is silently suppressed the second time it fires, for as long as nobody manually un-acknowledges it. The team sees an empty dashboard while the alert is firing, and the UI blames an engineer who did not acknowledge this incident and may not know it happened. This is the failure mode an alerting tool exists to prevent.

It also corrupts statistics: `AcknowledgedAt` predates `StartsAt` for the new firing, so any MTTA or fix-time derived from it is negative.

## Proposed approach

Scope an acknowledgment to a firing rather than to a label hash. Two workable directions:

- **Preferred:** clear the acknowledgment rows for a fingerprint when the WebUI stores the resolved alert. `CreateResolvedAlert` already receives the ack snapshot for history, so deleting the live rows in the same transaction loses nothing (`internal/backend/services/services.go:1072`, `internal/backend/database/gorm_db.go:516`).
- **Alternative:** carry the firing's `StartsAt` alongside the ack and have `GetAllAcknowledgedAlerts` only return acks created at or after the current firing started. Heavier — it changes the schema and the batch query signature.

Either way, `loadAcknowledgmentsEfficiently` should clear the ack fields for fingerprints the backend did *not* return, mirroring what `loadCommentCountsEfficiently` already does, so the cache converges instead of latching.

Existing deployments carry orphaned ack rows for alerts that already resolved, so a migration or one-off cleanup keyed on `resolved_alerts` is needed — otherwise the first re-fire after the fix still hits the old behaviour.

## Acceptance criteria

- After an acknowledged alert resolves and later fires again with identical labels, the new firing shows as **not** acknowledged and appears in the default classic dashboard.
- Acknowledgment history for the resolved incident is still visible on the resolved alert (the `resolved_alerts` snapshot is unchanged).
- Un-acknowledging still works, and a still-firing acknowledged alert stays acknowledged across refresh cycles.
- A test covers the resolve → re-fire cycle at the `AlertCache` level: seed an ack, resolve the alert, re-insert the same fingerprint, assert `IsAcknowledged == false`.
- Orphaned ack rows in existing deployments are addressed (migration or documented cleanup).

SoulKyu:
## Problem

`acknowledgments.alert_key` is the alert fingerprint — an md5 of the labels alone — and the rows were only ever deleted when a user explicitly un-acknowledged. So when an acknowledged alert resolved and later fired again with identical labels, `loadAcknowledgmentsEfficiently` found the old row and flipped the new firing to acknowledged, attributed to whoever handled the *previous* incident. `getStandardAlerts()` filters acknowledged alerts out, and classic is the default display mode, so the new firing never showed up on the dashboard. It also produced an `AcknowledgedAt` predating `StartsAt`, i.e. negative MTTA / fix-time.

## Changes

- **`database/gorm_db.go`** — `CreateResolvedAlert` now runs in a transaction that creates the `resolved_alerts` row *and* deletes the live `acknowledgments` rows for that fingerprint. The snapshot already carries the history, so nothing is lost and the next firing starts clean.
- **`webui/services/alert_cache.go`** — the ack-applying loop moves into `applyAcknowledgments`, which now iterates the cache and **clears** `IsAcknowledged` / `AcknowledgedBy` / `AcknowledgedAt` / `AcknowledgeReason` for fingerprints the backend did not return, mirroring `loadCommentCountsEfficiently`. The cache converges instead of latching.
- **`database/migrate.go`** — new `cleanupResolvedAcknowledgments` custom migration deletes acknowledgments whose `alert_key` appears in `resolved_alerts`, so existing deployments don't hit the old behaviour on the first re-fire. It is bounded by the resolved-alert retention (24h default); orphans older than that are indistinguishable from a live ack and must be un-acknowledged from the UI — documented in `openwiki/backend.md`.

## Tests

- `TestAlertCache_AcknowledgmentDoesNotSurviveResolution` — full resolve → re-fire cycle at the cache level: ack applied, alert stays acknowledged across refreshes while firing, resolves, fires again with the same fingerprint, comes back **not** acknowledged with all ack fields cleared.
- `TestCreateResolvedAlertClearsAcknowledgments` — live rows dropped, snapshot preserved, other alerts' acks untouched.
- `TestCleanupResolvedAcknowledgments` — orphan deleted, acknowledgment of a still-firing alert kept.

`go build ./...`, `go vet ./internal/...` and `go test ./internal/...` pass.

Closes #101

### DIFF
--- a/internal/backend/database/gorm_db.go
+++ b/internal/backend/database/gorm_db.go
@@ -2,11 +2,13 @@ package database
 
 import (
 	"crypto/rand"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"log"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"gorm.io/driver/postgres"
@@ -513,6 +515,42 @@ func (gdb *GormDB) GetAllAcknowledgedAlerts(alertKeys []string) (map[string]mode
 	return result, nil
 }
 
+// isEmptyAcknowledgmentSnapshot reports whether the caller supplied no usable
+// acknowledgment history — either nothing at all, because its fetch failed, or
+// an empty JSON payload.
+func isEmptyAcknowledgmentSnapshot(acknowledgments []byte) bool {
+	switch strings.TrimSpace(string(acknowledgments)) {
+	case "", "null", "[]", "{}":
+		return true
+	}
+	return false
+}
+
+// snapshotAcknowledgments reads the live acknowledgments of an alert through tx,
+// shaped like GetAcknowledgments so both snapshot sources serialize the same.
+// Returns nil when the alert has none.
+func snapshotAcknowledgments(tx *gorm.DB, fingerprint string) ([]byte, error) {
+	var acks []models.AcknowledgmentWithUser
+	if err := tx.Table("acknowledgments").
+		Select("acknowledgments.*, users.username").
+		Joins("JOIN users ON users.id = acknowledgments.user_id").
+		Where("acknowledgments.alert_key = ?", fingerprint).
+		Order("acknowledgments.created_at DESC").
+		Find(&acks).Error; err != nil {
+		return nil, fmt.Errorf("failed to read acknowledgments for resolved alert: %w", err)
+	}
+
+	if len(acks) == 0 {
+		return nil, nil
+	}
+
+	snapshot, err := json.Marshal(acks)
+	if err != nil {
+		return nil, fmt.Errorf("failed to serialize acknowledgments for resolved alert: %w", err)
+	}
+	return snapshot, nil
+}
+
 func (gdb *GormDB) CreateResolvedAlert(fingerprint, source string, alertData, comments, acknowledgments []byte, ttlHours int) (*models.ResolvedAlert, error) {
 	now := time.Now()
 	resolvedAlert := &models.ResolvedAlert{
@@ -525,11 +563,23 @@ func (gdb *GormDB) CreateResolvedAlert(fingerprint, source string, alertData, co
 		Source:          source,
 	}
 
-	// The acknowledgment belongs to the firing that just ended and is already
-	// snapshotted in the row above. Leaving the live rows behind would silently
-	// re-acknowledge the next firing of the same labels, which share a
-	// fingerprint, and hide it from the classic dashboard.
+	// The acknowledgment belongs to the firing that just ended. Leaving the live
+	// rows behind would silently re-acknowledge the next firing of the same
+	// labels, which share a fingerprint, and hide it from the classic dashboard.
+	//
+	// The caller's snapshot is best-effort — it comes from a separate RPC that
+	// may have failed — so the rows are read here when it is missing. Snapshot
+	// and delete then share one transaction and the history cannot be lost.
 	err := gdb.db.Transaction(func(tx *gorm.DB) error {
+		if isEmptyAcknowledgmentSnapshot(acknowledgments) {
+			snapshot, err := snapshotAcknowledgments(tx, fingerprint)
+			if err != nil {
+				return err
+			}
+			if snapshot != nil {
+				resolvedAlert.Acknowledgments = models.JSONB(snapshot)
+			}
+		}
 		if err := tx.Create(resolvedAlert).Error; err != nil {
 			return fmt.Errorf("failed to create resolved alert: %w", err)
 		}
--- a/internal/backend/database/gorm_db.go
+++ b/internal/backend/database/gorm_db.go
@@ -525,8 +525,21 @@ func (gdb *GormDB) CreateResolvedAlert(fingerprint, source string, alertData, co
 		Source:          source,
 	}
 
-	if err := gdb.db.Create(resolvedAlert).Error; err != nil {
-		return nil, fmt.Errorf("failed to create resolved alert: %w", err)
+	// The acknowledgment belongs to the firing that just ended and is already
+	// snapshotted in the row above. Leaving the live rows behind would silently
+	// re-acknowledge the next firing of the same labels, which share a
+	// fingerprint, and hide it from the classic dashboard.
+	err := gdb.db.Transaction(func(tx *gorm.DB) error {
+		if err := tx.Create(resolvedAlert).Error; err != nil {
+			return fmt.Errorf("failed to create resolved alert: %w", err)
+		}
+		if err := tx.Where("alert_key = ?", fingerprint).Delete(&models.Acknowledgment{}).Error; err != nil {
+			return fmt.Errorf("failed to clear acknowledgments for resolved alert: %w", err)
+		}
+		return nil
+	})
+	if err != nil {
+		return nil, err
 	}
 
 	return resolvedAlert, nil
--- a/internal/backend/database/migrate.go
+++ b/internal/backend/database/migrate.go
@@ -352,10 +352,15 @@ func (gdb *GormDB) backfillFixTimeSeconds() error {
 	return nil
 }
 
-// cleanupResolvedAcknowledgments drops acknowledgments whose alert already
-// resolved. Until CreateResolvedAlert started clearing them, these rows survived
-// the resolution and re-acknowledged the next firing of the same labels, which
-// share a fingerprint.
+// cleanupResolvedAcknowledgments drops acknowledgments left behind by a
+// resolution. Until CreateResolvedAlert started clearing them, these rows
+// survived the resolution and re-acknowledged the next firing of the same
+// labels, which share a fingerprint.
+//
+// This runs on every backend start, not once, so the predicate has to tell a
+// stale row from a live one: only an acknowledgment created *before* the
+// resolution belongs to the firing that ended. An alert that resolves, fires
+// again and gets acknowledged keeps that acknowledgment across restarts.
 //
 // ponytail: keyed on resolved_alerts, so it only reaches orphans whose resolved
 // snapshot has not expired yet (retention default 24h). Older orphans are
@@ -369,7 +374,11 @@ func (gdb *GormDB) cleanupResolvedAcknowledgments() error {
 
 	result := gdb.db.Exec(`
 		DELETE FROM acknowledgments
-		WHERE alert_key IN (SELECT fingerprint FROM resolved_alerts)
+		WHERE EXISTS (
+			SELECT 1 FROM resolved_alerts r
+			WHERE r.fingerprint = acknowledgments.alert_key
+			  AND r.resolved_at > acknowledgments.created_at
+		)
 	`)
 	if result.Error != nil {
 		return fmt.Errorf("failed to delete acknowledgments of resolved alerts: %w", result.Error)
--- a/internal/backend/database/migrate.go
+++ b/internal/backend/database/migrate.go
@@ -29,6 +29,11 @@ func (gdb *GormDB) RunCustomMigrations() error {
 		return fmt.Errorf("failed to backfill fix_time_seconds: %w", err)
 	}
 
+	// Drop acknowledgments left behind by alerts that already resolved
+	if err := gdb.cleanupResolvedAcknowledgments(); err != nil {
+		return fmt.Errorf("failed to cleanup resolved acknowledgments: %w", err)
+	}
+
 	log.Println("✅ Custom migrations completed")
 	return nil
 }
@@ -342,6 +347,38 @@ func (gdb *GormDB) backfillFixTimeSeconds() error {
 		log.Printf("✅ Backfilled fix_time_seconds for %d alert records", result.RowsAffected)
 	} else {
 		log.Println("ℹ️  No alerts need fix_time_seconds backfill")
+	}
+
+	return nil
+}
+
+// cleanupResolvedAcknowledgments drops acknowledgments whose alert already
+// resolved. Until CreateResolvedAlert started clearing them, these rows survived
+// the resolution and re-acknowledged the next firing of the same labels, which
+// share a fingerprint.
+//
+// ponytail: keyed on resolved_alerts, so it only reaches orphans whose resolved
+// snapshot has not expired yet (retention default 24h). Older orphans are
+// indistinguishable from a live acknowledgment and must be un-acknowledged from
+// the UI.
+func (gdb *GormDB) cleanupResolvedAcknowledgments() error {
+	if !gdb.db.Migrator().HasTable("acknowledgments") || !gdb.db.Migrator().HasTable("resolved_alerts") {
+		log.Println("ℹ️  acknowledgments or resolved_alerts table doesn't exist yet, skipping cleanup")
+		return nil
+	}
+
+	result := gdb.db.Exec(`
+		DELETE FROM acknowledgments
+		WHERE alert_key IN (SELECT fingerprint FROM resolved_alerts)
+	`)
+	if result.Error != nil {
+		return fmt.Errorf("failed to delete acknowledgments of resolved alerts: %w", result.Error)
+	}
+
+	if result.RowsAffected > 0 {
+		log.Printf("✅ Cleaned up %d acknowledgments left behind by resolved alerts", result.RowsAffected)
+	} else {
+		log.Println("ℹ️  No acknowledgments left behind by resolved alerts")
 	}
 
 	return nil
--- a/internal/backend/services/acknowledged_alerts_test.go
+++ b/internal/backend/services/acknowledged_alerts_test.go
@@ -19,7 +19,7 @@ func TestGetAllAcknowledgedAlerts_SurfacesDatabaseError(t *testing.T) {
 	if err != nil {
 		t.Fatalf("failed to create test database: %v", err)
 	}
-	svc := NewAlertServiceGorm(db)
+	svc := NewAlertServiceGorm(db, &config.AdminConfig{})
 
 	resp, err := svc.GetAllAcknowledgedAlerts(context.Background(), &alertpb.GetAllAcknowledgedAlertsRequest{
 		AlertKeys: []string{"fp-1"},
--- a/internal/backend/services/acknowledged_alerts_test.go
+++ b/internal/backend/services/acknowledged_alerts_test.go
@@ -1,0 +1,33 @@
+package services
+
+import (
+	"context"
+	"testing"
+
+	"notificator/config"
+	"notificator/internal/backend/database"
+	alertpb "notificator/internal/backend/proto/alert"
+)
+
+// TestGetAllAcknowledgedAlerts_SurfacesDatabaseError guards the webui cache:
+// applyAcknowledgments clears every fingerprint missing from the response, so an
+// empty map returned with a nil error would un-acknowledge the whole dashboard
+// on a statement timeout or a failover. The failure has to reach the caller.
+func TestGetAllAcknowledgedAlerts_SurfacesDatabaseError(t *testing.T) {
+	// No AutoMigrate: the acknowledgments table is missing, so the query fails.
+	db, err := database.NewGormDB("sqlite", config.DatabaseConfig{SQLitePath: t.TempDir() + "/test.db"})
+	if err != nil {
+		t.Fatalf("failed to create test database: %v", err)
+	}
+	svc := NewAlertServiceGorm(db)
+
+	resp, err := svc.GetAllAcknowledgedAlerts(context.Background(), &alertpb.GetAllAcknowledgedAlertsRequest{
+		AlertKeys: []string{"fp-1"},
+	})
+	if err == nil {
+		t.Fatalf("expected a database failure to surface as an error, got response %+v", resp)
+	}
+	if resp != nil {
+		t.Errorf("expected no response alongside the error, got %+v", resp)
+	}
+}
--- a/internal/webui/services/alert_cache.go
+++ b/internal/webui/services/alert_cache.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"notificator/internal/alertmanager"
+	alertpb "notificator/internal/backend/proto/alert"
 	"notificator/internal/models"
 	"notificator/internal/webui/client"
 	webuimodels "notificator/internal/webui/models"
@@ -501,9 +502,20 @@ func (ac *AlertCache) loadAcknowledgmentsEfficiently() {
 
 	log.Printf("Received %d acknowledged alerts from backend", len(acknowledgedAlerts))
 
+	ac.applyAcknowledgments(acknowledgedAlerts)
+
+	log.Printf("Successfully updated %d alerts with acknowledgment data", len(acknowledgedAlerts))
+}
+
+// applyAcknowledgments makes the cache mirror the backend exactly: alerts the
+// backend no longer reports as acknowledged are cleared, so a re-firing alert
+// does not latch onto the previous incident's acknowledgment.
+func (ac *AlertCache) applyAcknowledgments(acknowledgedAlerts map[string]*alertpb.Acknowledgment) {
 	ac.mu.Lock()
-	for fingerprint, acknowledgment := range acknowledgedAlerts {
-		if alert, exists := ac.alerts[fingerprint]; exists {
+	defer ac.mu.Unlock()
+
+	for fingerprint, alert := range ac.alerts {
+		if acknowledgment, exists := acknowledgedAlerts[fingerprint]; exists {
 			alert.IsAcknowledged = true
 			alert.AcknowledgedBy = acknowledgment.Username
 			alert.AcknowledgedAt = acknowledgment.CreatedAt.AsTime()
@@ -513,11 +525,13 @@ func (ac *AlertCache) loadAcknowledgmentsEfficiently() {
 			// Note: We don't capture statistics here because this is loading historical
 			// acknowledgments. Statistics should only be captured when alerts are
 			// acknowledged in real-time to avoid negative MTTR calculations.
+		} else {
+			alert.IsAcknowledged = false
+			alert.AcknowledgedBy = ""
+			alert.AcknowledgedAt = time.Time{}
+			alert.AcknowledgeReason = ""
 		}
 	}
-	ac.mu.Unlock()
-
-	log.Printf("Successfully updated %d alerts with acknowledgment data", len(acknowledgedAlerts))
 }
 
 func (ac *AlertCache) loadCommentCountsEfficiently() {
--- a/internal/webui/services/alert_cache.go
+++ b/internal/webui/services/alert_cache.go
@@ -510,6 +510,15 @@ func (ac *AlertCache) loadAcknowledgmentsEfficiently() {
 // applyAcknowledgments makes the cache mirror the backend exactly: alerts the
 // backend no longer reports as acknowledged are cleared, so a re-firing alert
 // does not latch onto the previous incident's acknowledgment.
+//
+// acknowledgedAlerts must be a snapshot the backend actually produced — never an
+// empty map standing in for a failed fetch, which would un-acknowledge the whole
+// dashboard. GetAllAcknowledgedAlerts returns a gRPC error on a database failure
+// for that reason, and callers must return before reaching here.
+//
+// The snapshot is point-in-time: an acknowledgment written locally (the
+// acknowledge handler's MutateAlert) after the map was built is cleared here and
+// reappears on the next refresh.
 func (ac *AlertCache) applyAcknowledgments(acknowledgedAlerts map[string]*alertpb.Acknowledgment) {
 	ac.mu.Lock()
 	defer ac.mu.Unlock()
--- a/internal/backend/database/gorm_db_test.go
+++ b/internal/backend/database/gorm_db_test.go
@@ -1,6 +1,7 @@
 package database
 
 import (
+	"encoding/json"
 	"testing"
 	"time"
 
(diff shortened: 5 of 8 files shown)


internal/backend/database/migrate.go
```diff
@@ -346,3 +351,35 @@ func (gdb *GormDB) backfillFixTimeSeconds() error {
...
+// the UI.
+func (gdb *GormDB) cleanupResolvedAcknowledgments() error {
+	if !gdb.db.Migrator().HasTable("acknowledgments") || !gdb.db.Migrator().HasTable("resolved_alerts") {
+		log.Println("ℹ️  acknowledgments or resolved_alerts table doesn't exist yet, skipping cleanup")
+		return nil
+	}
+
+	result := gdb.db.Exec(`
+		DELETE FROM acknowledgments
+		WHERE alert_key IN (SELECT fingerprint FROM resolved_alerts)
```
SoulKyu: **The cleanup migration runs on every backend start and deletes acknowledgments of the current firing.**

`cleanupResolvedAcknowledgments` is called from `RunCustomMigrations` (`internal/backend/database/migrate.go:33`), which `AutoMigrate` invokes unconditionally (`internal/backend/database/gorm_db.go:116`, reached from `internal/backend/server.go:53`). It is not a one-shot migration, so this `DELETE` re-executes on every process start, and the predicate only asks "does this fingerprint appear in `resolved_alerts`?" - it never compares timestamps.

That makes the following sequence lose data, which is the exact failure mode this PR sets out to fix, just triggered from the other side:

1. Alert X resolves. A `resolved_alerts` row is written with a 24h TTL.
2. Ten minutes later X fires again and an operator acknowledges the new firing.
3. The backend restarts (deploy, crash loop, pod reschedule) while the `resolved_alerts` row is still inside its retention window.
4. This statement deletes the operator's ack for the live firing, and the alert silently pops back into the dashboard as un-acknowledged.

I verified this against the PR head with a throwaway test in `internal/backend/database`: a `resolved_alerts` row resolved one hour ago plus an `acknowledgments` row created five minutes ago, then a call to `cleanupResolvedAcknowledgments()`. The live ack is deleted - `got 0, want 1`.

Because acknowledgments carry `created_at` and resolved alerts carry `resolved_at`, the predicate can distinguish a stale ack from a live one without any migration bookkeeping. Only acks that predate a resolution are orphans:

```sql
DELETE FROM acknowledgments
WHERE EXISTS (
    SELECT 1 FROM resolved_alerts r
    WHERE r.fingerprint = acknowledgments.alert_key
      AND r.resolved_at > acknowledgments.created_at
)
```

I ran this variant locally on the SQLite test DB and it passes the scenario above while still deleting genuine orphans; the correlated-subquery `DELETE ... WHERE EXISTS` form is valid on both SQLite and PostgreSQL, so it fits the `IsSQLite()` / `IsPostgreSQL()`-free style of this function.

Two follow-ups that belong with the change:

- `TestCleanupResolvedAcknowledgments` (`internal/backend/database/gorm_db_test.go:128`) currently builds its orphan with an implicit `CreatedAt` of `now` while `ResolvedAt` is `now-1h`, i.e. an ack that postdates the resolution. That is not what a pre-fix orphan looks like. Set the orphan's `CreatedAt` explicitly to before `ResolvedAt`, and add a third fixture - a `resolved_alerts` row plus an ack created *after* its `resolved_at` - asserting the ack survives. That third case is what pins the regression.
- The doc comment on that test calls it a "one-off migration", and `openwiki/backend.md` describes it as dropping "rows orphaned before that fix". Both read as one-time; please reword to say it runs on every startup and is bounded by the resolved-alert retention window.


internal/webui/services/alert_cache.go
```diff
@@ -501,9 +502,20 @@ func (ac *AlertCache) loadAcknowledgmentsEfficiently() {
 
 	log.Printf("Received %d acknowledged alerts from backend", len(acknowledgedAlerts))
 
+	ac.applyAcknowledgments(acknowledgedAlerts)
```
SoulKyu, line 505: **A backend DB error yields an empty map with a nil error, and this call then clears every acknowledgment in the cache.**

The rewrite makes `applyAcknowledgments` mirror the backend exactly: any fingerprint absent from `acknowledgedAlerts` has `IsAcknowledged` / `AcknowledgedBy` / `AcknowledgedAt` / `AcknowledgeReason` cleared. That is correct only if the map is a trustworthy snapshot of the backend, and on this path it is not.

`AlertServiceGorm.GetAllAcknowledgedAlerts` (`internal/backend/services/services.go:718-725`) swallows the database error:

- on `s.db.GetAllAcknowledgedAlerts(...)` failure it logs and returns `AcknowledgedAlerts: make(map[string]*alertpb.Acknowledgment), Count: 0` with a **nil** gRPC error;
- `BackendClient.GetAllAcknowledgedAlerts` (`internal/webui/client/backend_client.go:380-385`) only propagates a transport error, so the webui receives `(empty map, nil)`;
- the `err != nil` guard at line 498 above does not fire, and this line applies the empty map.

So a statement timeout, a pool exhaustion, or a Postgres failover - none of which break the gRPC connection - un-acknowledges the entire dashboard until the next successful refresh. Before this PR the loop was additive, so an empty map was a harmless no-op; the mirror semantics turn it into a destructive one. The user-visible effects are more than cosmetic: `getStandardAlerts()` filters on `IsAcknowledged`, so previously-hidden alerts reappear, and operators re-acknowledging them writes duplicate acks plus duplicate audit comments and re-captures acknowledgment statistics, skewing MTTA. This is why the `loadCommentCountsEfficiently` precedent cited in the PR description does not transfer: a comment count briefly reading 0 is cosmetic, an ack flipping off is not.

Suggested change: make the RPC surface the failure instead of masking it, so the existing early `return` at line 499 does the right thing. In `internal/backend/services/services.go`, replace the empty-map return with a real gRPC error, e.g. `return nil, status.Errorf(codes.Internal, "failed to load acknowledged alerts: %v", err)`. The webui client already returns that error unchanged, and `loadAcknowledgmentsEfficiently` already logs and returns without touching the cache. Worth a small test asserting the cache keeps its acknowledgments when `GetAllAcknowledgedAlerts` returns an error, alongside the new resolve/re-fire test.

Related, and much smaller: the acknowledge handler in `internal/webui/handlers/dashboard_handlers.go` writes to the backend and then calls `alertCache.MutateAlert(...)`. If a `loadAcknowledgmentsEfficiently` fetch is in flight and its map was built before the ack was persisted, this function now clears the freshly-set flag rather than merely not setting it. It self-heals on the next refresh, so it is not a blocker, but it is worth a sentence in the `applyAcknowledgments` doc comment so the next reader knows the map is assumed to be a point-in-time snapshot that may lose a very recent local write.


internal/backend/database/gorm_db.go
```diff
@@ -525,8 +525,21 @@ func (gdb *GormDB) CreateResolvedAlert(fingerprint, source string, alertData, co
...
+	// The acknowledgment belongs to the firing that just ended and is already
+	// snapshotted in the row above. Leaving the live rows behind would silently
+	// re-acknowledge the next firing of the same labels, which share a
+	// fingerprint, and hide it from the classic dashboard.
+	err := gdb.db.Transaction(func(tx *gorm.DB) error {
+		if err := tx.Create(resolvedAlert).Error; err != nil {
+			return fmt.Errorf("failed to create resolved alert: %w", err)
+		}
+		if err := tx.Where("alert_key = ?", fingerprint).Delete(&models.Acknowledgment{}).Error; err != nil {
+			return fmt.Errorf("failed to clear acknowledgments for resolved alert: %w", err)
```
SoulKyu, line 587: **The delete runs even when the caller failed to build the acknowledgment snapshot, which makes the history loss permanent.**

The transaction is right, and pairing the snapshot write with the delete is the correct atomicity boundary. The gap is that the snapshot is produced by the caller on a best-effort basis, while the delete is unconditional.

In `AlertCache.storeResolvedAlertInBackend` (`internal/webui/services/alert_cache.go`), the acknowledgments payload comes from `ac.backendClient.GetAcknowledgments(alert.Fingerprint)`; if that call errors, the code logs and leaves `acknowledgments` as `nil`, then calls `CreateResolvedAlert` anyway. Nothing upstream validates it: `AlertServiceGorm.CreateResolvedAlert` requires only `Fingerprint`, `Source` and `AlertData`, so an empty `Acknowledgments` field reaches this transaction and the rows are deleted with no snapshot behind them.

Before this PR that failure was harmless - the snapshot was empty but the `acknowledgments` rows stayed in the database. Now a single transient gRPC hiccup on the snapshot fetch permanently destroys who acknowledged the incident and why. There is no retry path either: the alert has already been removed from the live cache by `refreshAlerts` before `storeResolvedAlertInBackend` runs, so this is the only chance to preserve it.

Suggested change: take the snapshot server-side, inside the transaction that deletes, so it cannot be missing. In `CreateResolvedAlert`, before the delete, when the caller supplied no acknowledgments payload, read the rows through `tx` and marshal them into `resolvedAlert.Acknowledgments` yourself:

- select the `acknowledgments` rows for `fingerprint` (joined to `users` for the username, matching what `GetAcknowledgments` returns) using `tx`;
- if the caller passed a non-empty payload, keep it and skip the read;
- then perform the delete.

That removes the whole class - the rows can only disappear once the same transaction has recorded them. A cheaper alternative, if you prefer to keep the RPC dumb, is to have the webui skip the delete when its snapshot fetch failed (an explicit request flag), but the server-side read is the one that cannot be got wrong by a future caller.


SoulKyu (commented):
@SoulKyu The root-cause analysis here is spot on: `alert_key` being a label-only fingerprint really is why a re-firing alert inherited the previous incident's ack, and clearing the live rows inside the same transaction that writes the snapshot is the right shape for the fix. The cache-convergence rewrite of `applyAcknowledgments` and the resolve -> re-fire test are both good.

Two changed ranges reintroduce the same class of bug from a different direction, though, and I think they need to be addressed before merge:

1. `cleanupResolvedAcknowledgments` is wired into `RunCustomMigrations`, which runs on **every** backend start, not once. Any ack placed on a *new* firing is deleted on the next restart as long as the previous resolution is still inside the 24h retention window. I reproduced this locally against the PR head.
2. `applyAcknowledgments` now treats the fetched map as authoritative, but the `GetAllAcknowledgedAlerts` RPC returns an empty map with a nil error when its DB query fails, so a transient database problem silently un-acknowledges every alert in the cache.

There is also one non-blocking data-retention concern on the new delete in `CreateResolvedAlert`. Details inline. Nice work on the diagnosis - the remaining items are all about the blast radius of the deletes, not the direction of the fix.

_Posted as a COMMENT review because GitHub does not allow requesting changes on your own pull request; the two items above are meant as blocking._




internal/backend/database/migrate.go
SoulKyu: @SoulKyu This has been fixed in e550424.

cleanupResolvedAcknowledgments now deletes via WHERE EXISTS with r.resolved_at > acknowledgments.created_at (internal/backend/database/migrate.go:370), so an ack created after a resolution survives restarts. TestCleanupResolvedAcknowledgments gained an explicit pre-resolution CreatedAt on the orphan plus a third refired-key fixture asserting survival; verified it fails against the old predicate. Doc comment and openwiki/backend.md now say it runs on every startup, bounded by resolved-alert reten…




internal/webui/services/alert_cache.go
SoulKyu: @SoulKyu This has been fixed in e550424.

AlertServiceGorm.GetAllAcknowledgedAlerts now returns status.Errorf(codes.Internal, ...) instead of an empty map with nil error (internal/backend/services/services.go:728), so loadAcknowledgmentsEfficiently's existing early return keeps the cache intact on a DB failure. Added internal/backend/services/acknowledged_alerts_test.go asserting the error surfaces, and documented on applyAcknowledgments that the map must be a real backend snapshot and is point-in-time versus the handler's MutateAlert w…




internal/backend/database/gorm_db.go
SoulKyu: @SoulKyu This has been fixed in e550424.

CreateResolvedAlert now takes the snapshot server-side: when the caller supplies no acknowledgments payload (nil, null, [] or {}), snapshotAcknowledgments reads the rows through tx joined to users and marshals them into resolvedAlert.Acknowledgments before the delete (internal/backend/database/gorm_db.go:516-585). Added TestCreateResolvedAlertSnapshotsAcknowledgmentsWhenCallerHasNone covering all three empty payloads; verified it fails without the in-transaction read.




SoulKyu:
**Looper fixer round complete** — e550424

- ✅ Review comment on `internal/backend/database/migrate.go` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/105#discussion_r3650824059)
  - cleanupResolvedAcknowledgments now deletes via WHERE EXISTS with r.resolved_at > acknowledgments.created_at (internal/backend/database/migrate.go:370), so an ack created after a resolution survives restarts. TestCleanupResolvedAcknowledgments gained an explicit pre-resolution CreatedAt on the orphan plus a third refired-key fixture asserting survival; verified it fails against the old predicate. Doc comment and openwiki/backend.md now say it runs on every startup, bounded by resolved-alert reten…
- ✅ Review comment on `internal/webui/services/alert_cache.go:505` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/105#discussion_r3650824062)
  - AlertServiceGorm.GetAllAcknowledgedAlerts now returns status.Errorf(codes.Internal, ...) instead of an empty map with nil error (internal/backend/services/services.go:728), so loadAcknowledgmentsEfficiently's existing early return keeps the cache intact on a DB failure. Added internal/backend/services/acknowledged_alerts_test.go asserting the error surfaces, and documented on applyAcknowledgments that the map must be a real backend snapshot and is point-in-time versus the handler's MutateAlert w…
- ✅ Review comment on `internal/backend/database/gorm_db.go:587` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/105#discussion_r3650824065)
  - CreateResolvedAlert now takes the snapshot server-side: when the caller supplies no acknowledgments payload (nil, null, [] or {}), snapshotAcknowledgments reads the rows through tx joined to users and marshals them into resolvedAlert.Acknowledgments before the delete (internal/backend/database/gorm_db.go:516-585). Added TestCreateResolvedAlertSnapshotsAcknowledgmentsWhenCallerHasNone covering all three empty payloads; verified it fails without the in-transaction read.


SoulKyu:
## 🧪 QA Report

Env: `make docker-build-all` + `docker compose up` (postgres + fake alertmanager + backend + webui), webui sync interval 10s. Alert driven through the fake Alertmanager (`POST /api/v2/alerts`, labels `alertname=QAReFireTest, instance=qa-db-1, severity=critical, job=qa-exporter, team=qa` → fingerprint `e9f1a63a80ae1b49eef75fdbc71fde01`); resolution triggered by making the source stop reporting it. UI checked in a real Chromium (Playwright), logged in as `qaagent`.

|---|---|---|
| Re-firing alert with identical labels shows as **not** acknowledged and appears in the default classic dashboard | ✅ | Acked while firing → `acknowledgments` row present, alert gone from classic. Source stopped reporting it → resolved. Re-posted the same labels → same fingerprint `e9f1a63a…`, and classic returns it `state=active isAck=False acknowledgedBy=None acknowledgedAt=0001-01-01`. Browser: row `QAReFireTest qa-db-1 critical active` visible in the **Classic** tab with no ack badge. |
| Acknowledgment history still visible on the resolved alert | ✅ | `resolved_alerts.acknowledgments` for that fingerprint = `[{"username":"qaagent","reason":"QA acknowledge",…}]` after the delete. `displayMode=resolved` renders that incident (`startsAt 19:07:49`) as `isAck=True by=qaagent reason="QA acknowledge"`. |
| Un-acknowledging still works; a still-firing acknowledged alert stays acknowledged across refresh cycles | ✅ | Acked a firing alert (`QAAckActiveTest`): 3 consecutive refresh cycles all report `state=active isAck=True by=qaagent reason="QA active ack"`. Then `unacknowledge` → HTTP 200, DB rows `0`, alert back in classic as `state=active isAck=False`. `Acknowledged` tab correctly listed the acked alert and emptied after. |
| Test covers resolve → re-fire at the `AlertCache` level | ✅ | `go test -run TestAlertCache_AcknowledgmentDoesNotSurviveResolution ./internal/webui/services` → PASS. Also `TestCreateResolvedAlertClearsAcknowledgments`, `TestCleanupResolvedAcknowledgments` → PASS. Full suite for the touched packages: 75 passed, exit 0. |
| Orphaned ack rows in existing deployments addressed | ✅ | Seeded two rows: one keyed on a fingerprint present in `resolved_alerts`, one on a never-resolved key. Backend restart logged `✅ Cleaned up 1 acknowledgments left behind by resolved alerts`; the orphan was gone, the never-resolved one survived. |

**Exploratory:** ❌ `cleanupResolvedAcknowledgments` runs on **every** backend startup, not once, and it deletes by fingerprint alone — so it also deletes *live* acknowledgments of alerts that are firing **right now** but whose previous incident is still inside the resolved-alerts retention window (24h default). Reproduced twice: acknowledged the currently-firing `QAReFireTest` (live ack row = 1, 2 prior snapshots in `resolved_alerts`) → `docker restart notificator-backend` → log `✅ Cleaned up 1 acknowledgments left behind by resolved alerts`, live ack rows = **0**, ack silently lost. That is precisely the recurring-alert population this PR is about, so any deploy or pod restart drops those acknowledgments. Two ways out: bound the delete to acks that predate the resolution (e.g. only where `acknowledgments.created_at < resolved_alerts.resolved_at`), or gate the cleanup behind a run-once marker so it stays a one-off migration.

Everything the issue asked for works. Blocking on the regression the cleanup migration introduces, since it silently discards a user's acknowledgment.

**Verdict: FAIL**

SoulKyu:

internal/backend/services/acknowledged_alerts_test.go
```diff
@@ -0,0 +1,33 @@
...
+// applyAcknowledgments clears every fingerprint missing from the response, so an
+// empty map returned with a nil error would un-acknowledge the whole dashboard
+// on a statement timeout or a failover. The failure has to reach the caller.
+func TestGetAllAcknowledgedAlerts_SurfacesDatabaseError(t *testing.T) {
+	// No AutoMigrate: the acknowledgments table is missing, so the query fails.
+	db, err := database.NewGormDB("sqlite", config.DatabaseConfig{SQLitePath: t.TempDir() + "/test.db"})
+	if err != nil {
+		t.Fatalf("failed to create test database: %v", err)
+	}
+	svc := NewAlertServiceGorm(db)
```
SoulKyu: `NewAlertServiceGorm(db)` is called with a single argument here, but the constructor (`services.go:441`, unchanged by this PR) takes two: `NewAlertServiceGorm(db *database.GormDB, adminConfig *config.AdminConfig)`. This is a compile error, confirmed locally against this PR's head (e842c267):

```
internal/backend/services/acknowledged_alerts_test.go:22:29: not enough arguments in call to NewAlertServiceGorm
	have (*database.GormDB)
	want (*database.GormDB, *config.AdminConfig)
```

Because Go fails an entire package's test binary on one bad file, this doesn't just skip `TestGetAllAcknowledgedAlerts_SurfacesDatabaseError` — it also silently breaks the pre-existing `impersonation_test.go` and `remove_resolved_alerts_test.go` tests in the same package, so `go test ./internal/backend/services/...` reports zero passing tests rather than the PR description's claimed pass. Fix by matching the pattern the sibling tests already use, e.g. `svc := NewAlertServiceGorm(db, &config.AdminConfig{})` (see `remove_resolved_alerts_test.go:28`).

SoulKyu (commented):
@SoulKyu Nice follow-through on the previous review round: `cleanupResolvedAcknowledgments` now gates on `resolved_at > created_at` so it no longer sweeps up acknowledgments placed on a re-firing (verified in `TestCleanupResolvedAcknowledgments`), `GetAllAcknowledgedAlerts` surfaces DB failures as a gRPC error instead of an empty map, and `applyAcknowledgments` correctly returns before clearing the cache when that error hits. The transaction-based snapshot-then-delete in `CreateResolvedAlert` also closes the data-retention gap from before by reading the live rows itself when the caller's snapshot is empty.

One new problem stops this from being mergeable as-is: the new test file doesn't compile, so it breaks the whole `internal/backend/services` package's test suite, not just the new test. Details inline.

_Posted as a COMMENT review because GitHub does not allow requesting changes on your own pull request; the item below is meant as blocking._




internal/backend/services/acknowledged_alerts_test.go
SoulKyu: @SoulKyu This has been fixed in cacc39c.

Changed `NewAlertServiceGorm(db)` to `NewAlertServiceGorm(db, &config.AdminConfig{})` in internal/backend/services/acknowledged_alerts_test.go:22, matching the sibling test pattern. `go build ./...` and `go test ./internal/backend/services/...` now pass (16 tests).




SoulKyu:
**Looper fixer round complete** — cacc39c

- ✅ Review comment on `internal/backend/services/acknowledged_alerts_test.go` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/105#discussion_r3651915108)
  - Changed `NewAlertServiceGorm(db)` to `NewAlertServiceGorm(db, &config.AdminConfig{})` in internal/backend/services/acknowledged_alerts_test.go:22, matching the sibling test pattern. `go build ./...` and `go test ./internal/backend/services/...` now pass (16 tests).


SoulKyu:

SoulKyu (commented):
@SoulKyu This closes out the loop cleanly across all three review rounds. I re-verified the full diff at the current head against every point raised previously:

- The startup cleanup migration in internal/backend/database/migrate.go now gates its delete on the resolution timestamp being after the acknowledgment's creation timestamp, so it only sweeps orphans from the resolution that actually ended, not acknowledgments placed on a re-firing inside the retention window. The new migration test covers exactly that boundary.
- GetAllAcknowledgedAlerts in internal/backend/services/services.go now surfaces database failures as a gRPC Internal error instead of an empty map, and the webui cache's loader in internal/webui/services/alert_cache.go returns before touching the cache when that error hits, so a transient database problem can no longer mass-unacknowledge the dashboard.
- CreateResolvedAlert in internal/backend/database/gorm_db.go now snapshots then deletes inside a single transaction, reading the live rows itself when the caller's payload is empty or missing, so the data-retention gap from the second round is closed.
- The compile break from the previous round, a service constructor call missing its new admin-config argument in the new test file, is fixed in the latest commit.

I rebuilt and ran the full suite locally: the build succeeds and the backend and webui-services test suites pass in full, including all the new tests added across this PR's rounds. The raw SQL in the migration and the GORM queries are all parameterized, no injection concerns. The openwiki backend doc was updated to match the new lifecycle. No actionable findings remain, nice, methodical work chasing the blast radius of these deletes down to the edge cases.




SoulKyu:

SoulKyu (commented):
@SoulKyu I re-reviewed the current head against the prior three review rounds on this PR. The only content change since the last clean pass is a documentation line in `openwiki/backend.md` that was carried in by rebasing onto `main` — it updates the encryption-key note to match the mandatory-key behavior added in a separate, already-merged PR, and it reads correctly.

The actual fix code is unchanged and still holds up:

- `internal/backend/database/gorm_db.go`: `CreateResolvedAlert` snapshots and deletes the live `acknowledgments` rows for a fingerprint inside a single transaction, reading the rows itself via `snapshotAcknowledgments` when the caller's payload is empty (`isEmptyAcknowledgmentSnapshot` handles the nil/null/empty-array/empty-object cases), so history is never lost even if the caller's own fetch failed.
- `internal/backend/database/migrate.go`: `cleanupResolvedAcknowledgments` runs on every startup but correctly gates its delete on the resolution timestamp being after the acknowledgment's creation timestamp, so it only removes orphans from a resolution that already happened and leaves an acknowledgment placed on a re-firing alone.
- `internal/backend/services/services.go`: `GetAllAcknowledgedAlerts` now returns a gRPC internal error on a database failure instead of an empty map.
- `internal/webui/services/alert_cache.go`: `loadAcknowledgmentsEfficiently` returns before calling `applyAcknowledgments` when that error hits, so a transient database failure can no longer mass-unacknowledge the dashboard; I traced the client call path in `internal/webui/client` to confirm the error actually propagates through, and it does.

I ran the Go build and the backend and webui-services test suites locally at this head — everything compiles and all tests pass, including the tests covering resolved-alert acknowledgment clearing, the startup cleanup migration boundary, the database-error surfacing, and the full cache-level resolve-then-refire cycle. No actionable findings remain. Thanks for chasing this one all the way down to the restart and database-failure edge cases — solid, well-tested fix.




SoulKyu:
## 🧪 QA Report

Env: `make docker-build-all` + `docker compose up -d` (postgres + fake alertmanager + backend + webui), webui refresh 10s. Alerts driven through the fake Alertmanager (`POST /api/v2/alerts`); resolution triggered by making the source stop reporting the labels. Logged in as `qaagent`; UI checked in a real Chromium (Playwright).

|---|---|---|
| After an acknowledged alert resolves and later fires again with identical labels, the new firing shows as **not** acknowledged and appears in the default classic dashboard | ✅ | Two independent runs. `QAReFire105` (fp `fb03ad76…`) acked while firing → source stopped reporting → resolved → same labels re-posted → same fingerprint, classic returns `active isAck=False by=None acknowledgedAt=0001-01-01`. Repeated with `QABrowser105` (fp `5ca78dc3…`) and checked in the browser: Classic tab row `QABrowser105 qa-node-9 critical active … Fake Alertmanager`, no ack badge, and the **Acknowledged** tab empty for it. |
| Acknowledgment history for the resolved incident is still visible on the resolved alert | ✅ | `resolved_alerts.acknowledgments` for all three resolved incidents holds the full snapshot, e.g. `[{"username":"qaagent","reason":"QA browser ack incident 1",…}]`. `displayMode=resolved` returns each incident separately with `isAck=True by=qaagent reason=…` while the live re-firing shows `isAck=False`. Browser Resolved tab lists the incident (that table is an aggregate view — per-incident ack attribution was verified through the dashboard API that feeds it). |
| Un-acknowledging still works, and a still-firing acknowledged alert stays acknowledged across refresh cycles | ✅ | Acked the live firing: 5 samples across 4 refresh cycles all `active isAck=True by=qaagent reason="QA ack incident 2"`. Then `unacknowledge` → HTTP 200, `acknowledgments` rows `0`, alert back in classic as `active isAck=False`. |
| A test covers the resolve → re-fire cycle at the `AlertCache` level | ✅ | `TestAlertCache_AcknowledgmentDoesNotSurviveResolution` PASS, plus `TestCreateResolvedAlertClearsAcknowledgments`, `TestCreateResolvedAlertSnapshotsAcknowledgmentsWhenCallerHasNone`, `TestCleanupResolvedAcknowledgments`, `TestGetAllAcknowledgedAlerts_SurfacesDatabaseError`. Full suite on the touched packages (`webui/services`, `backend/database`, `backend/services`): 90 passed, exit 0. |
| Orphaned ack rows in existing deployments are addressed | ✅ | Seeded three rows, then restarted the backend: log `✅ Cleaned up 1 acknowledgments left behind by resolved alerts`. Deleted: the orphan whose `created_at` predates `resolved_alerts.resolved_at`. Kept: the ack of a never-resolved fingerprint **and** the live ack of the re-fired alert (`created_at` after `resolved_at`). The regression the previous QA run flagged (restart silently dropping live acks of recurring alerts) is fixed by the `r.resolved_at > acknowledgments.created_at` predicate. |

**Exploratory:** the risky part of this change is `applyAcknowledgments` now clearing every fingerprint the backend did not return, so I ran a backend outage: `docker stop notificator-backend` for ~35s. The webui logged `Failed to load acknowledged alerts from backend: rpc error … Unavailable` and returned before applying, and after restart the acknowledged alert was still `isAck=True by=qaagent` — no mass un-acknowledgment (the `services.go` change from empty-map to gRPC error is what makes this safe). Adjacent flows fine: search filter, Classic/Acknowledged/Resolved mode switching, ack audit comment still created (`commentCount 1`). Two unrelated nits, neither from this diff: right after acknowledging the dashboard briefly shows `acknowledgedBy` as the raw user id until the next refresh replaces it with the username, and `cleanupResolvedAcknowledgments` logs twice per backend start (idempotent, harmless).

**Verdict: PASS**

PR #110, opened 2026-07-26
fix(backend): require a real NOTIFICATOR_ENCRYPTION_KEY for Sentry token encryption

### LINKED ISSUE #63: fix(backend): Sentry personal tokens are encrypted with a hardcoded key that no deployment path overrides
## Problem

User Sentry personal access tokens are stored "encrypted" with a key hardcoded in this
repository, and no shipped deployment path sets a real one.

`internal/backend/database/sentry_db.go:16-31`:

```go
func getEncryptionKey() []byte {
	key := os.Getenv("NOTIFICATOR_ENCRYPTION_KEY")
	if key == "" {
		// Generate a default key for development - DO NOT USE IN PRODUCTION
		key = "dev-encryption-key-32-bytes-long"
	}
	// Ensure key is 32 bytes for AES-256
	keyBytes := []byte(key)
	if len(keyBytes) < 32 {
		// Pad with zeros
		padded := make([]byte, 32)
		copy(padded, keyBytes)
		return padded
	}
	return keyBytes[:32]
}
```

That literal is exactly 32 bytes, so `keyBytes[:32]` returns it untouched and it becomes the
AES-256 key for the GCM seal (`sentry_db.go:34-54`) and open (`sentry_db.go:56-86`) that protect
`UserSentryConfig.PersonalToken` in `SaveUserSentryConfig` / `GetUserSentryConfig`
(`sentry_db.go:88-125`). Anyone who reads this public repo has the key.

The fallback is silent on every surface an operator actually touches:

- **Prior art, and it does not help.** The repo wiki already documents the fallback —
  `openwiki/configuration.md:79` under a "⚠️ **Security gotcha**" heading, cross-referenced from
  `openwiki/backend.md:115`. That is documentation of a hazard, not a mitigation: nobody opens the
  wiki while writing a `values.yaml`.
- **Absent from every operator-facing config surface**: not in `.env.example`, not in
  `ENVIRONMENT_VARIABLES.md`, not in `docker-compose.yml`, not in
  `charts/notificator-app/values.yaml`. Contrast `NOTIFICATOR_SESSION_SECRET`, which is in
  `.env.example:5` and `docker-compose.yml:134`.
- **No startup warning, no log line.** `NOTIFICATOR_SESSION_SECRET` gets `⚠️  WARNING` twice
  (`internal/webui/router.go:124-125`) when it is missing. The encryption key gets silence.

An operator who follows the README, the Compose file, or the Helm chart therefore encrypts
production tokens with a key published on GitHub, and nothing in the running system says so.

Secondary hardening gap in the same function: a key shorter than 32 bytes is zero-padded
(`sentry_db.go:23-29`), so `NOTIFICATOR_ENCRYPTION_KEY=secret` silently becomes 6 bytes plus 26 NUL
bytes and is accepted as AES-256.

## Impact

- A Sentry personal token is a real credential: it grants API access to the org's Sentry issues and
  events for as long as it lives. It is user-supplied via `POST /api/v1/dashboard/sentry-token`
  (`internal/webui/router.go:293`) and stored per user.
- Encryption at rest is the stated protection — the RPC layer is careful, `GetUserSentryToken`
  refuses to hand another user's token out (`internal/backend/services/services.go:2092-2105`) —
  but that protection is void: anyone who obtains the database contents through any channel (a
  Postgres backup, a snapshot, a read replica, a `pg_dump` artifact in CI, a stolen SQLite file)
  decrypts every user's token with a constant copied from this repo.
- The failure is silent, so a security review of a running deployment cannot distinguish "properly
  keyed" from "using the published dev key" without inspecting the process environment.

## Proposed approach

Required behaviour. No alternatives, no feature flags.

1. **Fail fast at startup, unconditionally.** The backend refuses to start when
   `NOTIFICATOR_ENCRYPTION_KEY` is unset or invalid: log the reason and exit non-zero before the
   gRPC server begins serving (validation belongs with the other startup wiring in
   `internal/backend/server.go`, `NewServer` at `:40` / the `Start` path).
   The check is unconditional because there is no Sentry feature gate to condition it on:
   `internal/backend/server.go` contains no mention of Sentry, and the three Sentry RPCs
   (`services.go:2040`, `:2074`, `:2125`) plus `POST /api/v1/dashboard/sentry-token`
   (`internal/webui/router.go:293`) are always registered. "Only when Sentry can be used" is not
   expressible in this codebase; anything conditional would be invented state.
2. **One key format: 64 lowercase hex characters, decoded to 32 raw bytes.** This matches the shape
   already documented for the sibling secret (`.env.example:5`:
   `NOTIFICATOR_SESSION_SECRET=your-64-character-hex-secret-here`). Anything else — wrong length,
   non-hex — is rejected at the startup check with a message naming the variable and
   `openssl rand -hex 32`. Delete the zero-padding branch entirely; `getEncryptionKey` returns the
   validated 32 bytes and never fabricates a key.
3. **Document it where every other secret is documented**: `.env.example`,
   `ENVIRONMENT_VARIABLES.md`, `docker-compose.yml`, and the `webui`/`backend` `env` comment blocks
   in `charts/notificator-app/values.yaml`, each with the generation one-liner.
4. **Correct the wiki**, which currently describes the fallback as normal behaviour:
   `openwiki/configuration.md:79` and `openwiki/backend.md:115` must say the key is mandatory.

### Breaking change and migration

This is a deliberate breaking change, and the issue owns it:

- **Every existing deployment must set `NOTIFICATOR_ENCRYPTION_KEY` before the next rollout**,
  including deployments that never use Sentry. Without it the backend exits at startup instead of
  starting with a published key.
- **Tokens stored under the old key are unrecoverable.** Two populations are affected: deployments
  that never set the variable (tokens encrypted under `dev-encryption-key-32-bytes-long`) and
  deployments that set a key shorter than 32 bytes (tokens encrypted under the zero-padded value,
  whose key is now rejected outright). Both must re-enter their Sentry personal token through the
  Sentry settings modal. No migration script: a token encrypted with a key published on GitHub
  should be rotated in Sentry anyway.
- A decrypt failure already surfaces as an error from `GetUserSentryConfig` (`sentry_db.go:96-102`),
  so the failure mode is an error on read, not corruption.
- The same consequence applies to any later key rotation.

## Acceptance criteria

- [ ] `rg 'dev-encryption-key'` returns nothing anywhere in the tree.
- [ ] Starting the backend with `NOTIFICATOR_ENCRYPTION_KEY` unset logs an error naming the
      variable and the `openssl rand -hex 32` command, and exits non-zero before the gRPC server
      accepts a connection. It does not start with a built-in key, and it does not start with
      Sentry silently disabled.
- [ ] Starting the backend fails identically for each of these invalid keys: `secret`, a 63-hex
      string, a 65-hex string, a 64-character non-hex string.
- [ ] `rg 'Pad with zeros' internal/backend/` returns nothing; no code path fabricates or pads a key.
- [ ] With a valid 64-hex key, saving a Sentry personal token and reading it back through
      `GetUserSentryToken` returns the original value.
- [ ] `NOTIFICATOR_ENCRYPTION_KEY` is present with a generation command in all four of
      `.env.example`, `ENVIRONMENT_VARIABLES.md`, `docker-compose.yml`,
      `charts/notificator-app/values.yaml`.
- [ ] Upgrade notes and `openwiki/configuration.md` / `openwiki/backend.md` state: the key is now
      mandatory, the backend will not start without it, and Sentry tokens stored under the previous
      dev or zero-padded key are unrecoverable and must be re-entered — as is the case for any
      future key rotation.

SoulKyu:
## Summary
- Sentry personal access tokens were "encrypted" with a key hardcoded in this repo (`dev-encryption-key-32-bytes-long`) whenever `NOTIFICATOR_ENCRYPTION_KEY` was unset, and any shorter configured key was silently zero-padded to 32 bytes — both cases were invisible on every deployment surface (no `.env.example`/`docker-compose.yml`/Helm entry, no startup warning).
- `internal/backend/database/sentry_db.go`: replaced `getEncryptionKey`'s silent fallback/padding with strict validation — the key must be 64 lowercase hex characters (32 raw bytes), matching the format already used for `NOTIFICATOR_SESSION_SECRET`. Added `ValidateEncryptionKey()` for startup use.
- `internal/backend/server.go`: `Start()` now calls `database.ValidateEncryptionKey()` first and returns an error (which `cmd/backend.go` turns into `log.Fatalf`, i.e. non-zero exit) before initializing the database or starting the gRPC server — unconditionally, since Sentry RPCs are always registered.
- Documented `NOTIFICATOR_ENCRYPTION_KEY` (with the `openssl rand -hex 32` generation command) in `.env.example`, `ENVIRONMENT_VARIABLES.md`, `docker-compose.yml`, and the backend `env` block of `charts/notificator-app/values.yaml`.
- Corrected `openwiki/configuration.md` and `openwiki/backend.md`, which previously documented the fallback as expected behavior — they now state the key is mandatory and that rotating/first-setting it makes previously stored tokens unrecoverable.

## Breaking change
Every existing deployment must set `NOTIFICATOR_ENCRYPTION_KEY` before the next rollout — the backend now refuses to start without it, even for deployments that don't use Sentry. Sentry personal tokens stored under the old hardcoded or zero-padded key are unrecoverable and must be re-entered via the Sentry settings modal (decrypt failures already surfaced as an error from `GetUserSentryConfig`, so this is a read-time error, not corruption).

Closes #63

## Test plan
- [x] `go build ./...`
- [x] `go vet ./internal/backend/...`
- [x] `go test ./internal/backend/...` (25 passed, incl. new `sentry_db_test.go` covering unset/short/63-hex/64-hex/65-hex/non-hex keys and an encrypt/decrypt round trip)
- [x] `rg 'dev-encryption-key'` and `rg 'Pad with zeros' internal/backend/` both return nothing

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -3,3 +3,7 @@
 # Session secret for cookie encryption (REQUIRED for session persistence across restarts)
 # Generate with: openssl rand -hex 32
 NOTIFICATOR_SESSION_SECRET=your-64-character-hex-secret-here
+
+# Encryption key for Sentry personal tokens at rest (REQUIRED - backend refuses to start without it)
+# Generate with: openssl rand -hex 32
+NOTIFICATOR_ENCRYPTION_KEY=your-64-character-hex-secret-here
--- a/ENVIRONMENT_VARIABLES.md
+++ b/ENVIRONMENT_VARIABLES.md
@@ -45,6 +45,14 @@ The following standard database environment variables are also supported:
 - `DB_SSL_MODE` / `DATABASE_SSL_MODE` - SSL mode
 - `DB_PATH` / `DATABASE_PATH` - SQLite file path
 
+## Security Configuration
+
+- `NOTIFICATOR_ENCRYPTION_KEY` - **Required.** AES-256 key encrypting Sentry personal tokens at
+  rest, as 64 lowercase hex characters (32 bytes). The backend logs an error and exits non-zero at
+  startup if this is unset or invalid. Generate with: `openssl rand -hex 32`.
+  Rotating this key makes previously stored Sentry personal tokens unrecoverable; affected users
+  must re-enter their token via the Sentry settings modal.
+
 ## WebUI Configuration
 
 - `NOTIFICATOR_WEBUI_LISTEN` - WebUI server listen address (default: ":8081")
--- a/charts/notificator-app/values.yaml
+++ b/charts/notificator-app/values.yaml
@@ -55,7 +55,11 @@ backend:
     # NOTIFICATOR_BACKEND_ENABLED: "true"
     # NOTIFICATOR_BACKEND_GRPC_LISTEN: ":50051"
     # NOTIFICATOR_BACKEND_HTTP_LISTEN: ":8080"
-    
+
+    # Encryption key for Sentry personal tokens at rest (REQUIRED - backend refuses to start
+    # without it). 64 lowercase hex characters, generate with: openssl rand -hex 32
+    # NOTIFICATOR_ENCRYPTION_KEY: "your-64-character-hex-secret-here"
+
     # Database configuration - PostgreSQL URL (recommended)
     # POSTGRES_URL: "postgres://user:password@host:5432/database?sslmode=require"
     
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -87,6 +87,9 @@ services:
       - NOTIFICATOR_SENTRY_BASE_URL=https://sentry.example.com
       - NOTIFICATOR_SENTRY_GLOBAL_TOKEN=""
 
+      # Encryption key for Sentry personal tokens at rest (REQUIRED, generate with: openssl rand -hex 32)
+      - NOTIFICATOR_ENCRYPTION_KEY=${NOTIFICATOR_ENCRYPTION_KEY}
+
       # Logging
       - NOTIFICATOR_LOG_LEVEL=info
     depends_on:
--- a/internal/backend/database/sentry_db.go
+++ b/internal/backend/database/sentry_db.go
@@ -5,6 +5,7 @@ import (
 	"crypto/cipher"
 	"crypto/rand"
 	"encoding/base64"
+	"encoding/hex"
 	"fmt"
 	"io"
 	"os"
@@ -12,22 +13,38 @@ import (
 	"notificator/internal/backend/models"
 )
 
-// Encryption key for tokens - should be from environment variable
+// EncryptionKeyEnvVar is the environment variable holding the AES-256 key
+// used to encrypt Sentry personal tokens at rest.
+const EncryptionKeyEnvVar = "NOTIFICATOR_ENCRYPTION_KEY"
+
+// ValidateEncryptionKey checks that NOTIFICATOR_ENCRYPTION_KEY is set to 64
+// lowercase hex characters (32 raw bytes), returning an error naming the
+// variable and the generation command otherwise.
+func ValidateEncryptionKey() error {
+	_, err := encryptionKeyFromEnv()
+	return err
+}
+
+func encryptionKeyFromEnv() ([]byte, error) {
+	key := os.Getenv(EncryptionKeyEnvVar)
+	if len(key) != 64 {
+		return nil, fmt.Errorf("%s must be set to 64 lowercase hex characters (32 bytes); generate one with `openssl rand -hex 32`", EncryptionKeyEnvVar)
+	}
+	keyBytes, err := hex.DecodeString(key)
+	if err != nil {
+		return nil, fmt.Errorf("%s must be 64 lowercase hex characters (32 bytes); generate one with `openssl rand -hex 32`", EncryptionKeyEnvVar)
+	}
+	return keyBytes, nil
+}
+
 func getEncryptionKey() []byte {
-	key := os.Getenv("NOTIFICATOR_ENCRYPTION_KEY")
-	if key == "" {
-		// Generate a default key for development - DO NOT USE IN PRODUCTION
-		key = "dev-encryption-key-32-bytes-long"
+	key, err := encryptionKeyFromEnv()
+	if err != nil {
+		// ValidateEncryptionKey is called at startup, so this can only be
+		// reached if the environment changed after the process started.
+		panic(err)
 	}
-	// Ensure key is 32 bytes for AES-256
-	keyBytes := []byte(key)
-	if len(keyBytes) < 32 {
-		// Pad with zeros
-		padded := make([]byte, 32)
-		copy(padded, keyBytes)
-		return padded
-	}
-	return keyBytes[:32]
+	return key
 }
 
 // Encrypt encrypts plaintext using AES
--- a/internal/backend/database/sentry_db_test.go
+++ b/internal/backend/database/sentry_db_test.go
@@ -1,0 +1,48 @@
+package database
+
+import "testing"
+
+func TestValidateEncryptionKey(t *testing.T) {
+	cases := map[string]struct {
+		key     string
+		wantErr bool
+	}{
+		"unset":        {"", true},
+		"too short":    {"secret", true},
+		"63 hex chars": {"e4b6836c113e437c8353013501173098d2f496cd9d70afa96896057d495d88c", true},
+		"64 hex chars": {"e4b6836c113e437c8353013501173098d2f496cd9d70afa96896057d495d88cd", false},
+		"65 hex chars": {"e4b6836c113e437c8353013501173098d2f496cd9d70afa96896057d495d88cd0", true},
+		"64 non-hex":   {"zzb6836c113e437c8353013501173098d2f496cd9d70afa96896057d495d88cd", true},
+	}
+
+	for name, tc := range cases {
+		t.Run(name, func(t *testing.T) {
+			t.Setenv(EncryptionKeyEnvVar, tc.key)
+			err := ValidateEncryptionKey()
+			if tc.wantErr && err == nil {
+				t.Fatalf("expected error for key %q, got nil", tc.key)
+			}
+			if !tc.wantErr && err != nil {
+				t.Fatalf("expected no error for key %q, got %v", tc.key, err)
+			}
+		})
+	}
+}
+
+func TestEncryptDecryptRoundTrip(t *testing.T) {
+	t.Setenv(EncryptionKeyEnvVar, "e4b6836c113e437c8353013501173098d2f496cd9d70afa96896057d495d88cd")
+
+	const token = "sentry-personal-token-abc123"
+	ciphertext, err := encrypt(token)
+	if err != nil {
+		t.Fatalf("encrypt: %v", err)
+	}
+
+	plaintext, err := decrypt(ciphertext)
+	if err != nil {
+		t.Fatalf("decrypt: %v", err)
+	}
+	if plaintext != token {
+		t.Fatalf("expected %q, got %q", token, plaintext)
+	}
+}
--- a/internal/backend/server.go
+++ b/internal/backend/server.go
@@ -46,6 +46,10 @@ func NewServer(cfg *config.Config, dbType string) *Server {
 }
 
 func (s *Server) Start() error {
+	if err := database.ValidateEncryptionKey(); err != nil {
+		return fmt.Errorf("invalid encryption key configuration: %w", err)
+	}
+
 	if err := s.initDatabase(); err != nil {
 		return fmt.Errorf("failed to initialize database: %w", err)
 	}
--- a/openwiki/backend.md
+++ b/openwiki/backend.md
@@ -111,5 +111,6 @@ only** (no cross-replica fan-out). See [architecture](architecture.md#real-time)
   is inline in `AlertServiceGorm`.
 - **Silent job drops:** statistics worker pool drops events when full, with no alerting.
 - **Single-replica constraint:** in-memory subscriptions break under horizontal scaling.
-- **Encryption key fallback:** Sentry token storage falls back to a hardcoded dev key if
-  `NOTIFICATOR_ENCRYPTION_KEY` is unset — see [configuration](configuration.md#sentry).
+- **Encryption key is mandatory:** the backend refuses to start (non-zero exit before the gRPC
+  server accepts connections) unless `NOTIFICATOR_ENCRYPTION_KEY` is set to a valid 64-hex key —
+  see [configuration](configuration.md#sentry).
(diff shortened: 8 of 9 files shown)


SoulKyu (commented):
@SoulKyu Reviewed this fix for the Sentry personal-token encryption key handling.

What changed: `internal/backend/database/sentry_db.go` now requires `NOTIFICATOR_ENCRYPTION_KEY` to be exactly 64 lowercase hex characters (32 raw bytes) via the new `encryptionKeyFromEnv`/`ValidateEncryptionKey` helpers, replacing the old silent fallback to a hardcoded dev key and zero-padding of short keys. `internal/backend/server.go` calls `ValidateEncryptionKey()` at the top of `Start()` so the backend now fails fast with a clear error (turned into `log.Fatalf` in `cmd/backend.go`) instead of silently encrypting tokens with a predictable key. The env var is now documented consistently in `.env.example`, `ENVIRONMENT_VARIABLES.md`, `docker-compose.yml`, and the Helm chart's `values.yaml`, and the openwiki docs are corrected to describe the mandatory behavior and the unrecoverable-tokens-on-rotation caveat instead of the old fallback.

Verification: confirmed `go build ./...` and `go test ./internal/backend/database/...` pass locally, checked that `getEncryptionKey`/`encrypt`/`decrypt` have no other call sites that could run before `Start()`'s validation (so the documented panic path is unreachable in normal operation), and read through the new `sentry_db_test.go` boundary cases (unset, 63/64/65 hex chars, non-hex, round trip) — the fixture lengths and hex content check out. Docs across `.env.example`, `ENVIRONMENT_VARIABLES.md`, `docker-compose.yml`, `values.yaml`, and openwiki are all consistent with the new mandatory-key behavior.

This closes a real security gap (tokens previously encrypted under a publicly-known hardcoded key by default) with good test coverage and correctly documents the breaking change for existing deployments. Nice work tightening this up.



SoulKyu:
## 🧪 QA Report

Env: `docker compose` stack with images built from this branch (`Dockerfile.backend` / `Dockerfile.webui`), Postgres 15, fake Alertmanager, plus a stub Sentry API on the compose network so the real token-save flow (which pre-validates against `GET {base_url}/api/0/`) could be exercised end-to-end.

|---|---|---|
| `rg 'dev-encryption-key'` returns nothing anywhere in the tree | ✅ | `grep -rn "dev-encryption-key" .` (excluding `.git`) → no match, exit 1 |
| Unset `NOTIFICATOR_ENCRYPTION_KEY` → error naming the var + `openssl rand -hex 32`, exit non-zero before gRPC accepts a connection | ✅ | Ran the backend image against live Postgres with the var unset: `Backend server failed: invalid encryption key configuration: NOTIFICATOR_ENCRYPTION_KEY must be set to 64 lowercase hex characters (32 bytes); generate one with \`openssl rand -hex 32\`` — exit code 1, and no `gRPC server starting on :50051` line was ever emitted. Same for an explicitly empty value. |
| Fails identically for `secret`, 63-hex, 65-hex, 64-char non-hex | ✅ | Four separate container runs, all exit 1, all with the same `NOTIFICATOR_ENCRYPTION_KEY must be…` message and no gRPC listen line (the non-hex case takes the `must be 64 lowercase hex…` decode branch, same content) |
| `rg 'Pad with zeros' internal/backend/` returns nothing; no path fabricates/pads a key | ✅ | `grep -rn "Pad with zeros" internal/backend/` → no match, exit 1; `getEncryptionKey` now only returns `hex.DecodeString` output |
| With a valid 64-hex key, saving a token and reading it back through `GetUserSentryToken` returns the original | ✅ | Saved `sntrys_qa_ui_flow_777` through the real browser flow (Playwright: login → Dashboard Settings → **Sentry** tab → paste → **Save**, UI showed "Sentry token saved successfully" / "Token configured"). DB column holds ciphertext (`dhxeix4SI8muwaF4nJYeo7IjFRjVoG6A…`), and a direct gRPC call `GetUserSentryToken(user_id, session_id)` against `:50051` returned `success=true has_token=true token="sntrys_qa_ui_flow_777"`. A bogus session_id returns `Invalid session` — isolation intact. |
| Documented with a generation command in `.env.example`, `ENVIRONMENT_VARIABLES.md`, `docker-compose.yml`, `charts/notificator-app/values.yaml` | ✅ | Present in all four, each with `openssl rand -hex 32`; compose wires `NOTIFICATOR_ENCRYPTION_KEY=${NOTIFICATOR_ENCRYPTION_KEY}` and the stack booted healthy with it supplied |
| Wiki / upgrade notes state: mandatory, backend won't start without it, old tokens unrecoverable and must be re-entered | ✅ | `openwiki/configuration.md` states all three (incl. re-entry via the Sentry settings modal and the same consequence for rotation); `openwiki/backend.md` states mandatory + non-zero exit before gRPC serves and links to it; `ENVIRONMENT_VARIABLES.md` carries the rotation/re-entry note. The repo has no dedicated upgrade-notes file — the migration text lives in those docs plus the PR's "Breaking change" section. |

**Exploratory:** Rotating to a different valid key with a token already stored behaves as documented — the backend still starts, the row is untouched (68 bytes), and the user simply sees `has_token:false` / `auth_status:"none"` again, i.e. re-enter the token; no corruption, no crash (note it surfaces as "no token configured" rather than an explicit decrypt error, which is the pre-existing `GetUserSentryConfig` behavior). Login, dashboard alert list, the settings modal's other tabs and the Sentry test/save/remove controls all still work; `go test ./internal/backend/database/...` passes. Two unrelated 401s (`/api/v1/profile/timezone`, `/api/impersonate/status`) fire on dashboard load — outside this diff. One deviation worth noting, not a criterion: `hex.DecodeString` is case-insensitive, so a 64-char **uppercase** hex key is accepted even though the message and docs say "64 lowercase hex characters" (verified: backend started and logged `gRPC server starting on :50051`). Harmless — it is still 32 bytes of real entropy.

**Verdict: PASS**

SoulKyu:

SoulKyu (commented):
@SoulKyu Reviewed this fix for the Sentry personal-token encryption key handling.

What changed: `internal/backend/database/sentry_db.go` now requires `NOTIFICATOR_ENCRYPTION_KEY` to be exactly 64 lowercase hex characters (32 raw bytes) via the new `encryptionKeyFromEnv`/`ValidateEncryptionKey` helpers, fully replacing the old silent fallback to the hardcoded `dev-encryption-key-32-bytes-long` string and the zero-padding of short keys. `internal/backend/server.go` calls `ValidateEncryptionKey()` at the very top of `Start()`, unconditionally (Sentry RPCs are always registered), so the backend now fails fast with a clear, actionable error instead of silently encrypting tokens with a predictable key; `cmd/backend.go` turns that into `log.Fatalf`. The env var is documented consistently across `.env.example`, `ENVIRONMENT_VARIABLES.md`, `docker-compose.yml`, and the Helm chart's `values.yaml`, and the openwiki docs (`backend.md`, `configuration.md`) are corrected to describe the mandatory behavior and the unrecoverable-tokens-on-rotation caveat instead of the old fallback.

Verification performed: confirmed `go build ./...`, `go vet ./internal/backend/...`, and `go test ./internal/backend/database/...` all pass on the PR head; traced every call site of `getEncryptionKey`/`encrypt`/`decrypt` and confirmed none run before `Start()`'s `ValidateEncryptionKey()` call, so the panic path in `getEncryptionKey` is unreachable in normal operation; verified the new `sentry_db_test.go` boundary-case fixture strings (unset, 63/64/65 hex chars, non-hex) actually have the lengths their test-case names claim; grepped the repo for any remaining reference to the old hardcoded dev key or a bypass of the new validation and found none; confirmed `cmd/backend.go` treats `server.Start()`'s error as fatal so the documented "refuses to start" behavior actually holds.

This closes a real security gap (Sentry tokens previously encrypted under a publicly-known hardcoded key by default, or a zero-padded short key) with solid test coverage and correctly documents the breaking change for existing deployments. Nice, tight fix — thanks for tracking down and closing this one.



SoulKyu:
## 🧪 QA Report

Environment: `docker compose` stack (postgres 15 + fake alertmanager + backend image rebuilt from this branch + webui), backend `:8080`/`:50051`, webui `:8081`, both `healthy`.

|---|---|---|
| `rg 'dev-encryption-key'` returns nothing | ✅ | `rg dev-encryption-key` over the tree → 0 matches |
| Unset key → error naming the var + `openssl rand -hex 32`, non-zero exit before gRPC accepts a connection | ✅ | backend container against live postgres, no key: `exit=1`, `Backend server failed: invalid encryption key configuration: NOTIFICATOR_ENCRYPTION_KEY must be set to 64 lowercase hex characters (32 bytes); generate one with 'openssl rand -hex 32'`; no gRPC listen line, port 50051 never bound |
| Identical failure for `secret`, 63-hex, 65-hex, 64-char non-hex | ✅ | 4 runs of the same image with each key → all `exit=1` with the same message (`zzz…` 64-char non-hex variant: "must be 64 lowercase hex characters") |
| `rg 'Pad with zeros' internal/backend/` returns nothing; no path fabricates a key | ✅ | 0 matches; `getEncryptionKey` now panics instead of inventing a key, `encryptionKeyFromEnv` is the only source |
| Valid 64-hex key: save token, read back via `GetUserSentryToken` returns the original | ✅ | real gRPC client vs running backend: `Login` → `SaveUserSentryConfig("sntryu_qa110_secret_TOKEN_v1")` → `GetUserSentryToken` → `has_token=true match=true`; postgres row holds ciphertext `AzmAh6yqvandVEaURtXcQ48pbFAzMIKprr+lEvXY…` |
| Key documented with generation command in `.env.example`, `ENVIRONMENT_VARIABLES.md`, `docker-compose.yml`, `charts/notificator-app/values.yaml` | ✅ | all four contain the var and `openssl rand -hex 32` |
| Docs state key is mandatory + tokens under the old/rotated key unrecoverable and must be re-entered | ✅ | `openwiki/configuration.md:79-84`, `openwiki/backend.md:114-116`, `ENVIRONMENT_VARIABLES.md:50-55`; PR body carries the breaking-change/upgrade note (repo has no UPGRADING/CHANGELOG file) |

**Exploratory:** logged in through the real browser (Playwright) → dashboard renders, `GET /api/v1/dashboard/data` 200, Settings → Sentry tab shows "✓ Token configured" and `/sentry-config` returns `has_token=true`; webui container has no encryption key and is unaffected (enforcement is backend-only, not over-broad). Restarting the backend with a *different* valid 64-hex key makes the stored token unreadable (`GetUserSentryToken` → `has_token` false), confirming the env key really is the AES key and no fallback still decrypts. Three notes, none blocking: (1) `docker compose up` with the var absent from the shell/`.env` substitutes an empty string, so the backend crash-loops until the operator copies `.env.example` — correct fail-fast, but the shipped test env no longer boots out of the box; (2) `backend.migrate` runs `RunMigrations()` before `Start()`, so the DB is connected and migrated before the key check — still before gRPC serves, so the criterion holds, but not literally "before initializing the database" as the PR body says; (3) a 64-char *uppercase* hex key is accepted (`hex.DecodeString` is case-insensitive) although every message and doc says "lowercase" — cosmetic, and not one of the invalid cases the issue lists.

**Verdict: PASS**

PR #69, opened 2026-07-24
fix(dashboard): bulk-action failures are reported as success

SoulKyu:
## Problem

`POST /api/v1/dashboard/bulk-action` — the single funnel for every alert action — always answered with an outer envelope whose `success` was hardcoded `true` by `SuccessResponse()`, even when every alert failed. All twelve client call sites read only that outer flag, so a failed silence or acknowledgment closed the modal, cleared the selection, reloaded the data, and showed no error anywhere. The failure was invisible exactly when it matters most: Alertmanager degraded or unreachable during an incident.

### Server

- `BulkActionResponse` now carries a correlated per-target failure list (`failures: [{target, kind, error}]`, with `kind` = `"alert"` | `"group"`) instead of the consumer-less `Errors []string`. Populated in both the fingerprint loop and the group loop of `BulkActionAlerts`.
- The endpoint no longer calls `SuccessResponse`: it returns an `APIResponse` literal whose outer `success` is `false` whenever `failedCount > 0`, with an `"N of M actions failed"` summary in `error`. Still HTTP 200 — a partial success is not a transport error.

### Client (`dashboard_actions.templ`, regenerated with pinned templ v0.3.906)

- Every call site now treats a response as successful only when `result.success === true`.
- **Ack/silence modals** stay open on failure and show the backend error plus the first per-target failure messages in the existing `ackError`/`silenceError` slots.
- **Partial failure** reduces the selection to exactly the failed targets (`failures.filter(f => f.kind === 'alert')`) so the operator can retry them, and reloads data so the targets that did succeed update on screen.
- **Fire-and-forget actions** (unacknowledge, resolve, hide, unsilence, remove-resolved) surface failures through one shared `actionError` banner on the dashboard root (`showActionError()` helper, auto-dismiss after 8s, manual dismiss button). No more empty `else {}` blocks or `console.error`-only handling.
- Fully successful actions behave exactly as before: modal closes, selection cleared, data reloaded.

## Validation

- `go build ./...` and `go test ./...` pass (49 tests, 27 packages).
- New `dashboard_handlers_test.go` covers the three envelope shapes on the raw JSON: total failure (`success: false`, both `alert` and `group` failure entries), partial failure (`success: false`, only the failed fingerprint listed), and full success (`success: true`, empty `error`).
- `else {}` audit of `dashboard_actions.templ` returns nothing.

Closes #38

[...]

Issue #62, opened 2026-07-24
fix(backend): any authenticated user can wipe every resolved alert — RemoveAllResolvedAlerts has no session or admin check


SoulKyu:
## Problem

`DELETE /api/v1/dashboard/remove-resolved-alerts` hard-deletes **every** resolved alert in the
database, for **every** user, and the only thing standing in front of it is `RequireAuth()` —
i.e. any account at all.

The call chain has no ownership check, no role check, and — at the gRPC layer — no session check
at all:

- `internal/webui/router.go:267` — `dashboard.DELETE("/remove-resolved-alerts", handlers.RemoveAllResolvedAlerts)`
  sits in the `dashboard` group, which only applies `authMiddleware.RequireAuth()`.
- `internal/webui/handlers/dashboard_handlers.go:1865` — `RemoveAllResolvedAlerts` never looks at
  the current user; it only checks that `alertCache != nil`.
- `internal/webui/client/backend_client.go:540` — the gRPC request is built empty:
  `req := &alertpb.RemoveAllResolvedAlertsRequest{}`. The proto message itself has no fields
  (`proto/alert.proto:284-286`, "No parameters needed").
- `internal/backend/services/services.go:1145` — the backend RPC is the *only* `AlertService`
  mutating RPC with no `session_id` and no `GetUserBySession` call. It goes straight to the DB.
- `internal/backend/database/gorm_db.go:560` — `gdb.db.Delete(&models.ResolvedAlert{}, "1 = 1")`.
  `ResolvedAlert` (`internal/backend/models/models.go:150`) has no `gorm.DeletedAt`, so this is a
  **hard delete** — not recoverable without a DB restore.

Resolved alerts are not just a list of names: the archive carries the alert payload plus the
`comments` and `acknowledgments` JSONB captured at resolve time
(`internal/backend/models/models.go:157-158`). Wiping the table destroys the team's post-incident
record, not just a cache.

The UI exposes it as a red button in the dashboard settings modal
(`internal/webui/templates/components/modal_components.templ:154`), guarded by a single
`confirm()` (`internal/webui/templates/scripts/dashboard_actions.templ:384`). The server-side log
line (`services.go:1146`) does not record *who* triggered it, so after the fact there is no way to
tell who wiped the history.

## Impact

- Any authenticated user of a shared team dashboard — the least-privileged reader — can
  irreversibly destroy the resolved-alert history (with its archived comments and acks) for the
  whole organisation, by misclicking one button in a settings modal.
[...]

PR #133, opened 2026-07-27
fix(webui): check Success flag in acknowledge and comment RPCs

### LINKED ISSUE #128: fix(webui): acknowledge and comment RPCs ignore the backend's Success flag, so rejected writes are reported as successful
## Problem

The backend signals business-level failures in-band: every mutation handler returns a gRPC-OK response carrying `Success: false` plus a message (invalid session, DB write failure, unauthorized). See `internal/backend/services/services.go:473-489` (`AddComment`) and `:641-657` (`AddAcknowledgment`).

Four client wrappers throw that response away and only look at the transport error:

- `internal/webui/client/backend_client.go:325-326` — `_, err := c.alertClient.AddAcknowledgment(ctx, req); return err`
- `internal/webui/client/backend_client.go:343-344` — `DeleteAcknowledgment`, same shape
- `internal/webui/client/backend_client.go:458-459` — `AddComment`, same shape
- `internal/webui/client/backend_client.go:476-477` — `DeleteComment`, same shape

This is inconsistent with the file's own convention: neighbours such as `internal/webui/client/backend_client.go:301-303` do check `if !resp.Success { return fmt.Errorf("%s", resp.Error) }`.

Failure path, end to end:

1. The user posts a comment. `POST /dashboard/alerts/:fingerprint/comments` reaches `internal/webui/handlers/dashboard_handlers.go:1471`.
2. The backend rejects it — expired session, or `CreateComment` fails on a DB error — and answers `Success: false, Message: "Failed to create comment"` with gRPC status OK.
3. `AddComment` returns `nil`. The handler skips its error branch, bumps `CommentCount` in the cache (`:1478-1481`) and replies `200 {"message": "Comment added successfully"}` (`:1483-1485`).
4. The comment does not exist. The optimistic badge is silently rolled back by the next `loadCommentCountsEfficiently` cycle.

Same for acknowledgements: `internal/webui/handlers/dashboard_handlers.go:902` treats a rejected `AddAcknowledgment` as success, then marks the alert acknowledged locally at `:918-926`. Ten seconds later `applyAcknowledgments` mirrors the backend and the alert reappears unacknowledged, with no explanation and no error ever shown.

This also undercuts the bulk-action accounting added in #69/PR #62: `processAlertAction` (`internal/webui/handlers/dashboard_handlers.go:886`) is the function whose error the bulk summary counts, and it cannot see a backend rejection.

## Impact

During an incident a responder acknowledges an alert, sees the confirmation, moves on — and the alert comes back unacknowledged with no owner, or a written handover comment is lost. The user has every reason to believe the write landed. Expired sessions and short DB outages are exactly the conditions under which this happens, and they are also exactly when people are working the console hardest.

## Proposed approach

In `internal/webui/client/backend_client.go`, make the four wrappers read the response, matching the existing convention in the same file:

```go
resp, err := c.alertClient.AddComment(ctx, req)
if err != nil {
    return err
}
if !resp.GetSuccess() {
    return fmt.Errorf("%s", resp.GetMessage())
}
return nil
```

No handler changes needed — they already surface a non-nil error correctly. Worth a follow-up sweep of the statistics wrappers (`CaptureAlertFired`, `UpdateAlertResolved`, `UpdateAlertAcknowledged`, around `internal/webui/client/backend_client.go:1734-1783`), which drop `Success` the same way, though those are fire-and-forget and less visible.

## Acceptance criteria

- `AddComment`, `DeleteComment`, `AddAcknowledgment` and `DeleteAcknowledgment` return a non-nil error whenever the backend answers `Success: false`, carrying the backend's message.
- Posting a comment with an invalid session returns a non-2xx HTTP response and the cached `CommentCount` is not incremented.
- Acknowledging an alert the backend rejects returns a non-2xx response and does not set `IsAcknowledged` in the cache.
- Unit tests with a fake `AlertServiceClient` returning `Success: false` cover all four wrappers.
- A bulk action over a mix of accepted and rejected alerts reports the rejected ones in `FailedCount`.

SoulKyu:
## Problem

`AddAcknowledgment`, `DeleteAcknowledgment`, `AddComment` and `DeleteComment` in `internal/webui/client/backend_client.go` discarded the gRPC response and only checked the transport error:

```go
_, err := c.alertClient.AddAcknowledgment(ctx, req)
return err
```

The backend signals business-level failures (invalid session, DB write failure, unauthorized) in-band: it returns a gRPC-OK response carrying `Success: false` plus a `Message` (see `internal/backend/services/services.go` `AddComment`/`AddAcknowledgment`). Because the wrappers ignored `Success`, a rejected write was reported back to the caller as `nil` error — an acknowledged alert could silently revert ten seconds later with no explanation, and a posted comment could be lost while the UI showed a 200 OK.

This was inconsistent with the file's own convention — neighbours like `UpdateTimezone` already check `if !resp.Success { return fmt.Errorf(...) }`.

## Fix

Made the four wrappers check `resp.GetSuccess()` and surface `resp.GetMessage()` as an error, matching the existing pattern in the file. No handler changes were needed — `dashboard_handlers.go` already returns a non-nil error correctly wherever these wrappers are called, including in `processAlertAction`, so bulk-action `FailedCount` accounting now correctly reflects backend rejections.

## Testing

Added `internal/webui/client/backend_client_test.go` with a fake `AlertServiceClient` (embedding the real interface with only the four methods under test overridden) covering success and `Success: false` rejection paths for all four wrappers.

- `go build ./...`
- `go vet ./...`
- `go test ./internal/webui/client/...` — 8 passed

Closes #128

### DIFF
--- a/internal/webui/client/backend_client.go
+++ b/internal/webui/client/backend_client.go
@@ -346,8 +346,14 @@ func (c *BackendClient) AddAcknowledgment(sessionID, alertKey, reason string) er
 		Reason:    reason,
 	}
 
-	_, err := c.alertClient.AddAcknowledgment(ctx, req)
-	return err
+	resp, err := c.alertClient.AddAcknowledgment(ctx, req)
+	if err != nil {
+		return err
+	}
+	if !resp.GetSuccess() {
+		return fmt.Errorf("%s", resp.GetMessage())
+	}
+	return nil
 }
 
 // DeleteAcknowledgment removes acknowledgment from an alert
@@ -364,8 +370,14 @@ func (c *BackendClient) DeleteAcknowledgment(sessionID, alertKey string) error {
 		AlertKey:  alertKey,
 	}
 
-	_, err := c.alertClient.DeleteAcknowledgment(ctx, req)
-	return err
+	resp, err := c.alertClient.DeleteAcknowledgment(ctx, req)
+	if err != nil {
+		return err
+	}
+	if !resp.GetSuccess() {
+		return fmt.Errorf("%s", resp.GetMessage())
+	}
+	return nil
 }
 
 // GetAcknowledgments retrieves acknowledgments for an alert
@@ -491,8 +503,14 @@ func (c *BackendClient) addComment(sessionID, alertKey, content, kind string) er
 		Kind:      kind,
 	}
 
-	_, err := c.alertClient.AddComment(ctx, req)
-	return err
+	resp, err := c.alertClient.AddComment(ctx, req)
+	if err != nil {
+		return err
+	}
+	if !resp.GetSuccess() {
+		return fmt.Errorf("%s", resp.GetMessage())
+	}
+	return nil
 }
 
 // DeleteComment removes a comment from an alert
@@ -509,8 +527,14 @@ func (c *BackendClient) DeleteComment(sessionID, commentID string) error {
 		CommentId: commentID,
 	}
 
-	_, err := c.alertClient.DeleteComment(ctx, req)
-	return err
+	resp, err := c.alertClient.DeleteComment(ctx, req)
+	if err != nil {
+		return err
+	}
+	if !resp.GetSuccess() {
+		return fmt.Errorf("%s", resp.GetMessage())
+	}
+	return nil
 }
 
 // GetRecentActivity fetches recent cross-alert collaboration events for the activity feed.
--- a/internal/webui/client/backend_client_test.go
+++ b/internal/webui/client/backend_client_test.go
@@ -1,0 +1,94 @@
+package client
+
+import (
+	"context"
+	"testing"
+
+	alertpb "notificator/internal/backend/proto/alert"
+
+	"google.golang.org/grpc"
+)
+
+// fakeAlertClient embeds the real interface (nil) so it satisfies
+// alertpb.AlertServiceClient while only overriding the methods under test.
+type fakeAlertClient struct {
+	alertpb.AlertServiceClient
+	success bool
+	message string
+}
+
+func (f *fakeAlertClient) AddAcknowledgment(ctx context.Context, in *alertpb.AddAcknowledgmentRequest, opts ...grpc.CallOption) (*alertpb.AddAcknowledgmentResponse, error) {
+	return &alertpb.AddAcknowledgmentResponse{Success: f.success, Message: f.message}, nil
+}
+
+func (f *fakeAlertClient) DeleteAcknowledgment(ctx context.Context, in *alertpb.DeleteAcknowledgmentRequest, opts ...grpc.CallOption) (*alertpb.DeleteAcknowledgmentResponse, error) {
+	return &alertpb.DeleteAcknowledgmentResponse{Success: f.success, Message: f.message}, nil
+}
+
+func (f *fakeAlertClient) AddComment(ctx context.Context, in *alertpb.AddCommentRequest, opts ...grpc.CallOption) (*alertpb.AddCommentResponse, error) {
+	return &alertpb.AddCommentResponse{Success: f.success, Message: f.message}, nil
+}
+
+func (f *fakeAlertClient) DeleteComment(ctx context.Context, in *alertpb.DeleteCommentRequest, opts ...grpc.CallOption) (*alertpb.DeleteCommentResponse, error) {
+	return &alertpb.DeleteCommentResponse{Success: f.success, Message: f.message}, nil
+}
+
+func TestAddAcknowledgment_BackendRejection(t *testing.T) {
+	c := &BackendClient{alertClient: &fakeAlertClient{success: false, message: "invalid session"}}
+	err := c.AddAcknowledgment("session", "alert-key", "reason")
+	if err == nil || err.Error() != "invalid session" {
+		t.Fatalf("expected error %q, got %v", "invalid session", err)
+	}
+}
+
+func TestAddAcknowledgment_Success(t *testing.T) {
+	c := &BackendClient{alertClient: &fakeAlertClient{success: true}}
+	if err := c.AddAcknowledgment("session", "alert-key", "reason"); err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+}
+
+func TestDeleteAcknowledgment_BackendRejection(t *testing.T) {
+	c := &BackendClient{alertClient: &fakeAlertClient{success: false, message: "unauthorized"}}
+	err := c.DeleteAcknowledgment("session", "alert-key")
+	if err == nil || err.Error() != "unauthorized" {
+		t.Fatalf("expected error %q, got %v", "unauthorized", err)
+	}
+}
+
+func TestDeleteAcknowledgment_Success(t *testing.T) {
+	c := &BackendClient{alertClient: &fakeAlertClient{success: true}}
+	if err := c.DeleteAcknowledgment("session", "alert-key"); err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+}
+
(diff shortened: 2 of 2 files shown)


SoulKyu (commented):
@SoulKyu thanks for this fix — verified it end to end.

The four wrappers (`AddAcknowledgment`, `DeleteAcknowledgment`, `AddComment`, `DeleteComment` in `internal/webui/client/backend_client.go`) now correctly check `resp.GetSuccess()` and surface `resp.GetMessage()` as an error, exactly matching the existing `UpdateTimezone` convention in the same file. I confirmed all eight call sites in `dashboard_handlers.go` already propagate the returned error correctly (HTTP 500 with the message, or accumulated into `FailedCount` in the bulk-action path), so no handler changes were needed, as the PR description states.

I ran `go build ./...`, `go vet ./...`, and `go test ./internal/webui/client/...` locally — all clean, and the new `backend_client_test.go` covers both the success and `Success: false` rejection paths for all four RPCs using a nicely minimal fake client (embedding the real interface so only the methods under test need overriding).

Small, well-scoped fix for a real correctness bug (silent write failures reported as success). Nice work closing out #128.



SoulKyu:

SoulKyu (commented):
@SoulKyu re-verified this PR at the current head (8999360) since it was rebased after the prior review round.

The four wrappers in `internal/webui/client/backend_client.go` (`AddAcknowledgment`, `DeleteAcknowledgment`, `addComment` — used by both `AddComment` and `AddSystemComment` — and `DeleteComment`) now check `resp.GetSuccess()` and surface `resp.GetMessage()` as an error, matching the existing `UpdateTimezone` convention in the same file. Using `fmt.Errorf("%s", resp.GetMessage())` with a constant format string is also consistent with the repo's earlier vet-driven cleanup of non-constant `fmt.Errorf` calls.

I traced the backend side (`internal/backend/services/services.go`) to confirm `Success`/`Message` are populated on every rejection path (invalid session, empty required fields, DB write failure), and traced all call sites in `dashboard_handlers.go` (acknowledge/unacknowledge, add/delete comment, the bulk-action path) to confirm each already propagates the returned error correctly — no handler changes were needed, as the PR states. The proto getters (`GetSuccess`/`GetMessage`) are nil-receiver-safe, so the transport-error-but-nil-response edge case can't panic.

Ran `go build ./...`, `go vet ./internal/webui/client/...`, and `go test ./internal/webui/client/... -v` locally — all clean, 8 tests passing, covering both the success and rejection path for all four RPCs with a nicely minimal fake client.

Small, well-scoped, correctly targeted fix for a real silent-failure bug. Nice work closing out #128.



SoulKyu:
## 🧪 QA Report

Environment: `make test` (docker-compose: postgres + fake alertmanager + backend + webui), webui on :8081, real browser (Playwright/Chromium) for UI flows.

To exercise the `Success: false` path end to end I forced the backend to reject writes at the DB layer while auth kept working — a surgical `CHECK` constraint on `comments.content`/`acknowledgments.alert_key` and a `BEFORE DELETE` trigger. That reproduces the issue's "DB write failure" scenario without stubbing anything in the app.

|---|---|---|
| `AddComment`, `DeleteComment`, `AddAcknowledgment`, `DeleteAcknowledgment` return non-nil error carrying the backend message on `Success: false` | ✅ | All four driven live through HTTP. Messages surfaced verbatim: `Failed to create comment`, `Failed to delete comment or unauthorized`, `Failed to create acknowledgment`, `Failed to delete acknowledgment or unauthorized`. |
| Comment with invalid session → non-2xx, `CommentCount` not incremented | ✅ | Deleted the session row mid-flight → `401 {"error":"Invalid or expired session"}`, count stayed 3, 0 rows in `comments`. DB-rejection variant → `500 Failed to add comment: Failed to create comment`; count 2→3 on an accepted comment, then stayed 3 on the rejected one. |
| Rejected acknowledge → non-2xx, `IsAcknowledged` not set in cache | ✅ (see note) | `isAcknowledged` stayed `false`, `acknowledgedBy` empty, no row in `acknowledgments`. Browser: modal stays open and renders `1 of 1 actions failed: failed to store acknowledgment in backend: Failed to create acknowledgment` in red; row does not flip to acknowledged. **Note:** transport status is `200`, not non-2xx — the only ack path is `POST /dashboard/bulk-action`, which deliberately returns 200 and reports the verdict in the envelope (`dashboard_handlers.go:897`, from #69/PR #62). That predates this PR; the frontend gates on `result.success`, not on HTTP status, so the user-visible outcome the criterion is protecting is correct. Flagging this as a fail would push a regression of #62's contract. |
| Unit tests with a fake `AlertServiceClient` returning `Success: false` cover all four wrappers | ✅ | `go test ./internal/webui/client/...` → 8 passed. Mutation check: reverted the four wrappers to `_, err := …; return err` in a throwaway copy → all four `*_BackendRejection` tests fail (`expected error "…", got <nil>`). The tests are load-bearing, not vacuous. |
| Bulk action over accepted + rejected alerts reports the rejected ones in `FailedCount` | ✅ | 3 alerts, 1 rejected → `processedCount:2, failedCount:1`, `failures[0].target` = the rejected fingerprint. DB confirms exactly the 2 accepted acks landed. |

**Adversarial pass:**

- **Boundary & empty** — comment of exactly 1000 chars → 200 (accepted); 1001 chars → 400; whitespace-only → 400; missing `content` field → 400; empty `alertFingerprints: []` → 200 `processed:0 failed:0` (no divide-by-zero, no bogus summary). None of these reach the wrapper or produce a false success.
- **Malformed & hostile** — `{"content":12345}` (wrong type) → 400; truncated JSON → 400; nonexistent fingerprint on the comment endpoint → 404; nonexistent fingerprint inside a bulk → counted in `FailedCount` with `alert not found`. Format-verb payload `%s %d %n` as comment content → stored as literal text; the fix uses `fmt.Errorf("%s", resp.GetMessage())` rather than `fmt.Errorf(msg)`, so a backend-controlled message cannot become a format string. `go vet ./internal/webui/client/...` clean.
- **Error & unavailable path** — this is the core of the change and the main attack: forced the backend to answer `Success: false` with gRPC-OK on all four RPCs. In every case the UI showed the failure instead of a wiped/optimistic success. Specifically the comment modal keeps the user's text in the textarea, does not add the comment to the list, and does not bump the badge — before this fix that same flow took the *success* branch (cleared the textarea, logged "Comment added successfully"). Verified `resp` is read through nil-safe getters, so a nil response cannot panic.
- **Auth & scope** — unauthenticated POST → 401; session row deleted mid-session → 401 and no write; user2 attempting to unacknowledge an alert acknowledged by user1 → correctly reported as failed (`Failed to delete acknowledgment or unauthorized`) and user1's ack survives in the DB. Before this fix that cross-user unack reported success and silently reverted ~10s later.
- **Adjacent flows** — `resolve` with an audit comment the DB rejects still succeeds (`processed:1, failed:0`): `AddSystemComment` now returns an error but the handler still only logs it, so the "don't fail the action if the audit comment fails" contract is intact. Group actions (`processGroupAction`) aggregate through the same wrapper and correctly report `kind:"group"` failures. Full success path re-verified after removing all instrumentation: ack, comment, delete comment, unacknowledge all return 200 and land in the DB — the fix does not reject valid writes.
- **Persisted state** — all of the above ran against a database with pre-existing users, comments and acknowledgments created before the failure instrumentation was installed, on a persisted `postgres-data` volume. No schema migration in this PR.

Out of scope, noticed while testing (pre-existing, not caused by this PR): the dashboard fires `GET /api/v1/profile/timezone` and `GET /api/impersonate/status`, both 401 on every load; and the comment modal's failure branch reports only to `console.error` with no visible banner (the ack modal does render one).

**Verdict: PASS**

PR #134, opened 2026-07-27
fix(webui): Sentry personal tokens are sent to whatever host an alert annotation points at

### LINKED ISSUE #127: fix(webui): Sentry personal tokens are sent to whatever host an alert annotation points at
## Problem

The Sentry integration derives the API host from alert data, then sends the viewing user's personal Sentry token to it.

- `internal/webui/services/sentry_service.go:63-67` — the URL comes from the alert's `sentry` annotation, falling back to the `sentry` label. That is Alertmanager payload, i.e. attacker-influenceable data, not configuration.
- `internal/webui/services/sentry_service.go:225-231` — `parseSentryURL` builds `BaseURL` from the scheme and host of that URL. The only validation is a path regex requiring `/organizations/<org>/projects/<project>`.
- `internal/webui/services/sentry_service.go:87` then `:103`, `:335`, `:494`, `:555`, `:710`, `:741` — every request is built against that host, and each sets `req.Header.Set("Authorization", "Bearer "+token)` (`:277`, `:346`, `:504`, `:568`, `:688`, `:720`, `:748`).
- `internal/webui/services/sentry_service.go:179-218` — the token is the *viewing user's* personal Sentry token, fetched from the backend, or the admin-configured `GlobalToken` as fallback.

`config.SentryConfig.BaseURL` (`config/config.go:145-149`) — the configured Sentry instance — is only consulted when saving or testing a token (`internal/webui/handlers/sentry_handlers.go:103-120`, `:162-172`, `:260-265`). The data path never compares the annotation host against it.

Failure path: anyone able to POST an alert to a monitored Alertmanager (`/api/v2/alerts` is commonly reachable from inside the cluster), or able to add an annotation to a Prometheus rule, sets `sentry: https://attacker.example/organizations/o/projects/p`. The alert appears on the dashboard; the first responder opens it, the browser calls `GET /dashboard/sentry/:fingerprint` (`internal/webui/router.go:296`), and the WebUI sends that responder's Sentry personal token in an `Authorization: Bearer` header to `attacker.example`. Repeat per user. The same primitive also lets an alert drive authenticated GET requests to internal hosts the WebUI can reach.

## Impact

Credential exfiltration through the alert pipeline: a single crafted alert harvests the Sentry personal token of every user who opens it, plus the admin `global_token` for users who have not configured their own. Those tokens carry the user's Sentry permissions on the real instance. Alert content is treated as trusted configuration here, which it is not — the whole point of the tool is to display alerts from systems the WebUI does not control.

## Proposed approach

Constrain the host before any request carrying a token is made, in `internal/webui/services/sentry_service.go`:

- After `parseSentryURL`, reject the alert's URL unless its scheme+host matches the configured `config.Sentry.BaseURL` (and/or the per-user `SentryBaseURL` already stored with the token in `internal/backend/database/sentry_db.go`). On mismatch, return `SentryData{HasSentryLabel: true, SentryURL: …, Error: "Sentry host not allowed"}` and make no outbound request — the existing error field already renders in the modal.
- Optionally allow an explicit allowlist of additional hosts in `SentryConfig` for multi-instance setups; default to the single configured instance.
- Require `https` (or at least reject non-http(s) schemes) so a `file:`/other scheme cannot reach the transport.

## Acceptance criteria

- An alert whose `sentry` annotation points at a host other than the configured Sentry instance produces no outbound HTTP request at all (verifiable with a test using an `httptest` server that must not be hit), and the modal shows an explicit "host not allowed" error.
- An alert pointing at the configured instance keeps working exactly as today.
- The check covers both the annotation and the label fallback path.
- A unit test asserts that no `Authorization` header ever leaves for a non-allowlisted host, including on the project-ID resolution call, which is the first request made.

SoulKyu:
## Summary

- The Sentry integration derived the API host from the alert's `sentry` annotation/label (attacker-influenceable Alertmanager payload) and sent it the viewing user's personal Sentry token — or the admin `global_token` — in an `Authorization: Bearer` header. A crafted alert could exfiltrate every viewer's Sentry token to an attacker-controlled host, or drive authenticated GETs against internal hosts.
- `parseSentryURL` (`internal/webui/services/sentry_service.go`) now rejects any URL whose scheme+host doesn't match the configured `config.Sentry.BaseURL`, or an entry in a new optional `Sentry.AllowedHosts` list for multi-instance setups, **before** any outbound request is built (including the project-ID resolution call, which was the first request made). Only `http(s)` schemes are accepted, so a `file:`/other scheme can't reach the transport.
- On rejection, `GetSentryDataForAlert` returns `SentryData{HasSentryLabel: true, SentryURL: ..., Error: "Sentry host not allowed"}` and makes no request — the existing error field already renders in the modal.
- Both the annotation and the label fallback path go through the same check.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./config/... ./internal/webui/...`
- [x] `go test ./internal/webui/services/... ./config/... ./internal/webui/handlers/...`
- New `internal/webui/services/sentry_service_test.go`:
  - alert pointing at a non-allowlisted host: `httptest` server asserts zero requests received and no `Authorization` header, `SentryData.Error == "Sentry host not allowed"` — covers both annotation and label fallback
  - alert pointing at the configured instance: still resolves the project ID and fetches data, confirming the previously-working path is unaffected
  - `isAllowedSentryHost` table test for the configured host, an allowlisted host, an attacker host, a non-http(s) scheme, and http-on-configured-host

Closes #127


config/config.go
```diff
@@ -143,9 +143,10 @@ type WebUIConfig struct {
 }
 
 type SentryConfig struct {
-	Enabled     bool   `json:"enabled"`
-	BaseURL     string `json:"base_url"`     // Default Sentry instance URL (e.g., "https://sentry.io")
-	GlobalToken string `json:"global_token"` // Admin-configured fallback token
+	Enabled      bool     `json:"enabled"`
+	BaseURL      string   `json:"base_url"`               // Default Sentry instance URL (e.g., "https://sentry.io")
+	GlobalToken  string   `json:"global_token"`            // Admin-configured fallback token
+	AllowedHosts []string `json:"allowed_hosts,omitempty"` // Additional Sentry hosts (scheme+host) allowed to receive tokens, for multi-instance setups
```
SoulKyu: **Problem:** The comment says `AllowedHosts` entries are `scheme+host` (e.g. implying something like `https://sentry-eu.example.com`), but `isAllowedSentryHost` in `internal/webui/services/sentry_service.go` (lines 297-298) compares each entry with `strings.EqualFold(strings.TrimSpace(host), parsedURL.Host)`, where `parsedURL.Host` is scheme-stripped (just `host[:port]`, e.g. `sentry-eu.example.com`). An admin who follows this comment literally and sets `allowed_hosts: ["https://sentry-eu.example.com"]` will get an entry that can never match any parsed alert URL — the allowlist entry silently never takes effect (fails closed, so not a security hole, but it defeats the multi-instance support this PR advertises in its description).

There's a second related gotcha for the env-var path: `viper.BindEnv("sentry.allowed_hosts", "NOTIFICATOR_SENTRY_ALLOWED_HOSTS")` feeds `viper.GetStringSlice`, which for a plain env string value uses `cast.ToStringSliceE`'s `case string: return strings.Fields(v), nil` — i.e. it splits on **whitespace**, not commas. `NOTIFICATOR_SENTRY_ALLOWED_HOSTS="host1.com,host2.com"` becomes a single-element slice `["host1.com,host2.com"]`, not two hosts.

**Suggested fix:** Change the comment to something like `// Additional Sentry hosts (host[:port], no scheme, e.g. "sentry-eu.example.com") allowed to receive tokens; for env config use space-separated values (NOTIFICATOR_SENTRY_ALLOWED_HOSTS="a.example.com b.example.com")`, and/or normalize entries in `isAllowedSentryHost` by parsing them with `url.Parse` (stripping any scheme) so both `host.com` and `https://host.com` styles work.


SoulKyu (commented):
@SoulKyu Thanks for closing the loop on #127 — this is a solid, tightly-scoped fix. `isAllowedSentryHost` is checked before any outbound request is built (including the project-ID resolution call that used to go out first), it's exercised for both the annotation and label fallback paths, `parseSentryURL` is the single choke point every downstream fetch (`fetchProjectID`, `fetchProjectStats`, `fetchProjectDetails`, `fetchIssues`) relies on for `BaseURL`, and the new tests assert zero requests / no leaked `Authorization` header against a disallowed host with an `httptest` server, which is exactly the right way to prove this. `go build`/`go vet`/`go test` all pass locally. One non-blocking config-documentation nit inline on `AllowedHosts`.



config/config.go
SoulKyu: @SoulKyu This has been fixed in deca05a.

Reworded the AllowedHosts comment in config/config.go to document the actual host[:port] format and the env var's space-separated syntax, and updated isAllowedSentryHost in sentry_service.go to strip a scheme prefix from AllowedHosts entries before comparing, so both 'host.com' and 'https://host.com' styles now match.




SoulKyu:
**Looper fixer round complete** — deca05a

- ✅ Review comment on `config/config.go` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/134#discussion_r3655266090)
  - Reworded the AllowedHosts comment in config/config.go to document the actual host[:port] format and the env var's space-separated syntax, and updated isAllowedSentryHost in sentry_service.go to strip a scheme prefix from AllowedHosts entries before comparing, so both 'host.com' and 'https://host.com' styles now match.


SoulKyu:
🔀 Rebase (conflicts resolved): rebased onto `origin/main` (d42d5da).

**Why conflicts:** `main` had meanwhile landed 0d445f9 *"fix(webui): restrict Sentry requests to the configured instance host"* — an independent fix for the same vulnerability. Both branches touched `parseSentryURL` and both added a `sentry_service_test.go`.

**Resolution — `internal/webui/services/sentry_service.go`:**
- Kept this PR's `isAllowedSentryHost()` helper (superset: it adds the `Sentry.AllowedHosts` multi-instance allowlist, which `main` does not have).
- Kept `main`'s stricter scheme rule inside it: the URL scheme must match the configured `BaseURL` scheme, so an alert can no longer downgrade an `https` instance to plaintext `http`. Previously this PR compared host only. Scheme-prefixed `AllowedHosts` entries are matched on scheme+host; bare entries inherit the configured scheme.
- Removed the duplicate `errSentryHostNotAllowed` declaration (each side had added one).
- Error surfaced to the modal stays `"Sentry host not allowed"`; the wrapped error keeps `main`'s detailed `%w: scheme://host does not match the configured instance ...` context for logs.

**Resolution — `internal/webui/services/sentry_service_test.go` (add/add):**
- Kept `main`'s `TestSentryDataRefusesForeignHosts` / `TestSentryDataAllowsConfiguredHost` — they already covered this PR's foreign-host, label-fallback and positive-control cases, so the near-identical duplicates from this branch were dropped rather than kept twice.
- Kept this PR's `TestIsAllowedSentryHost` (the only unique coverage: the `AllowedHosts` allowlist), with two cases updated to the merged behaviour: scheme downgrade on the configured host and on an allowlisted host now both expect rejection, matching `main`'s guarantee.

**Net behaviour change vs. this PR before rebase:** `http://` on an `https://`-configured Sentry host is now refused instead of allowed. No other functional change; `config.Sentry.AllowedHosts` rebased cleanly.

**Verification:** `go build ./...` ✅ · `go vet ./...` ✅ · `go test ./...` → 193 passed / 27 packages ✅
No `*.templ` files touched, so no `make webui-templates` regeneration was needed.

internal/webui/services/sentry_service.go
```diff
@@ -288,6 +283,43 @@ func (s *SentryService) parseSentryURL(sentryURL string) (*SentryProjectInfo, er
...
+			parsedEntry, err := url.Parse(entry)
+			if err != nil || !strings.EqualFold(parsedEntry.Scheme, parsedURL.Scheme) {
+				continue
+			}
+			entry = parsedEntry.Host
+		}
+		if strings.EqualFold(entry, parsedURL.Host) {
+			return true
+		}
+	}
```
SoulKyu, line 330: **Problem:** `isAllowedSentryHost` rejects any URL whose scheme differs from `s.config.BaseURL`'s scheme *before* it ever consults `AllowedHosts`: line 297's `!strings.EqualFold(configured.Scheme, parsedURL.Scheme)` returns `false` immediately when the schemes differ. The `AllowedHosts` loop below (lines 306-318) tries to honor a scheme prefix on each entry (`parsedEntry.Scheme` vs `parsedURL.Scheme`), but that code path is unreachable whenever an entry's scheme differs from the primary `BaseURL`'s scheme, because the function already returned `false` on line 298.

**Why it matters:** Both this PR and the `config/config.go` field comment (`AllowedHosts ... host[:port], scheme optional`) advertise that `AllowedHosts` entries may carry their own scheme — e.g. an admin adding a legacy internal Sentry mirror served over plain `http` while the primary `BaseURL` is `https`. That configuration silently never works: every alert pointing at the allowlisted `http` host gets rejected with "Sentry host not allowed", even though the admin explicitly listed it. This fails closed (not an exploitable hole), but it means the multi-instance escape hatch this commit just reworked doesn't do what it's documented to do for any allowlisted host that isn't on the same scheme as `BaseURL`.

**Evidence:** Verified locally by constructing `SentryConfig{BaseURL: "https://sentry.example.com", AllowedHosts: []string{"http://sentry-legacy.example.com"}}` and calling `parseSentryURL("http://sentry-legacy.example.com/organizations/o/projects/p")` — it returns `sentry host not allowed: http://sentry-legacy.example.com does not match the configured instance https://sentry.example.com`, never reaching the `AllowedHosts` loop at all. None of the cases in the new `TestIsAllowedSentryHost` table (sentry_service_test.go) use an `AllowedHosts` entry whose scheme intentionally differs from `BaseURL`'s, so this gap isn't caught by the added tests.

**Suggested fix:** Don't gate the whole function on `configured.Scheme == parsedURL.Scheme` up front. Require that match only for the direct `BaseURL` comparison, and let each `AllowedHosts` entry independently determine which scheme it accepts — a bare entry (no `://`) defaults to `configured.Scheme`, while an entry with an explicit scheme prefix should be allowed to differ. Add a `TestIsAllowedSentryHost` case where an `AllowedHosts` entry's scheme intentionally differs from `BaseURL`'s to lock in the intended behavior.


SoulKyu (commented):
@SoulKyu Reviewed the follow-up commit (d19bbbb) that reworks `isAllowedSentryHost` in response to the AllowedHosts scheme-prefix nit from the previous round. The core fix for the token-exfiltration issue still holds: `parseSentryURL` remains the single choke point every downstream request (`fetchProjectID`, `fetchProjectStats`, `fetchProjectDetails`, `fetchIssues`) depends on, both the annotation and label fallback paths go through it, and `go build`/`go vet`/`go test ./internal/webui/services/... ./config/...` all pass locally with the new tests correctly proving zero requests reach a disallowed host. One non-blocking correctness gap remains in the reworked AllowedHosts scheme handling — left as an inline comment with a concrete repro. Thanks for the quick follow-up here, this is good work overall.



internal/webui/services/sentry_service.go
SoulKyu: @SoulKyu This has been fixed in a3a25b3.

Reworked isAllowedSentryHost in internal/webui/services/sentry_service.go so only the direct BaseURL comparison requires an exact scheme+host match; each AllowedHosts entry now independently resolves its own scheme (bare entries default to the configured scheme, scheme-prefixed entries can differ). Added a table-test case in sentry_service_test.go for an http-only AllowedHosts entry alongside an https BaseURL.




SoulKyu:
**Looper fixer round complete** — a3a25b3

- ✅ Review comment on `internal/webui/services/sentry_service.go:324` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/134#discussion_r3682511126)
  - Reworked isAllowedSentryHost in internal/webui/services/sentry_service.go so only the direct BaseURL comparison requires an exact scheme+host match; each AllowedHosts entry now independently resolves its own scheme (bare entries default to the configured scheme, scheme-prefixed entries can differ). Added a table-test case in sentry_service_test.go for an http-only AllowedHosts entry alongside an https BaseURL.


SoulKyu:
## 🧪 QA Report

Env: images built from `deca05a` (webui/backend/fake-alertmanager), docker-compose with `NOTIFICATOR_SENTRY_ENABLED=true`, `base_url=http://sentry-good:8000`, `allowed_hosts="other.example.com sentry-extra:8000"`, global token `qa-GLOBAL-token-SECRET` + a saved personal token `qa-PERSONAL-token-SECRET`. Three instrumented fake Sentry instances (`sentry-good` = configured, `sentry-extra` = allowlisted, `sentry-evil` = attacker) log every request path + `Authorization` header. 11 alerts injected through the fake Alertmanager `POST /api/v2/alerts`.

|---|---|---|
| Alert whose `sentry` annotation points at a non-configured host produces **no outbound HTTP request at all** | ✅ | `GET /api/v1/dashboard/sentry/<fp>` for 7 hostile URL variants → `error: "Sentry host not allowed"`, and `docker logs qa-sentry-evil` shows **0 requests** (per-request hit diffing on all three fakes: `new_hits=0`) |
| …and **the modal shows an explicit "host not allowed" error** | ❌ | Real browser (Playwright), user *has* a valid personal token: Sentry tab of `QAEvilAnnotation` renders **"Unable to load Sentry data / Sentry token not configured / Configure Sentry Token"** — the `error` field is never displayed. `loadSentryData` (`internal/webui/templates/scripts/dashboard_modal.templ:494-514`) branches only on `auth_status.has_api_token` and never reads `result.error`; the rejection path (`internal/webui/services/sentry_service.go:83-93`) returns early with an **unpopulated `AuthStatus`**, so `has_api_token=false` → the responder is told their token is missing and invited to re-enter it |
| Alert pointing at the configured instance keeps working exactly as today | ✅ | `QAGoodHost` → 6 API calls reach `sentry-good` (`/projects/qa/proj/`, `issues/`, `sessions/`, `stats/`, `releases/`) with `Authorization: Bearer qa-PERSONAL-token-SECRET`; modal renders project + "QAError: fake sentry issue". Same with global-token fallback. Allowlisted `sentry-extra` works too |
| The check covers both the annotation and the label fallback path | ✅ | `QAEvilLabel` (no annotation, `sentry` **label** = attacker host) → `"Sentry host not allowed"`, 0 requests to any fake |
| Unit test asserts no `Authorization` header leaves for a non-allowlisted host, incl. the project-ID resolution call | ✅ | `go test ./internal/webui/services/... ./config/...` → 78 pass. Non-vacuous: disabling the guard in a scratch copy (`if false && !s.isAllowedSentryHost(...)`) makes `TestGetSentryDataForAlert_RejectsUnallowedHost` **and** `_LabelFallback` fail |

**Adversarial pass:**

- **Malformed / hostile hosts** (all rejected, 0 outbound requests, no token): userinfo confusion `http://sentry-good:8000@sentry-evil:8000/organizations/qa/projects/proj/`; suffix confusion `http://sentry-good:8000.sentry-evil:8000/...`; same-host different port `http://sentry-good:9999/...`; `file:///organizations/qa/projects/proj/`; protocol-relative `//sentry-evil:8000/...`. Uppercase configured host `http://SENTRY-GOOD:8000/...` is correctly still **allowed** (case-insensitive compare).
- **Boundary / empty**: `sentry: ""` annotation → no request, error `"Sentry host not allowed"`. Minor: an empty/garbage URL now reports *host not allowed* where it used to say *Invalid Sentry URL format* — cosmetic only.
- **Error / unavailable path**: stopped the allowed fake instance → modal surfaces `Failed to resolve project ID: … dial tcp: lookup sentry-good … server misbehaving` (no silent empty "success", no stale data).
- **Redirect from the allowed instance to the attacker host** (configured Sentry answers `302 → http://sentry-evil:8000`): the 6 GETs are followed to the attacker but Go strips the credential — attacker logs show `AUTH=<none>` for every request. No token leaves; criterion holds.
- **Auth & scope**: unauthenticated `GET /api/v1/dashboard/sentry/<fp>` → `401 Authentication required`; non-existent fingerprint → `404 Alert not found`.
- **Persisted / stored state**: saved a **personal token whose own `base_url` is the attacker host** (`/dashboard/sentry-token` → `sentry-config` confirms `base_url: http://sentry-evil:8000`), then re-ran the attacker alert: still refused, still 0 requests — the per-user stored base URL does **not** widen the allowlist. Reused the pre-existing postgres volume/user (PR is config-only, no migration).
- **Config formats claimed by the head commit**: `NOTIFICATOR_SENTRY_ALLOWED_HOSTS="other.example.com sentry-extra:8000"` — space-separated env parsing works, and both entry styles (`host:port` and `https://host:port`) match. **Caveat:** a schemeless `sentry.base_url` (e.g. `sentry-good:8000`) makes the *configured* instance itself rejected — `url.Parse` yields an empty `Host`, so every Sentry panel dies with "Sentry host not allowed" (it worked before this PR, which never consulted `base_url` on the data path). Not counted as a criterion failure since `base_url` is documented/defaulted with a scheme, but `AllowedHosts` documents "scheme optional" right next to it, so it is an easy admin trap.

The backend hardening is solid — every attacker-influenceable path I could construct was blocked before any request was built. The failure is the user-facing half of criterion 1: the responder sees a wrong, misleading message. Suggested fix: populate `AuthStatus` on the rejection return and/or have `loadSentryData` prefer `result.error` when present.

**Verdict: FAIL**

SoulKyu:
## 🧪 QA Report

Env: images rebuilt from `a3a25b3` (webui + backend), docker-compose with `NOTIFICATOR_SENTRY_ENABLED=true`, `base_url=http://sentry-good:8000`, `allowed_hosts="other.example.com sentry-extra:8000"`, global token `qa-GLOBAL-token-SECRET` plus a saved personal token `qa-PERSONAL-token-SECRET`. Three instrumented fake Sentry instances (`sentry-good` = configured, `sentry-extra` = allowlisted, `sentry-evil` = attacker) log every request path + `Authorization` header. 16 alerts injected through the fake Alertmanager `POST /api/v2/alerts`; each probe diffs the per-container hit count before/after the request.

|---|---|---|
| Alert whose `sentry` annotation points at a non-configured host produces **no outbound HTTP request at all** | ✅ | 11 hostile URL variants → `GET /api/v1/dashboard/sentry/<fp>` returns `error: "Sentry host not allowed"` with `new_hits=0` on **all three** fakes (`qa-sentry-evil` logs zero requests, zero `Authorization` headers). Verified with global-token-only and with a personal token saved |
| …and **the modal shows an explicit "host not allowed" error** | ❌ | Real browser (Playwright), user *has* a valid personal token (`/dashboard/sentry-config` → `has_token: true`): the Sentry tab of `QAEvilAnnotation` renders **"Unable to load Sentry data / Sentry token not configured / Configure Sentry Token"**. The rejection branch (`internal/webui/services/sentry_service.go:90-94`) returns an **unpopulated `AuthStatus`**, and `loadSentryData` (`internal/webui/templates/scripts/dashboard_modal.templ:509-524`) branches only on `auth_status.has_api_token`, never reading `result.error` → the responder is told their token is missing and invited to re-enter it. The API carries the right message; nothing user-visible ever shows it |
| Alert pointing at the configured instance keeps working exactly as today | ✅ | `QAGoodHost` → 12 API calls reach `sentry-good` (`/projects/qa/proj/`, `issues/`, `sessions/`, `stats/`, `releases/`) with `Authorization: Bearer qa-PERSONAL-token-SECRET`; modal renders project, release 1.2.3, Apdex, and "QAError: fake sentry issue". Same via global-token fallback. Allowlisted `sentry-extra` works identically |
| The check covers both the annotation and the label fallback path | ✅ | `QAEvilLabel` (no annotation, `sentry` **label** = attacker host) → `"Sentry host not allowed"`, 0 requests to any fake |
| Unit test asserts no `Authorization` header leaves for a non-allowlisted host, incl. the project-ID resolution call | ✅ | `go test ./internal/webui/services/... ./config/...` → pass. `TestSentryDataRefusesForeignHosts` asserts `attackerHits == 0` (stronger than header-only). Non-vacuous: with the guard neutered in a scratch copy (`if false && !s.isAllowedSentryHost(...)`), 4 subtests of `TestSentryDataRefusesForeignHosts` + 5 of `TestIsAllowedSentryHost` fail |

**Adversarial pass:**

- **Malformed / hostile hosts** — all rejected, `new_hits=0` everywhere, no token: userinfo confusion `http://sentry-good:8000@sentry-evil:8000/...`; suffix confusion `http://sentry-good:8000.sentry-evil:8000/...`; same host other port `http://sentry-good:9999/...`; `file:///organizations/qa/projects/proj/`; protocol-relative `//sentry-evil:8000/...`; uppercase attacker `http://SENTRY-EVIL:8000/...`; allowlisted host without its port `http://sentry-extra/...`. Uppercase *configured* host `http://SENTRY-GOOD:8000/...` is correctly still allowed (case-insensitive compare).
- **Scheme confusion (the point of this head commit)** — `https://sentry-good:8000/...` (upgrade on the configured host) → rejected. `https://sentry-extra:8000/...` with the bare entry `sentry-extra:8000` → rejected, i.e. a bare entry really does inherit the configured scheme. Re-ran the stack with `NOTIFICATOR_SENTRY_ALLOWED_HOSTS="https://sentry-extra:8000"`: the same https alert is now **allowed** (error becomes `tls: first record does not look like a TLS handshake` — an outbound attempt to `https://sentry-extra:8000`), while `http://sentry-extra:8000/...` flips to `"Sentry host not allowed"`. Per-entry scheme behaves exactly as the commit documents.
- **Boundary / empty** — `sentry: ""` annotation → no request, error `"Sentry host not allowed"` (cosmetic: this used to read *Invalid Sentry URL format*).
- **Error / unavailable path** — allowlisted-but-unreachable `other.example.com`: API returns `Failed to resolve project ID: … dial tcp: lookup other.example.com … no such host`, but the modal shows the neutral placeholder **"Sentry Integration / Data will be loaded automatically when you view this tab."** for 12 s+ — the failure is swallowed. Same root cause as the ❌ above (the modal never renders `result.error`); pre-existing on `main`, so not counted as a separate criterion failure.
- **Redirect from the allowed instance to the attacker host** — reconfigured `sentry-good` to answer `302 → http://sentry-evil:8000`: the 12 GETs are followed to the attacker but Go strips the credential — attacker logs show `AUTH=<none>` on every request. No token leaves.
- **Auth & scope** — unauthenticated `GET /api/v1/dashboard/sentry/<fp>` → `401 Authentication required`; unknown fingerprint → `404 Alert not found`; `%20` → 400; `..%2F..%2Fetc` → 404.
- **Persisted / stored state** — saved a personal token whose own `base_url` is the **attacker** host (`sentry-config` confirms `base_url: http://sentry-evil:8000`), then re-ran every hostile alert: still refused, still 0 requests — the per-user stored base URL does not widen the allowlist.
- **Adjacent flows** — the good-host and allowlisted-host modals still render the full Sentry panel (issues, stats, release) after all of the above; the alert list, search filter and the other 7 modal tabs are unaffected.
- **Config trap (not a criterion failure)** — a schemeless `sentry.base_url` (e.g. `sentry-good:8000`) makes `url.Parse` yield an empty `Host`, so `isAllowedSentryHost` returns false for *everything*, including the configured instance: verified live, `QAGoodHost` → `"Sentry host not allowed"`. Already true on `main`, but `AllowedHosts` documents "scheme optional" right next to `BaseURL`, which invites the mistake.
- Unrelated pre-existing crash noticed while setting up: `POST /api/v1/auth/login` with wrong credentials panics on `resp.User.Id` (`internal/webui/client/backend_client.go:151`) and returns 500 instead of "invalid credentials".

The backend hardening is solid — every attacker-influenceable path I could construct was blocked before any request was built, and the new per-entry-scheme allowlist behaves as documented. The failure is the user-facing half of criterion 1: the responder sees a wrong, misleading message ("your token is missing") for what is actually a rejected host. Suggested fix: populate `AuthStatus` on the rejection return, and/or have `loadSentryData` prefer `result.error` when present.

**Verdict: FAIL**

SoulKyu:

SoulKyu (commented):
@SoulKyu Reviewed the follow-up commit (76e2813) that reworks `isAllowedSentryHost` in response to the AllowedHosts scheme-independence gap flagged in the previous round.

What I verified:
- `isAllowedSentryHost` (internal/webui/services/sentry_service.go) now treats the configured `BaseURL` match and each `AllowedHosts` entry independently: the configured host still requires an exact scheme+host match (no https→http downgrade), while a scheme-prefixed `AllowedHosts` entry can legitimately use a different scheme than `BaseURL` (e.g. an http-only legacy mirror) and a bare entry still inherits the configured scheme. This matches the new test cases (`allowlisted host with scheme differing from BaseURL`, `scheme upgrade on scheme-prefixed allowlisted host`) and closes the gap from the prior round.
- `parseSentryURL` is still the single choke point every downstream request (`fetchProjectID`, `fetchProjectStats`, `fetchProjectDetails`, `fetchIssues`) depends on for its base URL, and it builds that base URL from the already-validated `parsedURL.Scheme`/`Host` rather than re-parsing the raw string, so there's no parser-differential gap between validation and the outbound request.
- `config.SentryConfig.AllowedHosts` is wired through `LoadConfigWithViper` and `NOTIFICATOR_SENTRY_ALLOWED_HOSTS` (space-separated, confirmed viper/cast splits on whitespace via `strings.Fields`).
- `go build ./...`, `go test ./internal/webui/services/... ./config/...` all pass locally, and the table-driven `TestIsAllowedSentryHost` plus the existing `TestSentryDataRefusesForeignHosts`/`TestSentryDataAllowsConfiguredHost` cover the configured host, allowlisted bare/scheme-prefixed hosts, attacker hosts, non-http(s) schemes, and scheme downgrade/upgrade attempts.

No new issues found in this round — nice, tight fix on a real token-exfiltration bug. Thanks for iterating through the review feedback so quickly.



SoulKyu:

SoulKyu (commented):
@SoulKyu Thanks for rebasing this onto the current `main` — the Sentry host-allowlist fix itself is unchanged from the last round (the `config/config.go`, `internal/webui/services/sentry_service.go` and `sentry_service_test.go` trees are byte-identical to commit `76e2813`, which was already reviewed clean; only the commit hash moved because of the rebase over the newer `main` history, e.g. #139).

I re-verified the fix from scratch against this head rather than trusting the prior review:
- `isAllowedSentryHost` is the single gate every alert-supplied Sentry URL passes through before any outbound request (including the project-ID resolution call) is built, for both the `sentry` annotation and label fallback paths.
- The configured `BaseURL` requires an exact scheme+host match (no https→http downgrade), while each `AllowedHosts` entry is evaluated independently — a bare entry inherits the configured scheme, a scheme-prefixed entry (e.g. an http-only legacy mirror) can legitimately differ. I checked this against edge cases not explicitly in the test table (empty/whitespace-only entries, malformed `scheme://` entries, port-only differences) and didn't find a bypass — non-http(s) schemes and empty hosts fail closed before any host comparison happens.
- The actual outbound request base URL is built from the already-validated `parsedURL.Scheme`/`Host`, not re-derived from `config.BaseURL`, so allowlisted secondary hosts genuinely receive their own traffic rather than silently being redirected — and there's no parse differential between the validation step and the request-building step.
- `go build ./...`, `go vet ./config/... ./internal/webui/...`, and `go test ./internal/webui/services/... ./config/...` all pass locally, and the `httptest`-based tests confirm zero requests reach a disallowed host with no leaked `Authorization` header.

No new issues found in this round. Solid, well-tested close on the token-exfiltration report in #127 — appreciate the follow-through across the review rounds.



SoulKyu:
## 🧪 QA Report

Live env: `docker compose up` with the PR's `webui` image, Sentry enabled, `NOTIFICATOR_SENTRY_BASE_URL=http://sentry-legit:9000`, `NOTIFICATOR_SENTRY_ALLOWED_HOSTS="sentry-eu:9000 http://sentry-apac:9000"`, global token `QA-GLOBAL-TOKEN-MUST-NOT-LEAK` + a saved **personal** token `QA-PERSONAL-TOKEN-MUST-NOT-LEAK`. Two request-logging probe servers on the compose network (`sentry-legit`, `attacker`) record every path + `Authorization` header. 17 crafted alerts POSTed to the fake Alertmanager, each opened through `GET /api/v1/dashboard/sentry/:fingerprint` and through the real modal in Chromium (Playwright).

|---|---|---|
| Alert `sentry` annotation pointing at a foreign host → **no outbound HTTP request at all** | ✅ | `QA-attacker-annotation` (`http://attacker:9000/organizations/org/projects/proj/?project=39`) → API returns `error: "Sentry host not allowed"`; `attacker` probe log stayed **empty** (0 requests) while `sentry-legit` logged 36. Same for the pre-existing seeded alert `QuerylyApplicationError` (`https://sentry.company.net/...`) — blocked, no request. |
| …**and the modal shows an explicit "host not allowed" error** | ❌ | Opened `QA-attacker-annotation` in Chromium with a **valid personal token saved**. Sentry tab renders: *"Unable to load Sentry data / **Sentry token not configured** / [Configure Sentry Token]"*. The JSON `error` field is never read: `internal/webui/templates/scripts/dashboard_modal.templ:509-524` branches only on `result.auth_status.has_api_token`, and the refusal path returns a zero-valued `auth_status` → `has_api_token=false`. The PR body's claim *"the existing error field already renders in the modal"* is not true. Result is actively misleading: the operator is told their token is missing (it is not) and is pushed to re-enter a Sentry token. |
| An alert pointing at the configured instance keeps working exactly as today | ✅ | `QA-legit` (`?project=39`) and `QA-legit-noid` (slug only → forces the project-ID resolution call) both return `project_info`, `issues`, `stats`; probe log shows the full request chain (`/api/0/projects/org/proj/`, `/issues/`, `/sessions/`, `/releases/`) with `Bearer QA-PERSONAL-TOKEN…`, all with `Host: sentry-legit:9000`. Modal renders the Sentry panel ("View in Sentry"). |
| The check covers both the annotation and the label fallback path | ✅ | `QA-attacker-label` (no annotation, `sentry` **label** = `http://attacker:9000/...`) → `error: "Sentry host not allowed"`, 0 requests to the attacker probe. |
| Unit test asserts no `Authorization` ever leaves for a non-allowlisted host, incl. the project-ID resolution call | ✅ | `go test ./internal/webui/services/ ./config/` → 84 pass; `TestSentryDataRefusesForeignHosts` asserts `attackerHits == 0` (strictly stronger than header inspection) and its `label fallback` case has no `?project=`, so it goes through `fetchProjectID` — the first request — with `GlobalToken` set, so the assertion is not vacuous. |

**Adversarial pass** (all run against the live app; ✅ = held):

- **Host/URL confusion** — `http://sentry-legit:9000@attacker:9000/...` (userinfo) ✅ blocked; `http://sentry-legit.attacker:9000/...` (allowed host as a prefix of the attacker's) ✅ blocked; `//attacker:9000/...` (scheme-relative, empty scheme) ✅ blocked; `HTTP://SENTRY-LEGIT:9000/...` ✅ correctly **allowed** (case-insensitive, request sent with `Host: SENTRY-LEGIT:9000`); `http://sentry-legit.:9000` (trailing dot) blocked — fails closed.
- **Scheme** — `https://sentry-legit:9000/...` against an `http://` configured instance ✅ blocked (no silent upgrade/downgrade); `file:///organizations/o/projects/p` ✅ blocked before the transport.
- **Boundary / empty / malformed** — empty `sentry: ""` annotation ✅ blocked (`Sentry host not allowed`); `http://sentry-legit:9000/nothing-here` → `Invalid Sentry URL format`, no request; path traversal in the org segment `organizations/..%2f..%2fattacker/projects/proj/` ✅ rejected by the path regex, no request; non-existent fingerprint → 404.
- **AllowedHosts escape hatch (new config key)** — space-separated `NOTIFICATOR_SENTRY_ALLOWED_HOSTS` really is parsed as a list by viper: bare entry `sentry-eu:9000` **and** scheme-prefixed `http://sentry-apac:9000` both resolve and fetch successfully (probe log shows `Host: sentry-eu:9000` / `Host: sentry-apac:9000`), while `attacker:9000` stays refused. Foot-gun worth a doc line: sibling vars in this repo (`NOTIFICATOR_ADMIN_USERS`) are **comma**-separated — a comma-separated value here becomes one bogus entry (fails closed, no security impact).
- **Error / unavailable path** — the refusal is not rendered as an empty success: `has_sentry_label` stays `true`, so the modal does surface *an* error box — but with the wrong text (see ❌ above).
- **Redirect bypass on an allowlisted host** — made `sentry-legit` answer `302 → http://attacker:9000/redirected-leak` for the attacker-controlled org segment. Go follows the redirect, but the attacker probe recorded `"authorization": ""` — the bearer token is stripped cross-domain. ✅ no credential leak (residual: an open redirect on the trusted instance can drive an unauthenticated GET elsewhere — out of scope here).
- **Auth & scope** — `GET /api/v1/dashboard/sentry/<fp>` unauthenticated → `401`; forged `notificator-session=deadbeef` → `401`.
- **Persisted / pre-existing state** — the alert seeded by the fake Alertmanager before this change (`sentry` annotation → `https://sentry.company.net/...`) is blocked identically; token save/test flow (`POST /dashboard/sentry-token` → `TestConnection`) still works against the configured instance.
- **Adjacent flow (pre-existing, not caused by this PR)** — the alert-annotation "Sentry" button in the modal header still opens the raw attacker URL in the operator's browser. No token is sent by the app, so it is out of this issue's scope, but the alert-controlled URL is still one click away.

**Suggested fix for the ❌:** in `internal/webui/templates/scripts/dashboard_modal.templ` (~line 509), check `result.error` **before** the `auth_status.has_api_token` branch — set `component.sentryError = result.error` and suppress the "Configure Sentry Token" CTA (`hasSentryToken = true`) for that case, so a refused host reads as a refused host.

**Verdict: FAIL** (4/5 criteria pass; the modal message requirement of criterion 1 fails). The backend half of the fix — the part that actually stops the credential exfiltration — is solid and I could not break it.

SoulKyu:

SoulKyu (commented):
@SoulKyu Re-reviewed this from scratch against head `a289cc5` rather than trusting the prior rounds. The three-commit fix for the token-exfiltration report in #127 holds up well:

- `isAllowedSentryHost` (internal/webui/services/sentry_service.go) is the single gate every alert-supplied Sentry URL passes through in `parseSentryURL`, before any outbound request (including the project-ID resolution call, which used to fire first) is built. Both the `sentry` annotation and the label fallback go through it.
- The configured `BaseURL` requires an exact scheme+host match (no https→http downgrade), while each `AllowedHosts` entry is evaluated independently: a bare entry inherits the configured scheme, a scheme-prefixed entry (e.g. an http-only legacy mirror) may legitimately differ. I traced through empty/whitespace entries, malformed `scheme://` entries, and port-only differences and didn't find a bypass — non-http(s) schemes and empty hosts fail closed before any host comparison runs.
- The outbound base URL is built from the already-validated `parsedURL.Scheme`/`Host`, not re-derived from `config.BaseURL`, so there's no parser-differential gap between validation and the request that's actually sent.
- Confirmed `NOTIFICATOR_SENTRY_ALLOWED_HOSTS` really is space-separated end-to-end: viper's `GetStringSlice` on an env-bound value splits on whitespace, matching the doc comment in `config/config.go`.
- `go build ./...`, `go vet ./config/... ./internal/webui/...`, and `go test ./internal/webui/services/... ./config/...` all pass locally (84 tests across the two packages), and the `httptest`-based tests assert zero requests reach a disallowed host with no leaked `Authorization` header, plus a table-driven `TestIsAllowedSentryHost` covering the configured host, allowlisted bare/scheme-prefixed hosts, attacker hosts, non-http(s) schemes, and scheme downgrade/upgrade attempts.

No new issues found. This is a well-scoped, well-tested fix for a real vulnerability — nice work closing this out.



SoulKyu:
## 🧪 QA Report

Live env: images rebuilt from `a289cc5` (`make docker-build-all`), `docker compose up` with `NOTIFICATOR_SENTRY_ENABLED=true`, `base_url=http://sentry-good:9000`, `allowed_hosts="sentry-extra:9000 http://sentry-apac:9000"`, global token `QA-GLOBAL-TOKEN-MUST-NOT-LEAK` and a saved **personal** token `QA-PERSONAL-TOKEN-MUST-NOT-LEAK`. Four instrumented fake Sentry instances on the compose network (`sentry-good` = configured, `sentry-extra` / `sentry-apac` = allowlisted, `sentry-evil` = attacker) log every request path + `Authorization` header and expose `/_hits`. 40 crafted alerts POSTed to the fake Alertmanager; every probe resets the four counters, issues `GET /api/v1/dashboard/sentry/<fp>`, then reads them back — so "no outbound request" is measured, not inferred. Web UI driven in real Chromium (Playwright).

|---|---|---|
| Alert whose `sentry` annotation points at a non-configured host produces **no outbound HTTP request at all** | ✅ | 31 hostile URL variants → `error: "Sentry host not allowed"` (or `Invalid Sentry URL format`) with `sentry-evil` at **0 hits, 0 `Authorization` headers** in every single case. Verified both with global-token-only and with a personal token saved. |
| …and **the modal shows an explicit "host not allowed" error** | ❌ | Real browser, user has a valid personal token (`/dashboard/sentry-config` → `has_token: true`, `auth_status: personal_token`). Sentry tab of `QAEvilAnnotation` renders **"Unable to load Sentry data / Sentry token not configured / [Configure Sentry Token]"**. Live Alpine state: `{"sentryError":"Sentry token not configured","hasSentryToken":false,"hasProject":false}`. The refusal branch (`internal/webui/services/sentry_service.go:90-94`) returns an **unpopulated `AuthStatus`**, and `loadSentryData` (`internal/webui/templates/scripts/dashboard_modal.templ:509-524`) branches only on `auth_status.has_api_token` and never reads `result.error`. The API carries the right message; nothing user-visible ever shows it. The PR's own test comment says *"HasSentryLabel should stay true so the modal can explain the refusal"* — the modal does not. This head touches only `config/config.go` + `sentry_service.go` + the test, so the ❌ reported on `deca05a`, `a3a25b3` and `4afedaf` is unchanged. |
| Alert pointing at the configured instance keeps working exactly as today | ✅ | `QAGoodHost` → 5 API calls reach `sentry-good` (`/projects/qa/proj/`, `issues/`, `stats/`, `sessions/`, `releases/`) all with `Authorization: Bearer QA-PERSONAL-TOKEN-MUST-NOT-LEAK` and `Host: sentry-good:9000`; modal renders "QA Project / python", release 1.2.3, Apdex 0.95, "QAError: fake sentry issue" (screenshot). `QAGoodHostNoID` (no `?project=`) does 6 calls incl. the project-ID resolution. Global-token fallback identical. Allowlisted `sentry-extra` (bare entry) and `sentry-apac` (scheme-prefixed entry) both work. |
| The check covers both the annotation and the label fallback path | ✅ | `QAEvilLabel` (no annotation, `sentry` **label** = attacker host) → `"Sentry host not allowed"`, 0 hits on all four probes. |
| Unit test asserts no `Authorization` ever leaves for a non-allowlisted host, incl. the project-ID resolution call | ✅ | `go test ./internal/webui/services/... ./config/...` → 84 pass. `TestSentryDataRefusesForeignHosts` asserts `attackerHits == 0` (strictly stronger than header inspection), and its `label fallback` case has no `?project=` so it goes through `fetchProjectID` — the first request — with `GlobalToken` set. Non-vacuous: with the guard neutered in a scratch copy (`if false && !s.isAllowedSentryHost(...)`), **11 subtests fail** (4 of `TestSentryDataRefusesForeignHosts`, 5 of `TestIsAllowedSentryHost`, + parents). |

**Adversarial pass** — every case below was run against the live app with per-request probe-counter diffing; ✅ = held (0 hits on the attacker, no token).

- **URL-parser confusion (20 payloads, `LEAKS: none`)** — userinfo `http://sentry-good:9000@sentry-evil:9000/…`; backslash `http://sentry-good:9000\@sentry-evil:9000/…` and `http:/\sentry-evil:9000/…`; encoded slash `http://sentry-good:9000%2f@sentry-evil:9000/…`; fragment `http://sentry-evil:9000#@sentry-good:9000/…`; NUL `http://sentry-good:9000%00.sentry-evil:9000/…`; embedded space / tab / newline in the authority; `http:////sentry-evil:9000/…`; `[::ffff:sentry-evil]:9000`; uppercase attacker `HTTP://SENTRY-EVIL:9000/…`; `data:`, `javascript:`, `gopher:`, `file:` schemes; scheme-relative `//sentry-evil:9000/…`; bare `sentry-evil:9000`; relative path only; `not a url at all`. All refused before any request was built.
- **Host confusion around the allowed name** — suffix `sentry-good:9000.sentry-evil:9000` ✅; prefix `sentry-good.sentry-evil:9000` ✅; same host different port `sentry-good:9999` ✅; trailing dot `sentry-good.:9000` ✅ (fails closed). Uppercase configured host `http://SENTRY-GOOD:9000/…` is correctly still **allowed** (case-insensitive compare) and fetches normally.
- **Boundary / oversized** — 20 000-char query string on the allowed host → still works (5 calls, correct host); 5 000-char attacker subdomain → refused; empty `sentry: ""` → refused; whitespace-padded URL → `Invalid Sentry URL format`, no request; alert with **no** `sentry` key → `has_sentry_label: false`, no request, no error box.
- **The head commit's specific claim — per-entry scheme independence.** Reconfigured to `base_url=https://sentry-good:9000`, `allowed_hosts="sentry-extra:9000 http://sentry-apac:9000"` and re-ran: `http://sentry-apac:9000/…` is **allowed** (6 hits on `sentry-apac`) even though the primary is https ✅; the bare entry `sentry-extra:9000` **inherits https**, so an `http://sentry-extra:9000/…` alert is refused ✅; `http://sentry-good:9000/…` against the https-configured instance is refused — an alert cannot downgrade the primary ✅; `https://sentry-good:9000/…` is allowed and attempts the connection (TLS error against the plaintext probe, i.e. it got past the guard, as intended). Env parsing of the space-separated `NOTIFICATOR_SENTRY_ALLOWED_HOSTS` works for both entry styles.
- **Error / unavailable path** — stopped the configured instance: `QAGoodHostNoID` correctly surfaces `Failed to resolve project ID: … dial tcp: connect: connection refused`. **Pre-existing wart, not caused by this PR:** when the URL carries `?project=<id>` the resolution call is skipped and every downstream fetch failure is swallowed (`issues = []`, defaulted stats), so the modal renders a Sentry panel that reads as success with 0 issues instead of an error. Same on `main` — flagging it, not counting it against a criterion.
- **Redirect off an allowed host** — made `sentry-good` answer `302 → http://sentry-evil:9000/redirected-leak` for every `/api/0/` call. Go follows all 5 redirects to the attacker but strips the credential: attacker log shows `AUTH= ''` on every hit, while `sentry-good` saw the bearer token. No token leak ✅. Residual (out of this issue's scope): an open redirect on the trusted instance can still drive unauthenticated GETs elsewhere.
- **Persisted / pre-existing state** — reused the existing `postgres-data` volume and user. Saved a personal token whose **own `base_url` is the attacker host** (`/dashboard/sentry-token` → `sentry-config` confirms `base_url: http://sentry-evil:9000`), then re-ran the attacker alert: still refused, still 0 requests — the per-user stored base URL does **not** widen the allowlist ✅. The alert seeded by the fake Alertmanager *before* this change (`QuerylyApplicationError` → `https://sentry.company.net/…`) is blocked identically, 0 requests.
- **Auth & scope** — unauthenticated `GET /api/v1/dashboard/sentry/<fp>` → `401 Authentication required`; unknown fingerprint → `404 Alert not found`.
- **Adjacent flows** — dashboard still lists all 59 alerts, modal opens, other tabs (Overview/Details/Labels/Annotations/History/Comments) unaffected; token save + `TestConnection` against the configured instance still succeeds. Note (pre-existing, unchanged by this PR): the modal header's "Sentry" button still opens the raw alert-controlled URL in the operator's browser, and `POST /dashboard/sentry-token` still lets a user send their own token to an arbitrary `base_url` — neither is alert-driven, both out of scope for #127.

The backend half of the fix — the part that actually stops the credential exfiltration — is solid; I could not construct a single input that got a request or a token to the attacker host. The user-facing half of criterion 1 is still unimplemented at this SHA.

**Suggested fix:** in `internal/webui/templates/scripts/dashboard_modal.templ` (~line 509), test `result.error` **before** the `auth_status.has_api_token` branch — set `component.sentryError = result.error` and `hasSentryToken = true` so the "Configure Sentry Token" CTA is suppressed; and/or populate `AuthStatus` on the refusal return in `sentry_service.go:90-94`.

🔁 Fixer summoned.

**Verdict: FAIL** (4/5 criteria; the modal-message half of criterion 1 fails)

SoulKyu:

internal/webui/services/sentry_service.go
```diff
@@ -78,14 +83,14 @@ func (s *SentryService) GetSentryDataForAlert(alert *models.Alert, userID, sessi
...
 		if errors.Is(err, errSentryHostNotAllowed) {
-			message = "Sentry host not allowed: the alert points at a different host than the configured Sentry instance"
+			errMsg = "Sentry host not allowed"
 		}
 		return &webuimodels.SentryData{
 			HasSentryLabel: true,
 			SentryURL:      sentryURL,
-			Error:          message,
+			Error:          errMsg,
 		}
```
SoulKyu, line 102: **Problem:** On host refusal, `GetSentryDataForAlert` returns `SentryData{HasSentryLabel: true, SentryURL: sentryURL, Error: errMsg}` with `AuthStatus` left at its zero value (`AuthStatus{HasAPIToken: false, AuthMethod: ""}`) — `auth := s.getAuthForUser(...)` isn't computed until line 98, after this return. The frontend (`internal/webui/templates/scripts/dashboard_modal.templ:511-524`, unchanged by this PR) branches on `result.auth_status?.has_api_token` *before* ever reading `result.error`: since `has_sentry_label` is true but `has_api_token` is false, it takes the `else` branch and hardcodes `component.sentryError = 'Sentry token not configured'` with a "Configure Sentry Token" CTA — the real `"Sentry host not allowed"` message from this refusal never reaches the UI, even for a user who already has a valid personal token saved.

**Why it matters:** This is an explicit acceptance criterion of #127, which this PR closes: *"An alert whose `sentry` annotation points at a host other than the configured Sentry instance produces no outbound HTTP request at all… **and the modal shows an explicit "host not allowed" error**."* The PR description itself claims *"the existing error field already renders in the modal"* — that's not what happens for a user with a configured token. The security half is solid (verified: zero requests, zero leaked Authorization headers), but the user-visible half of the criterion doesn't hold, and a responder investigating a real alert has no way to tell "this alert is misconfigured/malicious" apart from "I forgot to set up my Sentry token."

**Evidence:** Live QA against this exact code (unchanged since `deca05a`, confirmed again at head `a289cc5` — same tree as this head `e9be716`) reproduced it with a real browser: user has `has_token: true` / `auth_status: personal_token` configured, yet the Sentry tab renders "Unable to load Sentry data / Sentry token not configured / [Configure Sentry Token]" for an alert whose `sentry` annotation points at a disallowed host, while the live Alpine component state showed `{"sentryError":"Sentry token not configured","hasSentryToken":false}`. Reproducible by reading `dashboard_modal.templ:511`: `if (result.auth_status?.has_api_token) { … } else { component.sentryError = 'Sentry token not configured'; }` — `result.error` is never consulted in that branch.

**Suggested fix:** within this PR's own file, move `auth := s.getAuthForUser(userID, sessionID)` above the `parseSentryURL` call (or call it again on this path) and populate `AuthStatus` on the refusal return, e.g. `AuthStatus: webuimodels.SentryAuthStatus{HasAPIToken: auth.AuthMethod == "personal_token", AuthMethod: auth.AuthMethod}`, mirroring what the `auth.AuthMethod == "none"` and project-ID-resolution-failure branches already do a few lines below. That alone won't fix the frontend's error-branch ordering, so `dashboard_modal.templ:511` should also check `result.error` before `auth_status.has_api_token` so the actual refusal reason surfaces regardless of token state — but the backend half of that fix is squarely inside this PR's changed lines.


SoulKyu (commented):
@SoulKyu Re-verified this Sentry host-allowlist fix from scratch against head `e9be716` (byte-identical to the previously-reviewed `a289cc5` in `config/config.go`, `sentry_service.go`, `sentry_service_test.go` — only an unrelated `devtools/factory-tui` file moved in the rebase). `isAllowedSentryHost` is a solid single choke point: it's checked before any outbound request (including project-ID resolution), covers both the annotation and label fallback, refuses scheme downgrades on the configured host while letting `AllowedHosts` entries pick their own scheme, and `go build`/`go test ./internal/webui/services/...` pass locally (`TestIsAllowedSentryHost` and `TestSentryDataRefusesForeignHosts` both green). The core credential-exfiltration vulnerability from #127 is closed: I could not get a request or an Authorization header to a disallowed host.

One actionable gap remains, verified against the exact lines this PR touches — see inline comment. It doesn't undermine the security fix, so it's non-blocking, but it's worth closing out since it's called out explicitly in #127's own acceptance criteria.



internal/webui/services/sentry_service.go
SoulKyu: @SoulKyu This has been fixed in 3949a75.

Moved getAuthForUser() above the host-check in GetSentryDataForAlert (internal/webui/services/sentry_service.go) so AuthStatus is populated on the refusal return instead of left zero-valued, and updated dashboard_modal.templ:509-517 to check result.error before auth_status.has_api_token so the real 'Sentry host not allowed' message now reaches the UI. Added a regression test asserting AuthStatus.AuthMethod is resolved on refusal.




SoulKyu:
**Looper fixer round complete** — 3949a75

- ✅ Review comment on `internal/webui/services/sentry_service.go:102` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/134#discussion_r3689601288)
  - Moved getAuthForUser() above the host-check in GetSentryDataForAlert (internal/webui/services/sentry_service.go) so AuthStatus is populated on the refusal return instead of left zero-valued, and updated dashboard_modal.templ:509-517 to check result.error before auth_status.has_api_token so the real 'Sentry host not allowed' message now reaches the UI. Added a regression test asserting AuthStatus.AuthMethod is resolved on refusal.


SoulKyu:
## 🧪 QA Report

Env: `make test` (full rebuild) + a compose override adding three mock Sentry instances on the
compose network — `sentry-good:9000` (= `NOTIFICATOR_SENTRY_BASE_URL`), `sentry-evil:9000`
(attacker, logs every request + `Authorization`), `sentry-mirror:9000` (allowlisted).
`NOTIFICATOR_SENTRY_ALLOWED_HOSTS="sentry-mirror:9000 https://sentry-httpsonly:9000"`.
20 crafted alerts POSTed to the fake Alertmanager's `/api/v2/alerts`, then driven through
`GET /api/v1/dashboard/sentry/:fingerprint` and through the real modal in headless Chromium.

|---|---|---|
| Alert whose `sentry` annotation points at a foreign host → **no outbound HTTP request at all** | ✅ | Attacker mock received **0 requests** across the whole matrix (20 URLs, 4 full driver runs, all token modes). Refusals return `{"error":"Sentry host not allowed"}` |
| …and the modal shows an explicit "host not allowed" error | ✅ | Browser: Sentry tab renders "**Unable to load Sentry data — Sentry host not allowed**". Verified with a personal token, with only the global token, and with **no token at all** (`auth_method:"none"`) — the pre-fix masking as "Sentry token not configured" is gone |
| Alert pointing at the configured instance keeps working | ✅ | `z01`: project resolved (`id 4242`), issues/sessions/stats/releases all fetched from `sentry-good` with `Bearer …`; modal renders the full Sentry panel (project card, Crash-Free tiles, Recent Issues). Same via `AllowedHosts` (`sentry-mirror`, 6 requests) |
| The check covers both annotation **and** label fallback | ✅ | `z02` (annotation) and `z03` (label only, no annotation) both → `Sentry host not allowed`, 0 attacker hits; `z18` (label, configured host) still fetches data |
| Unit test asserts no `Authorization` ever leaves for a non-allowlisted host, incl. the project-ID call | ✅ | `TestSentryDataRefusesForeignHosts` fails on any hit to the `httptest` attacker server (the project-ID call is the first request built) and now also asserts `AuthStatus.AuthMethod` is resolved. `go test ./internal/webui/services/... ./config/...` → 84 pass |

**Adversarial pass** — what I threw at it (all against the running app):

- **Host-parsing tricks** — userinfo prefix `http://sentry-good:9000@sentry-evil:9000/…` → refused, attacker got nothing; reverse form `http://sentry-evil:9000@sentry-good:9000/…` → correctly allowed (the real host is the configured one). Suffix trick `http://sentry-good.evil.test:9000/…` (DNS alias on the attacker container) → refused. Trailing-dot FQDN `http://sentry-good.:9000/…` → refused. Scheme-relative `//sentry-evil:9000/…` and opaque `https:sentry-evil:9000/…` → refused by the `Host == ""` guard.
- **Schemes** — `file:///etc/passwd/organizations/…` → refused. Uppercase `HTTP://sentry-good:9000/…` → allowed (correct, scheme is case-insensitive). Scheme downgrade against an https-only allowlist entry (`http://sentry-httpsonly:9000/…`) → refused, proving each `AllowedHosts` entry keeps its own scheme.
- **Case folding** — `http://SENTRY-GOOD:9000/…` is accepted by `EqualFold`. I first "broke" this by registering `SENTRY-GOOD` as a Docker network alias on the attacker container, and did see the token land there — but that only works because Docker's embedded DNS matches aliases case-sensitively. With the alias removed, `SENTRY-GOOD` resolves to the real host, as any RFC-4343 resolver would. **Not a defect** — recorded so nobody re-derives it as one.
- **Malformed / empty** — empty `sentry: ""` → refused; leading-space URL → `Invalid Sentry URL format`; allowed host with a non-project path (`/foo/bar`) → `Invalid Sentry URL format`; CRLF in the path (`…/qa-proj/%0d%0aX-Injected:%20yes`) → org/project still parse clean, no header injection, request goes to the configured host only.
- **Error / unavailable path** — configured host returning 500 on the project-ID call → modal shows `Failed to resolve project ID: project API request failed with status 500` instead of a blank "success" panel (the new `if (result.error)` branch in `dashboard_modal.templ` actually improves this). Configured host 302-redirecting to the attacker → Go strips `Authorization` on the cross-host redirect: attacker logged 9 hits with `AUTH=''`, no token leak; that path needs the *trusted* instance to turn hostile, outside this issue's threat model.
- **Auth & scope** — unauthenticated and forged-cookie `GET /api/v1/dashboard/sentry/<fp>` → 401; unknown fingerprint → 404. Ran the full matrix under all three auth modes (personal token / global token / none): the refusal is identical and no token leaves in any of them.
- **Persisted state** — saved a per-user Sentry config with `base_url=http://sentry-evil:9000`; the alert data path still refuses `sentry-evil` (it reads `config.Sentry.BaseURL`, not the stored per-user URL). Side note, out of scope: `POST /api/v1/dashboard/sentry-token` still runs `TestConnection` against the caller-supplied `base_url`, and the attacker mock did receive `Bearer QA-EVIL-BASEURL-TOKEN` there. Self-inflicted by the authenticated user, not alert-driven, pre-existing.
- **Adjacent flows** — alert with no `sentry` annotation/label: the Sentry tab isn't rendered at all (unchanged). "Sentry token not configured" still shows for an allowed host with no token. Configured-instance rendering unchanged for both annotation and label sources.

⚠️ **One limitation worth a follow-up (not a criterion failure).** Host matching is literal, so a
default port is not canonicalised. With `base_url: http://sentry-p80`, an annotation of
`http://sentry-p80:80/organizations/…` — the same URL — is refused with "Sentry host not allowed"
(verified live: the bare form fetched data, the `:80` form was refused). Real-world equivalent:
`base_url: https://sentry.example.com` plus an annotation carrying `:443`, or the reverse. It fails
**closed** with a clear message and an operator can work around it by adding `host:443` to
`allowed_hosts`, so it does not block this fix — but stripping the default port before comparing
would remove the footgun.

**Verdict: PASS**

Issue #130, opened 2026-07-27
feat(silences): advanced silence creation modal on the Silences page


SoulKyu:
## Problem

The Silences page (`/silences`) is currently read-only for creation: silences can only be created implicitly through the alert bulk *silence* action, which auto-derives matchers from one alert's labels and offers a fixed duration. There is no way to author a silence from scratch — arbitrary matchers, regex, chosen sources, future schedule.

## Feature

Add a **Create silence** button on the Silences page opening a full-featured creation modal. The modal must be power-user grade and impose **no artificial limits** on what the user can express — everything Alertmanager's API accepts must be expressible.

### Matchers
- Dynamic matcher rows (add/remove, no cap): `name` + operator (`=`, `!=`, `=~`, `!~`) + `value`.
- Regex operators map to `isRegex: true`, negations to `isEqual: false`.
- Autocomplete label names and values from the live alert cache.
- Paste-to-import: parse a PromQL-style matcher block (`{alertname=~"Vault.*", team!="infra"}`) into rows.
- Client-side regex validation (invalid regex flagged before submit; Alertmanager anchors with `^(?:...)$` — reuse the same semantics as `compileMatchers` in `internal/webui/handlers/silence_handlers.go`).

### Targets
- Choose the destination Alertmanager(s): one, several, or all configured sources (checkbox list from `GetClientNames()`), default all.
- Per-source result reporting: creating on 3 sources where 1 fails must show which failed and why, not a global error.

### Schedule
- `startsAt`: now (default) or a future datetime (pending silence).
- `endsAt`: duration presets (1h, 4h, 24h, 7d, …) + free-form duration input (`90m`, `36h`, `2w`) + absolute datetime picker.
- **No maximum duration clamp** — the 30-day `maxSilenceExtension` cap of the extend endpoint must NOT apply to creation.

### Preview (the killer part)
- Live preview of the alerts the silence would match, per selected source, before submitting — count and expandable list (alertname + key labels), reusing the `countMatchedAlerts`/`matchLabels` logic server-side.
- Warn (non-blocking) when the silence would match **0** alerts or an unusually large number — inform, never prevent.

### Misc
- Comment field (required by Alertmanager) with a sensible default.
- `createdBy` auto-filled with the authenticated user's username (see the fix that switched creator display from IDs to usernames).
[...]

PR #139, opened 2026-07-30
feat(webui): editable silence matchers in the silence modal

### LINKED ISSUE #129: feat(webui): editable silence matchers in the silence modal
> Requalifies #27 (user request from @Aluxima, groomed but outside the coordinator's issue-age window). This issue carries the spec forward so the autonomous pipeline can pick it up; #27 stays open as the user-facing thread.

> **Updated 2026-07-27, after #130/#132 merged** (advanced silence creation modal on `/silences`): a large part of the machinery this issue originally planned to build now exists and must be **reused, not reimplemented**. Line references refreshed against current main (`765225a`).

## Problem
Silence matchers are built server-side from the cached alert's labels with no visibility or control. `processSilenceAction` (`internal/webui/handlers/dashboard_handlers.go:1960`, label loop at `:1983-1996`) emits one exact matcher per label (`IsRegex: false, IsEqual: true`), skipping only `__name__`/`__tmp_`. The silence modal shows just two derived fields — `Alert:` and `Instance:` (`internal/webui/templates/components/modal_components.templ:1749-1758`) — so the operator cannot see what the silence will actually match, let alone edit it. The request payload has no matcher field at all (`BulkActionRequest`, `internal/webui/models/dashboard.go:194-202`).

Because every label is pinned, every silence is maximally narrow: silencing a group of related alerts means N individual silences instead of one regexp matcher on the shared label.

Note: #130 shipped a standalone creation modal on `/silences`, but it deliberately did not touch this flow (explicit non-goal of its spec) — silencing **from the dashboard, in the context of an alert**, is still blind and uneditable. That is exactly the pain reported in #27.

## Impact
Operators cannot verify the blast radius of a silence before submitting, and cannot broaden it (drop `instance`, regexp a `pod` name) without leaving for the Alertmanager UI. The audit comment (`dashboard_handlers.go:2064`) records the duration but not what was matched.

## Existing machinery to reuse (shipped by #130)
- **Per-row matcher validation**: `compileDraftMatchers` (`internal/webui/handlers/silence_handlers.go:318`) already compiles each matcher individually with Alertmanager's `^(?:...)$` anchoring and returns `{index, error}` per invalid row. The handler must call it, not duplicate it.
- **Live preview endpoint**: `POST /api/v1/dashboard/silences/preview` (`silence_handlers.go:461`) returns exact per-source match counts + a capped alert sample against the whole alert cache, plus per-row matcher errors. The original plan for a client-side "estimate over loaded alerts" is superseded — call this endpoint instead (debounced, same as the `/silences` modal does).
- **Matcher-row UI pattern**: dynamic rows with `name` / operator select (`=`, `!=`, `=~`, `!~`) / `value` / remove / add, label-name and value autocomplete via `/api/v1/dashboard/available-labels` — all in `internal/webui/templates/pages/Silences.templ:218` onward. Copy the pattern, or factor it into a shared templ component if the duplication feels heavy.

## Proposed approach
Make the matcher set explicit in the modal and send it with the request; keep today's server-side derivation as the fallback so untouched submissions behave identically.

- **Model**: `models.SilenceMatcher` (`internal/models/silence.go:22-27`) already carries `Name/Value/IsRegex/IsEqual` and is what `client.CreateSilence` consumes — no model change. Its custom `UnmarshalJSON` already defaults a missing `isEqual` to `true`, which covers client payloads that omit it. Add a `SilenceMatchers []models.SilenceMatcher` field (JSON key `silenceMatchers`, `omitempty`) to `BulkActionRequest`.
- **Handler**: in `processSilenceAction`, if the request carries matchers, validate them with `compileDraftMatchers(matchers, true)` and use them verbatim; otherwise fall back to the current label loop (`:1983-1996`). Invalid rows → HTTP 400 carrying the per-row `matcherErrors` list (same wire shape as `CreateSilence` at `silence_handlers.go:495`), before any Alertmanager call. An explicit empty matcher list is rejected (would silence everything).
- **Server-side scope guard** (from the roast, the UI must not be trusted): reject explicit `silenceMatchers` when `len(alertFingerprints) != 1` — otherwise two alerts would create the same silence twice. Bulk/group silences keep the per-alert derivation unconditionally.
- **UI** (`modal_components.templ`, silence modal block at `:1749-1758`): replace the Alert/Instance block with an editable matcher list pre-filled from `currentSilenceAlert.labels` — the full alert object is already handed to Alpine at `internal/webui/templates/scripts/dashboard_actions.templ:741`, so no new endpoint is needed. Reuse the row pattern from `Silences.templ:218`. Collapsed behind a "Matchers (N)" disclosure so the common path stays one click.
- **Preview**: debounced call to `POST /api/v1/dashboard/silences/preview` on matcher edits; show the per-source count (this is exact, not an estimate — drop the old "estimate" wording).
- **Scope**: single-alert silences only. Bulk/group silences (`request.GroupNames` loop) keep the per-alert derivation; the matcher editor is hidden when `silenceAction !== 'single'` — and rejected server-side regardless (guard above).
- **Considered alternative** (rejected as primary UX, viable as a bonus): an "Advanced…" button deep-linking into the #130 creation modal prefilled from the alert's labels — the `duplicate()` prefill (`Silences.templ:541`) already does 90% of it. Rejected because it context-switches away from the dashboard for the common "just drop `instance`" case; may still be worth adding as an escape hatch to multi-source/scheduled silences.
- Templ workflow: edit `.templ` sources, then `make webui-templates`; never edit `*_templ.go`.

## Acceptance criteria
- [ ] The silence modal lists every matcher that will be sent (not just Alert/Instance), pre-filled from the alert's labels minus `__name__`/`__tmp_`
- [ ] A matcher can be edited (name and value), removed, and added
- [ ] Each matcher can be switched between `=`, `!=`, `=~`, `!~`
- [ ] An invalid regexp is rejected with HTTP 400 identifying the offending matcher row (reusing `compileDraftMatchers`), before any Alertmanager call
- [ ] A request with an explicit empty matcher list is rejected (would silence everything)
- [ ] A request with explicit matchers and more than one fingerprint is rejected server-side
- [ ] The preview count comes from `POST /api/v1/dashboard/silences/preview` and updates as matchers are edited
- [ ] Submitting the modal without touching the matcher list produces a semantically identical silence to today's (same matcher set; ordering is not deterministic today and is not part of the contract)
- [ ] The audit comment includes the matcher count
- [ ] Bulk and group silences are unaffected (no matcher editor shown, same server path)
- [ ] Only `.templ` sources edited; generated files come from `make webui-templates`

SoulKyu:
## Summary
- The dashboard silence modal previously derived matchers server-side from the alert's labels with no visibility or control — every label was pinned exact, so silencing a group of related alerts meant N separate silences.
- Adds an optional `silenceMatchers` field to `BulkActionRequest`; for a single-alert silence, `processSilenceAction` uses it verbatim (validated with the existing `compileDraftMatchers`), falling back to today's label-derived matchers when absent so untouched submissions are unaffected.
- Server-side guard: explicit matchers are rejected when more than one alert fingerprint is targeted, or when the list is explicitly empty (would silence everything) — before any Alertmanager call. Bulk/group silences always derive matchers per alert regardless of what the client sends.
- Silence modal: single-alert flow gets an editable matcher list (label name, operator `=`/`!=`/`=~`/`!~`, value, add/remove) behind a "Matchers (N)" disclosure, pre-filled from the alert's labels minus `__name__`/`__tmp_`. Edits debounce into the existing `POST /api/v1/dashboard/silences/preview` endpoint for a live, exact match count and surface per-row validation errors.
- The silence audit comment now records the matcher count alongside the duration.
- Reuses existing #130 machinery (`compileDraftMatchers`, the preview endpoint, and the `Silences.templ` matcher-row UI pattern) rather than reimplementing it.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/webui/...` (134 tests, including new coverage: empty explicit matcher list rejected, explicit matchers + multiple fingerprints rejected, invalid regex matcher rejected with the correct row index)
- [x] `make webui-templates` regenerates `*_templ.go` with no diff (only `.templ` sources hand-edited)

Closes #129


internal/webui/templates/components/modal_components.templ
```diff
@@ -1786,15 +1786,64 @@ templ SilenceModal() {
...
 								<div x-show="silenceAction === 'single' && currentSilenceAlert" class="mb-4 p-3 bg-gray-50 dark:bg-dark-bg-tertiary rounded-md">
 									<div class="flex items-center space-x-2 text-sm">
 										<span class="font-medium text-gray-900 dark:text-white">Alert:</span>
 										<span class="text-gray-600 dark:text-gray-300" x-text="currentSilenceAlert?.alertName"></span>
 									</div>
-									<div class="flex items-center space-x-2 text-sm mt-1">
-										<span class="font-medium text-gray-900 dark:text-white">Instance:</span>
-										<span class="text-gray-600 dark:text-gray-300" x-text="currentSilenceAlert?.instance"></span>
+
+									<button type="button" @click="showSilenceMatchers = !showSilenceMatchers"
```
SoulKyu, line 1800: **Single-alert silence modal drops the always-visible `Instance:` label.**

Problem: the previous info panel always showed both `Alert:` and `Instance:` for the alert being silenced. This change deletes the `Instance:` row (`origin/main` had `<span class="font-medium ...">Instance:</span><span ... x-text="currentSilenceAlert?.instance">` right after the `Alert:` row) and replaces it with the `Matchers (N)` disclosure toggle added here, which defaults to collapsed (`showSilenceMatchers: false` in `dashboard_core.templ`).

Why it matters: `instance` is exactly the label an operator uses to confirm *which* target is about to be silenced when several alerts share the same alert name (e.g. `HighCPU` firing on ten different pods/hosts). That confirmation used to be on screen unconditionally; now it only appears if the user manually expands the matcher editor, so someone moving quickly (or a half-asleep on-call engineer) can hit "Silence" without ever seeing which instance is targeted — an easy way to silence the wrong instance's alert.

Evidence: `git diff origin/main...HEAD -- internal/webui/templates/components/modal_components.templ` shows the `Instance:` block removed with no replacement instance-specific display; the new block only re-shows the alert name plus the collapsed matcher editor.

Suggested change: keep a small always-visible instance/target line (e.g. reading the `instance` label straight off `currentSilenceAlert`) above or beside the `Matchers (N)` toggle, independent of whether the matcher editor is expanded.


SoulKyu (commented):
@SoulKyu Reviewed the editable-silence-matchers change (#137/#130 groundwork reused nicely). Verified the server-side guard in `BulkActionAlerts`/`processSilenceAction`: explicit matchers are rejected outright when more than one fingerprint is targeted or the list is empty, compiled through the existing `compileDraftMatchers(..., true)` before ever touching Alertmanager, and group/bulk silences never receive the explicit set (they always pass `nil` through `processGroupAction`) — so there's no way for a bulk/group request to smuggle a single alert's matcher override onto other alerts. `go build`, `go vet`, and `go test ./internal/webui/...` (134 tests) all pass, and `make webui-templates` regenerates with no diff. One UX regression worth a look below, otherwise this holds up well.



internal/webui/templates/components/modal_components.templ
SoulKyu: @SoulKyu This has been fixed in 8c0c781.

Restored the always-visible Instance row above the Matchers disclosure toggle in internal/webui/templates/components/modal_components.templ (single-alert silence info panel), reading currentSilenceAlert?.instance the same way currentAckAlert?.instance is shown elsewhere.




SoulKyu:
**Looper fixer round complete** — 8c0c781

- ✅ Review comment on `internal/webui/templates/components/modal_components.templ:1800` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/139#discussion_r3682842546)
  - Restored the always-visible Instance row above the Matchers disclosure toggle in internal/webui/templates/components/modal_components.templ (single-alert silence info panel), reading currentSilenceAlert?.instance the same way currentAckAlert?.instance is shown elsewhere.


SoulKyu:
## 🧪 QA Report

Booted the full stack (`make test` path: templ generate → tailwind → `docker-build-all` → `docker compose up`; postgres + fake-alertmanager + backend + webui all healthy). Exercised every criterion against the live app with `curl` and a real Chromium session (Playwright), then attacked each one.

|---|---|---|
| Modal lists every matcher, pre-filled from labels minus `__name__`/`__tmp_` | ✅ | Opened the modal on a live alert: disclosure reads `Matchers (8)`, rows exactly equal `currentSilenceAlert.labels` (incl. the long `sentry` URL label), no `__name__`/`__tmp_`. Collapsed by default (one-click common path preserved). |
| A matcher can be edited (name and value), removed, and added | ✅ | Typed into a value input → row updated + preview refetched; `+ Add matcher` → 9 rows, label became `Matchers (9)`; removed → 8. Removing a **middle** row (idx 1) kept Alpine state and DOM inputs in sync despite `x-for :key="idx"`. |
| Each matcher switchable between `=`, `!=`, `=~`, `!~` | ✅ | Drove the real `<select>` on row 2. Payload mapping verified per option: `=`→`{isRegex:false,isEqual:true}`, `!=`→`{false,false}`, `=~`→`{true,true}`, `!~`→`{true,false}`. |
| Invalid regexp → HTTP 400 identifying the row, before any Alertmanager call | ✅ | `[unclosed` at index 1 → `400 {"error":"invalid matchers","matcherErrors":[{"index":1,"error":"invalid regex: ... missing closing ]: \`[unclosed)$\`"}]}`. Alertmanager silence count unchanged (4→4) and no silence with that comment ever reached AM. In the UI the row border turns red and the message renders under the row. |
| Explicit empty matcher list rejected | ✅ | `"silenceMatchers": []` → `400 "at least one matcher is required"`. UI counterpart: deleting all rows disables the Silence button (`silenceMatchersInvalid()`). |
| Explicit matchers + more than one fingerprint rejected server-side | ✅ | 2 fingerprints + matchers → `400 "explicit silence matchers require exactly one alert fingerprint"`. Also rejected for a group silence (0 fingerprints). |
| Preview count comes from `POST /silences/preview` and updates as matchers are edited | ✅ | Observed the debounced POST on every edit; count moved `1 → 0 → 41 → 6` as matchers changed. Counts are internally exact: `severity=warning` (22) + `severity!=warning` (13) = 35 = full cache (`alertname=~.*`). Regex anchoring correct — `severity=~warn` matches 0, `warn.*` matches all warnings. |
| Untouched submission = semantically identical silence | ✅ | UI submit with the matcher list untouched sent exactly the label-derived set (5 matchers, all `isRegex:false,isEqual:true`). Fallback path (field omitted) produced a byte-identical matcher set in Alertmanager vs. the alert's labels. |
| Audit comment includes the matcher count | ✅ | Explicit 2 matchers → `🔇 Alert silenced for 1h: audit-check (2 matcher(s))`. Fallback path also counts correctly → `... fallback-audit (5 matcher(s))`. |
| Bulk and group silences unaffected | ✅ | Both modals render **no** matcher disclosure (`matcherDisclosureVisible=0`), `silenceMatcherRows=[]`, and their payloads carry no `silenceMatchers` key. Group silence still succeeds (`processedCount:1`); server rejects explicit matchers on the group path regardless. |
| Only `.templ` sources edited; generated files from `make webui-templates` | ✅ | Ran `templ generate` at this SHA → **zero** diff across 143 updated files. |
| ⚠️ Layout: expanding "Matchers (N)" clips the modal's own fields | ❌ *(no stated criterion; reported as a regression)* | Modal card is 512px with `overflow-x:hidden`. Collapsed, the duration select and Silence Reason textarea sit 25px **inside** the card; expanded they extend **31px past** it and are permanently cut off (no scroll possible). Reproducible at 1500/1280/768px; fine at 390px. See detail below. |

### Layout regression detail

Introduced by this PR and triggered by the feature's primary interaction — one click on `Matchers (N)`:

```
768px viewport, same modal, same alert:
  COLLAPSED: cardR=640  textareaR=615  → -25px (inside)   durationR=615 → -25px
  EXPANDED : cardR=640  textareaR=671  → +31px (CLIPPED)  durationR=671 → +31px
  card overflow-x: hidden  → the clipped region is unreachable
```

Cause: the matcher row `div.flex.items-center.gap-2` establishes a min-content width (170 + 49 + 170 + 26 + gaps) larger than the content column; the `flex-1 min-w-0` on the inputs is defeated because the ancestor `div.sm:flex.sm:items-start` has no `min-w-0` (`scrollWidth 518` vs `clientWidth 462`). The whole content column widens, so `w-full` children below the matcher block (duration select, the **required** Silence Reason textarea, the info banner) render past the card edge and are clipped.

**Not a functional blocker** — I verified the flow still completes with the block expanded: typed into the clipped textarea, changed the duration to `4h`, submitted, modal closed cleanly with 5 matchers sent. Cosmetic but on the feature's main path; fix is a `min-w-0` (or `flex-wrap`) on the container.

### Adversarial pass

Everything below was run against the live app; nothing here broke an acceptance criterion.

- **Boundary & empty** — empty matcher list → 400; all rows deleted in the UI → submit button disabled; single matcher → OK; 1000 matchers → 200, no crash; alert whose only labels are the standard five → count matches exactly.
- **Malformed & hostile** — `[unclosed` and `[bad` regexes → 400 with the correct row index (and surfaced per-row in the UI); 64KB regex value → handled; catastrophic-backtracking shape `(a+)+$` → compiled safely (Go RE2, no ReDoS); `isRegex:"yes"` (wrong type) → `400 Invalid request format`; nonexistent fingerprint → `1 of 1 actions failed`, no silence created; typed `a/b?c*d"e\f` straight into a value input then switched it to `=~` — no crash, no console error, correct preview.
- **`isEqual` omitted** (the dangerous default) — `{"name":...,"value":...,"isRegex":false}` with no `isEqual` decoded to `isEqual:true`, not a silently negated `!=` matcher. Confirmed on the created Alertmanager silence.
- **Error / unavailable path** — aborted the `/silences/preview` request mid-edit: the UI shows `Failed to fetch` in red. It does leave the previous count rendered next to the error (`Will match 0 alert(s)  Failed to fetch`), so it's a visible failure rather than a silent success — worth a `silencePreviewCount = null` in the catch block, but not a wiped-state-as-success.
- **Auth & scope** — unauthenticated `POST /bulk-action` with matchers → `401 Authentication required`. Explicit matchers rejected for >1 fingerprint and for group flows (server-side, not UI-trusted).
- **Adjacent flows** — the `/silences` page (#130 machinery this PR reuses) still renders, no duplicate DOM ids from the new `dashboard-silence-label-*` datalists; typing `/` inside a matcher input does not trigger a dashboard hotkey and the modal stays open; bulk and group silence flows unchanged.
- **State that outlives the modal** — poisoned `silenceMatcherRows`/`silenceMatcherErrors`/`silencePreviewCount`, cancelled, reopened on a *different* alert: no leak, rows re-derived from the new alert, errors cleared, disclosure re-collapsed, count refetched.
- **Persisted state** — verified against pre-existing silences in the store and against alerts created before the change; the fallback (no `silenceMatchers`) path produces matcher sets identical to `main`'s.

**Out of scope, pre-existing (not this PR):** every webui-created silence sends `"id": ""` (`models.Silence.ID` lacks `omitempty`), so the fake Alertmanager treats each new one as an update and overwrites the previous webui silence. Reproduced identically on the fallback path, which is `main`'s code path — flagging only because it makes silence counts confusing when testing. Also: `compileDraftMatchers` rejects only empty matcher names, not invalid Prometheus label names (`"   "`, `"bad name!"` pass through); the fake AM accepts them, a real Alertmanager would 400 and surface as a failed action.

**Verdict: PASS**

SoulKyu:
## 🧪 QA Report

Full stack booted from this SHA (`make test`: `templ generate` → tailwind → `docker-build-all` → `docker compose up`; postgres + fake-alertmanager + backend + webui all healthy). Every criterion exercised against the live app with `curl` and a real Chromium session (Playwright), then attacked.

This head adds one commit on top of the previously-QA'd `8f5931c`: `fix(webui): restore always-visible Instance row`. That row works. The **layout clipping flagged on the previous head was not fixed, and this run proves it is not cosmetic** — it breaks matcher removal.

|---|---|---|
| Modal lists every matcher, pre-filled from labels minus `__name__`/`__tmp_` | ✅ | Rows exactly equal `currentSilenceAlert.labels`. Injected an alert carrying literal `__name__=up` and `__tmp_=scratch` into the fake AM: labels `[__name__, __tmp_, alertname, instance, job, severity, team]` → rows `[alertname, instance, job, severity, team]`, and the submitted payload carried the same five. |
| A matcher can be edited (name and value), removed, and added | ❌ | **Removal is broken for a real user at every viewport ≥640px.** See below. |
| Each matcher switchable between `=`, `!=`, `=~`, `!~` | ✅ | Drove the real `<select>`. Payload mapping per option: `=`→`{isRegex:false,isEqual:true}`, `!=`→`{false,false}`, `=~`→`{true,true}`, `!~`→`{true,false}`. |
| Invalid regexp → HTTP 400 identifying the row, before any Alertmanager call | ✅ | `[unclosed` at index 1 → `400 {"error":"invalid matchers","matcherErrors":[{"index":1,"error":"invalid regex: … missing closing ]: \`[unclosed)$\`"}]}`. AM silence count unchanged across all four reject cases. In the UI the row border turns red and the message renders under the row. |
| Explicit empty matcher list rejected | ✅ | `"silenceMatchers": []` → `400 "at least one matcher is required"`. All-empty-name rows → `400 {"index":0,"error":"matcher name is required"}`. UI: deleting all rows disables the Silence button (`silenceMatchersInvalid()` true). |
| Explicit matchers + more than one fingerprint rejected server-side | ✅ | 2 fingerprints → `400 "explicit silence matchers require exactly one alert fingerprint"`; same for a group silence (0 fingerprints). |
| Preview count from `POST /silences/preview`, updates as matchers are edited | ✅ | Debounced POST on every edit; count moved `1 → 3 → 4 → 2`. Internally exact: `severity=critical` (12) + `severity!=critical` (14) = 26 = `alertname=~.*` (26) = full cache. Empty set → 0 (not "everything"). |
| Untouched submission = semantically identical silence | ✅ | UI submit untouched sent exactly the label-derived set (8 matchers, all `isRegex:false,isEqual:true`). Fallback path (field omitted **and** `null`) produced the identical 5-matcher set in Alertmanager. |
| Audit comment includes the matcher count | ✅ | Explicit 2 matchers → `🔇 Alert silenced for 1h: qa-audit-check (2 matcher(s))`; fallback → `… qa-fallback (5 matcher(s))`. |
| Bulk and group silences unaffected | ✅ | Both modals render 0 matcher inputs and 0 `Instance:` rows, `silenceMatcherRows=[]`, payloads carry no `silenceMatchers` key; server rejects explicit matchers on those paths regardless. |
| Instance row restored (this commit) | ✅ | Visible for every single-alert modal with the correct value across 4 alerts (`server-14.example.com`, `queryly-back-19.company.net`, `lb-14.example.com`, `server-10.example.com`); `unknown` for an injected alert with no `instance` label (matches `GetInstance()` behaviour elsewhere in the UI); absent in bulk/group modals. |
| Only `.templ` sources edited; generated files from `make webui-templates` | ✅ | `templ generate` (v0.3.906, the go.mod version) at this SHA → **zero** diff. |
| `go test ./internal/webui/...` | ✅ | 134 tests pass. |

### ❌ The remove (✕) button of every matcher row is rendered outside the modal card — clicking it closes the modal instead of removing the matcher

`modal_components.templ:1817` puts the remove button at the end of a `div.flex.items-center.gap-2` row whose min-content width (170 + 49 + 170 + 26 + gaps ≈ **542px**) exceeds the modal card's 510px content box. `flex-1 min-w-0` on the inputs is defeated because the ancestor `div.sm:flex.sm:items-start` (`:1765`) has no `min-w-0`. The card carries `overflow-hidden` (`:1755`), so the excess is clipped and unreachable — there is no scrollbar.

Measured at 1280×900, first matcher row:

```
card right edge      = 896
remove button        = x 889 … 915   → 19 of its 26px clipped
document.elementFromPoint(902, 236)  → DIV (the modal backdrop), NOT the button
```

Real mouse click on the button's visual centre, at seven viewport widths — a reason was typed first to show what is lost:

```
W=1920 click(1222,236) rows 5->5  modalOpen true->false
W=1440 click( 982,236) rows 5->5  modalOpen true->false
W=1280 click( 902,236) rows 5->5  modalOpen true->false
W=1024 click( 774,236) rows 5->5  modalOpen true->false
W=768  click( 646,236) rows 5->5  modalOpen true->false
W=640  click( 582,236) rows 5->5  modalOpen true->false
W=390  click( 324,300) rows 5->4  modalOpen true->true    <- only mobile works
```

The click lands on the backdrop `div` whose `@click="showSilenceModal = false"` fires: the matcher is **not** removed, the modal closes, and the operator's edits and typed reason are discarded. Only a ~7px invisible sliver (x 889–896) actually hits the button — reachable by accident, not by aim. In the screenshot no ✕ is visible on any row at all.

Same clipping pushes the duration `<select>`, the **required** Silence Reason textarea and the info banner 31px past the card edge (`scrollWidth 542` vs `clientWidth 510`) whenever "Matchers (N)" is expanded — collapsed they sit 25px inside. Expanding the disclosure is this feature's primary interaction.

For comparison, the `/silences` creation modal this PR says it reuses (#130, `Silences.templ:218`) puts the same button **25px inside** its card and hit-tests to the button correctly — the pattern was copied into a container 1.5× narrower without re-checking the fit.

Fix is small: `min-w-0` on the `sm:flex sm:items-start` wrapper (or `flex-wrap` on the matcher row).

### Adversarial pass

Everything below ran against the live app.

- **Boundary & empty** — `silenceMatchers: []` → 400; a single row with an empty name → 400 index 0; one valid + one empty name → 400 index 1; all rows deleted in the UI → Silence button disabled; single matcher → OK; alert with only 3 labels → exactly 3 rows, preview 2.
- **Malformed & hostile** — `[unclosed` / `[bad` → 400 with the correct row index, surfaced per-row in the UI; catastrophic-backtracking shape `(a+)+` → compiled safely (Go RE2) and stored verbatim; `isRegex:"yes"` (wrong type) → `400 Invalid request format`; nonexistent fingerprint + matchers → `1 of 1 actions failed`, no silence created; typed `a/b?c*d` straight into a value input — no crash, no console error, modal stayed open.
- **`isEqual` omitted** (the dangerous default) — `{"name":"team","value":"platform","isRegex":false}` decoded to `isEqual:true` on the created Alertmanager silence, not a silently negated `!=`.
- **Error / unavailable path** — aborted `/silences/preview` mid-edit → `silencePreviewCount=null`, red `Failed to fetch`; forced a 500 → red `boom`, count null. One nit: if a count was already rendered and a *later* edit's preview fails, the stale count stays on screen next to the error (`Will match 2 alert(s)  Failed to fetch`) — visible failure, not a silent success, but `silencePreviewCount = null` in the catch would be cleaner.
- **Auth & scope** — unauthenticated `POST /bulk-action` with matchers → `401 Authentication required`. Explicit matchers rejected server-side for >1 fingerprint and for group flows regardless of what the client sends.
- **Adjacent flows** — bulk and group modals show no matcher editor and no Instance row; the `/silences` page modal still opens and renders its own matcher rows correctly; typing `/` inside a matcher value input does not fire a dashboard hotkey and the modal stays open.
- **State that outlives the modal** — poisoned `silenceMatcherErrors`/`silencePreviewCount`, closed, reopened on a *different* alert: rows re-derived from the new alert, errors cleared, disclosure re-collapsed, count refetched, Instance row updated. No leak.
- **Persisted / pre-existing state** — verified against silences already in the store and alerts created before this change; the fallback path produces matcher sets identical to `main`'s.
- **Removal specifically** — `:key="idx"` on the `x-for` is fine: removing the **middle** row (idx 1) kept Alpine state and the DOM inputs in sync (`[alertname, instance, job, severity, team]` → `[alertname, job, severity, team]`, DOM values matching, preview recomputed 1→4). The defect is purely that the control is unclickable.

**Out of scope, pre-existing (not this PR):** every webui-created silence sends `"id": ""` (`models.Silence.ID` lacks `omitempty`), so the fake Alertmanager treats each new one as an update and overwrites the previous webui silence — reproduced identically on the fallback path, which is `main`'s code. Also `compileDraftMatchers` rejects empty matcher names but not invalid Prometheus label names.

**Verdict: FAIL** (1 of 11 criteria) — fixer summoned.

SoulKyu:

SoulKyu (commented):
@SoulKyu Reviewed the current head (cb96d5f) on top of the prior QA/review rounds — the restored `Instance:` row looks correct, and the server-side guard (`BulkActionAlerts` → `processSilenceAction`: explicit matchers rejected for >1 fingerprint or an empty list, compiled through `compileDraftMatchers(..., true)` before any Alertmanager call, and `processAlertAction(..., nil)` for the group path so bulk/group requests can never smuggle a single alert's override) still holds up well, with solid new backend test coverage for the three rejection paths.

**This review reads as BLOCKING** even though GitHub shows it as a plain comment: self-authored PRs can't receive `REQUEST_CHANGES` from the same account (GitHub 422s it), so this is posted as COMMENT with the severity stated here in prose instead. One blocking issue survives from the earlier QA rounds: the new matcher-row markup overflows its container and makes the per-row remove button unclickable on desktop. Left one inline comment with the concrete repro and a minimal fix. Thanks for iterating on this — the matcher preview/debounce/validation wiring is genuinely solid, this is just a container-width miss.



internal/webui/templates/components/modal_components.templ
```diff
@@ -1796,6 +1796,59 @@ templ SilenceModal() {
...
+													<datalist :id="'dashboard-silence-label-values-' + idx">
+														<template x-for="value in (silenceLabelOptions[row.name] || [])" :key="value">
+															<option :value="value"></option>
+														</template>
+													</datalist>
+													<button type="button" @click="removeSilenceMatcher(idx)" class="p-1.5 rounded-md text-gray-400 hover:text-red-600 dark:hover:text-red-400 hover:bg-gray-100 dark:hover:bg-dark-bg-secondary">
+														<svg class="w-3.5 h-3.5" fill="none" stroke="currentColor" viewBox="0 0 24 24">
+															<path stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="M6 18L18 6M6 6l12 12"/>
+														</svg>
+													</button>
```
SoulKyu: **Matcher row overflows the modal card, making the remove (✕) button unclickable on desktop.**

Problem: this new row (`<div class="flex items-center gap-2">` at 1816, holding a `flex-1 min-w-0` name input, the op `<select>`, a `flex-1 min-w-0` value input, and the remove button at 1835) is min-content-wide roughly 542px — two flex-1 inputs each need their own minimum content width, plus the select and the button. It lives inside `<div class="mt-3 ... sm:ml-4 sm:text-left w-full">` (line 1772) which is itself a flex item of `<div class="sm:flex sm:items-start">` (line 1765). Neither of those pre-existing ancestors sets `min-w-0`, so per the flexbox spec they clamp to their content's min-width instead of shrinking to the 512px card (`sm:max-w-lg`, line 1755, which also carries `overflow-hidden`). `flex-1 min-w-0` on the two inputs is defeated because the *ancestor* never shrinks in the first place.

Why it matters: at every viewport ≥640px the remove button's clickable area is clipped to a ~7px sliver at the card's right edge; a real click on the button's visual position lands on the modal backdrop instead (`@click="showSilenceModal = false"` on the backdrop, elsewhere in this file), closing the modal and discarding the operator's in-progress matcher edits and typed reason. This also pushes the duration `<select>` and the required Silence Reason textarea outside the card once "Matchers (N)" is expanded — expanding that disclosure is this feature's primary interaction. Only works by accident at ≤390px, where everything wraps to full width. (Confirmed by two live-app QA passes on this PR at commits 8f5931c and 8c0c781 — reproduced with real mouse clicks at 1920/1440/1280/1024/768/640px, all closing the modal instead of removing the row; not fixed by the intervening Instance-row-restore commit.)

Suggested fix: constrain the new row instead of relying on the unrelated ancestor markup — add `flex-wrap` to this row's class list (line 1816: `class="flex items-center gap-2 flex-wrap"`) so the inputs/select/button wrap onto a second line instead of forcing the container wider than the card. The `/silences` creation modal this PR says it reuses (`Silences.templ`, wider `sm:max-w-4xl` card) doesn't hit this because it has ~3x the available width for the same row shape — the pattern needs a width-independent fix here, not a copy of that container's sizing.

SoulKyu: @SoulKyu This has been fixed in 6e15040.

Added flex-wrap to the matcher row's class list at internal/webui/templates/components/modal_components.templ:1816 so name/op/value/remove wrap onto a second line instead of overflowing the sm:max-w-lg card, restoring the remove button's clickable area.




SoulKyu:
**Looper fixer round complete** — 6e15040

- ✅ Review comment on `internal/webui/templates/components/modal_components.templ` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/139#discussion_r3683743980)
  - Added flex-wrap to the matcher row's class list at internal/webui/templates/components/modal_components.templ:1816 so name/op/value/remove wrap onto a second line instead of overflowing the sm:max-w-lg card, restoring the remove button's clickable area.


SoulKyu:
## 🧪 QA Report

Re-test after the rebase. Full stack rebuilt and booted from this SHA (`make test`: `templ generate` → tailwind → `docker-build-all` → `docker compose up`; postgres + fake-alertmanager + backend + webui all healthy). Every criterion re-exercised against the live app with `curl` and a real Chromium session (Playwright), then attacked.

**The rebase changed nothing in this PR's own files.** `git diff 8c0c781..cb96d5f` touches only `internal/webui/client/backend_client.go` + its test — that is main's #133 landing underneath. Every file this PR owns is byte-identical to the previously-QA'd head, and the blocking defect reproduces unchanged.

|---|---|---|
| Modal lists every matcher, pre-filled from labels minus `__name__`/`__tmp_` | ✅ | Injected an alert into the fake AM carrying literal `__name__=up` and `__tmp_=scratch`: labels `[__name__, __tmp_, alertname, evil, instance, severity, team]` → rows `[alertname, evil, instance, severity, team]`. Disclosure collapsed by default. |
| A matcher can be edited (name and value), removed, and added | ❌ | Edit ✅ (typed into name and value inputs, state + preview followed). Add ✅ (`+ Add matcher` → rows 5→6). **Remove is unreachable by mouse at every viewport ≥640px** — see below. |
| Each matcher switchable between `=`, `!=`, `=~`, `!~` | ✅ | Drove the real `<select>`; payload mapping per option: `=`→`{isRegex:false,isEqual:true}`, `!=`→`{false,false}`, `=~`→`{true,true}`, `!~`→`{true,false}`. |
| Invalid regexp → HTTP 400 identifying the row, before any Alertmanager call | ✅ | `[unclosed` at index 1 → `400 {"error":"invalid matchers","matcherErrors":[{"index":1,"error":"invalid regex: … missing closing ]: \`[unclosed)$\`"}]}`. AM silence count unchanged across all reject cases. In the UI the row goes red and the message renders under it (`matcherErrors=[{"index":0,…}]`). |
| Explicit empty matcher list rejected | ✅ | `"silenceMatchers": []` → `400 "at least one matcher is required"`. UI: deleting all rows sets `silenceMatchersInvalid()` true. |
| Explicit matchers + more than one fingerprint rejected server-side | ✅ | 2 fingerprints → `400 "explicit silence matchers require exactly one alert fingerprint"`; group silence (0 fingerprints) → same 400. |
| Preview count from `POST /silences/preview`, updates as matchers are edited | ✅ | Debounced POST on every edit. Internally exact: `severity=critical` (11) + `severity!=critical` (20) = 31 = `severity=~.*` (31) = full cache. Count tracked edits live (1 → 0 → 31 → 11 → 20). |
| Untouched submission = semantically identical silence | ✅ | UI submit untouched sent exactly the label-derived set (5 matchers, all `isRegex:false,isEqual:true`, matching the alert's labels 1:1). Fallback path (field omitted **and** explicit `null`) produced the identical set in Alertmanager. |
| Audit comment includes the matcher count | ✅ | Explicit 2 matchers → `🔇 Alert silenced for 1h: qa139-audit-explicit (2 matcher(s))`; fallback → `… qa139-fallback (5 matcher(s))`. |
| Bulk and group silences unaffected | ✅ | Both modals render 0 matcher inputs, `silenceMatcherRows=[]`, payloads carry no `silenceMatchers` key; server rejects explicit matchers on those paths regardless of the client. |
| Only `.templ` sources edited; generated files from `make webui-templates` | ✅ | `templ generate` with the go.mod-pinned v0.3.906 at this SHA → **zero** diff across 144 updates. `go test ./internal/webui/...` green. |

### ❌ The remove (✕) button of every matcher row renders outside the modal card — clicking it closes the modal instead of removing the matcher

`modal_components.templ:1765` — the `div.sm:flex.sm:items-start` wrapper has no `min-w-0`, so the `flex-1 min-w-0` on the matcher inputs is defeated. The row's min-content width (~542px) exceeds the `sm:max-w-lg` card's 510px content box, and the card carries `overflow-hidden` (`:1755`), so the excess is clipped with no scrollbar.

Real mouse click on the ✕'s visual centre, seven viewport widths, a reason typed first to show what is lost:

```
W=1920 click(1222,236) btn[1209..1235] cardRight=1216 scrollW/clientW=542/510 hit=DIV.fixed.inset-0 | rows 5->5 modalOpen true->false
W=1440 click( 982,236) btn[ 969.. 995] cardRight= 976 scrollW/clientW=542/510 hit=DIV.fixed.inset-0 | rows 5->5 modalOpen true->false
W=1280 click( 902,236) btn[ 889.. 915] cardRight= 896 scrollW/clientW=542/510 hit=DIV.fixed.inset-0 | rows 5->5 modalOpen true->false
W=1024 click( 774,236) btn[ 761.. 787] cardRight= 768 scrollW/clientW=542/510 hit=DIV.fixed.inset-0 | rows 5->5 modalOpen true->false
W= 768 click( 646,236) btn[ 633.. 659] cardRight= 640 scrollW/clientW=542/510 hit=DIV.fixed.inset-0 | rows 5->5 modalOpen true->false
W= 640 click( 582,236) btn[ 569.. 595] cardRight= 576 scrollW/clientW=542/510 hit=DIV.fixed.inset-0 | rows 5->5 modalOpen true->false
W= 390 click( 324,300) btn[ 311.. 337] cardRight= 374 scrollW/clientW=356/356 hit=path      | rows 5->4 modalOpen true->true   <- only mobile works
```

`document.elementFromPoint` on the button centre returns `DIV.fixed.inset-0` — the modal **backdrop**, whose `@click="showSilenceModal = false"` fires. The matcher is not removed, the modal closes, and the operator's edits and typed reason are discarded: reopening the modal shows `silenceReason === ""`. Only a ~7px invisible sliver (x 889–896 at 1280) actually hits the button.

The handler itself is fine — focusing the button and pressing Enter removes the row (`rows 5->4`, modal stays open, `tabIndex 0`). The defect is purely that no mouse user can reach it. In the attached-style screenshot at 1280 no ✕ is visible on any of the 5 rows.

Same clipping (`scrollWidth 542` vs `clientWidth 510`) pushes the value inputs (`backup-9.example.co` — truncated), the duration `<select>`, the **required** Silence Reason textarea and the info banner past the card edge whenever "Matchers (N)" is expanded. Expanding that disclosure is this feature's primary interaction.

For comparison, the `/silences` creation modal this PR reuses (#130, `Silences.templ:218`) puts the same button inside its card and hit-tests correctly — the pattern was copied into a container 1.5× narrower without re-checking the fit.

Fix stays small: `min-w-0` on the `sm:flex sm:items-start` wrapper (or `flex-wrap` on the matcher row).

### Adversarial pass

Everything below ran against the live app at this SHA.

- **Boundary & empty** — `silenceMatchers: []` → 400 `at least one matcher is required`; all rows deleted in the UI → `silenceMatchersInvalid()` true; single matcher → accepted; a 5-label alert → exactly 5 rows; preview of an empty set → 0, not "everything".
- **Malformed & hostile** — `[unclosed` as a regex at index 1 → 400 with the correct row index, no AM call (silence count unchanged); `isRegex:"yes"` (wrong type) → `400 Invalid request format`; nonexistent fingerprint + matchers → `1 of 1 actions failed`, no silence created; typed `a/b?c*d` straight into a value input → no crash, no console error, modal stayed open.
- **Attacker-influenceable label** — injected an alert whose `evil` label value is ``. It reaches the matcher editor as an input *value* only: no `img[src=x]` in the DOM, no JS dialog, preview still computed (3). Reserved `__name__`/`__tmp_` stripped from the rows on that same alert.
- **`isEqual` omitted** (the dangerous default) — `{"name":"team","value":"platform","isRegex":false}` decoded to `isEqual:true` on the created Alertmanager silence, not a silently negated `!=`.
- **`silenceMatchers: null` explicitly** — falls back to the label-derived set rather than tripping the empty-list guard.
- **Error / unavailable path** — forced `POST /silences/preview` to 500 → `silencePreviewCount=null` and a visible red `boom`. No silent success, no stale count pushed as valid. (Nit, unchanged: if a count was already rendered and a *later* edit's preview fails, the old count stays on screen next to the error — `silencePreviewCount = null` in the catch would be cleaner.)
- **Auth & scope** — unauthenticated `POST /bulk-action` with matchers → `401 Authentication required`. Explicit matchers rejected server-side for >1 fingerprint and for the group path regardless of what the client sends.
- **Adjacent flows** — bulk and group modals show no matcher editor; the `/silences` page still renders its silence table and its own modal; typing `/` inside a matcher value input does not fire a dashboard hotkey and does not steal focus (`focus=value`, modal open).
- **Persisted / pre-existing state** — ran against the retained `postgres-data` volume and silences created before this change (`Test silence 1` still listed and matched correctly); the fallback path produces matcher sets identical to `main`'s.
- **Removal specifically** — `:key="idx"` on the `x-for` is *not* the problem: removing a middle row programmatically keeps Alpine state and DOM inputs in sync, and keyboard Enter on the focused button works. The break is hit-testing only.

**Out of scope, pre-existing (not this PR):** every webui-created silence sends `"id": ""` (`models.Silence.ID` lacks `omitempty`), so the fake Alertmanager treats each new one as an update and overwrites the previous webui silence — reproduced identically on the fallback path, which is main's code. Also `compileDraftMatchers` rejects empty matcher names but not invalid Prometheus label names.

**Verdict: FAIL** (1 of 11 criteria) — unchanged by the rebase; fixer summoned.

SoulKyu:
## 🧪 QA Report

Re-test after the fixer round. Full stack rebuilt and booted from this SHA (`make test`: `templ generate` → tailwind → `docker-build-all` → `docker compose up`; postgres + fake-alertmanager + backend + webui all healthy). Every criterion re-exercised against the live app with `curl` and a real Chromium session (Playwright), then attacked.

`git diff cb96d5f..6e15040` is a single line: `flex-wrap` added to the matcher row's class list (`modal_components.templ:1816`). **That is the exact fix the previous report named, and it holds** — the card no longer overflows and every ✕ is reachable by mouse from 320px to 1920px.

|---|---|---|
| Modal lists every matcher, pre-filled from labels minus `__name__`/`__tmp_` | ✅ | Alert injected with literal `__name__=up` + `__tmp_=scratch` + 6 real labels → rows `[alertname, evil, instance, job, severity, team]`, reserved pair stripped. Disclosure collapsed by default, `Matchers (6)`. |
| A matcher can be edited (name and value), removed, and added | ✅ | **Was the blocker, now fixed.** Edit: typed into name+value, state and preview followed. Add: `+ Add matcher` 6→7. Remove: real `page.mouse.click()` on each ✕'s visual centre, **all 6 rows removed one by one at every one of 10 viewports (1920/1440/1280/1024/768/640/480/390/360/320)** — modal stayed open and the typed reason survived each time. Card `scrollWidth == clientWidth` at all widths (was 542/510). |
| Each matcher switchable between `=`, `!=`, `=~`, `!~` | ✅ | Drove the real `<select>`: `=`→`{isRegex:false,isEqual:true}` (preview 1), `!=`→`{false,false}` (0), `=~`→`{true,true}` (1), `!~`→`{true,false}` (0). |
| Invalid regexp → HTTP 400 identifying the row, before any Alertmanager call | ✅ | API: `[unclosed` at index 1 → `400 {"error":"invalid matchers","matcherErrors":[{"index":1,"error":"invalid regex: … missing closing ]: \`[unclosed)$\`"}]}`; AM silence count unchanged across every reject case. UI: same value typed into row 6 → `matcherErrors=[{"index":6,…}]`, 2 red-bordered inputs on that row only, message rendered beneath it. |
| Explicit empty matcher list rejected | ✅ | `"silenceMatchers": []` → `400 "at least one matcher is required"`. UI: deleting all 6 rows → `silenceMatchersInvalid()` true and the **Silence** button goes `disabled`. |
| Explicit matchers + more than one fingerprint rejected server-side | ✅ | 2 fingerprints → `400 "explicit silence matchers require exactly one alert fingerprint"`; group path (0 fingerprints + `groupNames`) → same 400. |
| Preview count from `POST /silences/preview`, updates as matchers are edited | ✅ | 10 debounced POSTs across the edit session, all 200. Count tracked live (1 → 0 → 1 → 0 → 0 when emptied). |
| Untouched submission = semantically identical silence | ✅ | Real UI submit without touching the list sent all 6 labels as `{isRegex:false,isEqual:true}`; the resulting Alertmanager silence is matcher-for-matcher identical to the fallback path (field omitted **and** explicit `null`, both tested). |
| Audit comment includes the matcher count | ✅ | `🔇 Alert silenced for 1h: qa139b-explicit (2 matcher(s))`, `… qa139b-fallback-omit (6 matcher(s))`, `… qa139b-fallback-null (6 matcher(s))`. |
| Bulk and group silences unaffected | ✅ | Both modals: `silenceMatcherRows=[]`, 0 visible `label name` inputs, 0 visible `Matchers (N)` disclosure; server rejects explicit matchers on those paths regardless of client. |
| Only `.templ` sources edited; generated files from `make webui-templates` | ✅ | `templ generate` with the go.mod-pinned v0.3.906 at this SHA → **zero** `git status` diff. `go test ./internal/webui/...` → 142 passed, 11 packages. |

### Adversarial pass

- **The regression that failed last round** — real mouse clicks (not `.click()` dispatch) on every row's ✕ at 10 viewports 320→1920, each with the row scrolled into view and `document.elementFromPoint` asserted to land on the button, not the backdrop. 60/60 removals succeeded; `elementFromPoint` returned the button's `path` every time, never `DIV.fixed.inset-0`. Modal never closed, `silenceReason` never wiped.
- **Layout stress on the new `flex-wrap`** — 200-char label name + 300-char value in one row → `scrollW/clientW = 510/510`, all 7 ✕ still hittable. 30 matcher rows at 1440/640/390/360/320 → no overflow at any width (`286/286` at 320). Forced the row to genuinely wrap (`html{font-size:24px}` and `32px`, i.e. a browser large-font a11y setting) at 1280 and 390: still no overflow, ✕ still hit-tested to the button, correct row removed, modal open. The wrap path the fix relies on works — it isn't just never exercised.
- **Boundary & empty** — `silenceMatchers: []` → 400; all rows deleted in the UI → `Silence` button disabled; single matcher → accepted; matcher with empty `name` → `400 matcherErrors[{"index":0,"error":"matcher name is required"}]`; preview of an empty set → 0, not "everything".
- **Malformed & hostile** — `[unclosed` regex → 400 with correct index, AM silence count unchanged (no AM call); `isRegex:"yes"` (wrong type) → `400 Invalid request format`; typed `/etc/passwd` into a value input → stored verbatim, no crash.
- **Attacker-influenceable label** — alert carrying `evil=` reaches the editor as an input *value* only: no `img[src=x]` in the DOM, no JS dialog fired (dialog handler armed for the whole run), preview still computed. Reserved `__name__`/`__tmp_` stripped from that same alert's rows.
- **`isEqual` omitted** (the dangerous default) — `{"name":"team","value":"platform","isRegex":false}` landed as `isEqual:true` on the created silence, not a silently negated `!=`.
- **`silenceMatchers: null` explicitly** — falls back to the label-derived set instead of tripping the empty-list guard.
- **Error / unavailable path** — forced `POST /silences/preview` to 500 → `silencePreviewCount=null`, `silencePreviewError="boom"` rendered in red, the "Will match N" span hidden. No silent success, no fabricated count.
- **Auth & scope** — unauthenticated `POST /bulk-action` with matchers → `401 Authentication required`. Explicit matchers rejected server-side for >1 fingerprint and for the group path regardless of client input.
- **Adjacent flows** — bulk and group modals render no matcher editor; typing `/` inside a matcher value input keeps focus in the input, fires no dashboard hotkey, modal stays open; `/silences` page still renders its table (2 rows) and its own modal.
- **Persisted / pre-existing state** — ran against the retained `postgres-data` volume; the pre-existing `Test silence 1` is still listed and intact after all of the above.

**Out of scope, pre-existing (not this PR, unchanged):** every webui-created silence sends `"id": ""` (`models.Silence.ID` lacks `omitempty`), so the fake Alertmanager treats each new one as an update and overwrites the previous webui silence — reproduces identically on the fallback path, which is main's code. `compileDraftMatchers` rejects empty matcher names but not otherwise-invalid Prometheus label names.

**Nit (not a criterion, carried over):** if a count is already rendered and a *later* edit's preview call fails, the stale count stays on screen next to the error — `silencePreviewCount = null` in the catch would be cleaner. Not reachable as a wrong-outcome path in the flows tested (a first failure correctly shows no count).

**Verdict: PASS**

Issue #95, opened 2026-07-25
fix(dashboard): browser Back/Forward is desynced from the alert modal — no popstate handler, and closing the modal wipes filter URL params


SoulKyu:
## Problem

The alert-details modal manipulates browser history but nothing ever listens to it:

- Opening the modal pushes a history entry: `window.history.pushState({ alertId }, '', '/dashboard/alert/<fp>')` — `internal/webui/templates/scripts/dashboard_modal.templ:13-16`, and the same pattern for resolved alerts in `internal/webui/templates/scripts/dashboard_resolved_alerts.templ:253-257`.
- Closing the modal pushes a **second** entry: `window.history.pushState({}, '', '/dashboard')` — `dashboard_modal.templ:54-56`.
- There is **no `popstate` listener anywhere** in `internal/webui/` (`rg popstate` returns zero hits outside generated files). `checkAlertFromURL()` (`dashboard_utilities.templ:43`) only runs once at page init.

Two concrete failure paths:

1. Open an alert modal, press browser **Back** → the URL returns to `/dashboard` but the modal stays open. Press Back again and you leave the page entirely while a modal is still on screen.
2. Close the modal with the X, press **Back** → the URL becomes `/dashboard/alert/<fp>` but nothing opens. Every open/close cycle deposits two dead history entries, so an on-call engineer who inspected 10 alerts needs ~20 Back presses to actually leave the dashboard.

Bonus damage: the dashboard maintains shareable filter state in the query string via `replaceState` (`dashboard_utilities.templ:6-26`), but `closeAlertModal()` pushes the bare path `'/dashboard'` — so opening and closing any alert silently **wipes `?severities=…&teams=…` from the URL**, breaking the copy-the-URL-to-share workflow the filter code deliberately supports.

## Impact

Every dashboard user, every day: the Back button — the most reflexive navigation gesture in a browser — misbehaves on the app's primary operational page, and the shareable filtered URL (a real on-call workflow: paste the filtered view in Slack) is destroyed by the routine act of looking at an alert.

## Proposed approach

[...]

PR #142, opened 2026-07-30
fix(webui): silent failure when an alert deep link is unknown to both live cache and resolved_alerts

### LINKED ISSUE #140: fix(webui): silent failure when an alert deep link is unknown to both live cache and resolved_alerts
## Problem

Deep links `/dashboard/alert/<fingerprint>` are **not** live-cache-only, contrary to what this issue originally claimed. The handler chain `GetAlertDetails` → `GetAlertByFingerprint` → `GetAlert` (`internal/webui/services/alert_cache.go:806-830`) falls back on a live-cache miss to `backendClient.GetResolvedAlert`, i.e. the backend's `resolved_alerts` table — written on every resolve with a TTL of `resolvedRetentionDays` (default **90 days**, `alert_cache.go:106-108`; row created at `alert_cache.go:1139`; `expires_at` filter at `internal/backend/database/gorm_db.go:641-644`). So with a connected backend, resolved-alert copy-links (#137), activity-feed `resolve` rows (#136) and stale Slack links all work today.

The real gap is the case where the fingerprint is unknown to **both** the live cache and `resolved_alerts`:

- row expired past `resolvedRetentionDays` (stale link older than retention),
- resolve happened while the webui was down, so no `resolved_alerts` row was ever written (`alert_cache.go:1139` only runs when the webui observes the resolve),
- backend disconnected — the fallback is gated on `IsConnected()` (`alert_cache.go:819`), so the handler 404s even for alerts that have a row.

In all of these, `showAlertDetails`'s failure branch (`internal/webui/templates/scripts/dashboard_modal.templ:80-88`, both the `!result.success` and `catch` paths) does `console.error` + `closeAlertModal()`, which silently rewrites the URL back to `/dashboard` (`dashboard_modal.templ:104-111`). Zero user-visible feedback: the recipient lands on a plain dashboard and cannot tell "link expired" from "app is broken".

## Impact

Narrow but real: links older than the retention window, resolves missed during webui downtime, and any deep link opened while the backend is down fail with no message at all. The failure mode is indistinguishable from a broken app, which is exactly the wrong signal during or after an incident.

## Proposed approach

In the failure branch of `showAlertDetails` (`dashboard_modal.templ:80-88`), surface a user-visible notice before closing the modal — e.g. a small Alpine toast/banner on the dashboard: "Alert not found — it may have expired or been purged." Keep the existing URL-rewrite/popstate behaviour unchanged. There is no shared dashboard toast today (only a page-local one in `pages/Profile.templ:241-262`), so this is a few lines of Alpine state + markup, not a component library.

Optional follow-up, only if the purged-from-`resolved_alerts`-but-still-in-statistics window is shown to matter in practice: fall back to `/api/v1/statistics/alert/<fp>` before giving up. Not required for this issue.

Edit the `.templ` sources and regenerate with `make webui-templates` (never touch `*_templ.go`).

## Acceptance criteria

- [ ] Opening `/dashboard/alert/<fp>` for a fingerprint unknown to both the live cache and `resolved_alerts` shows a user-visible message instead of a silent URL rewrite (verify with a bogus fingerprint).
- [ ] The same feedback appears on the fetch-error path (backend/webui API unreachable).
- [ ] Back/Forward behaviour from #95/#100 (popstate sync, no history stacking) is preserved.

SoulKyu:
## Summary
- `showAlertDetails`'s failure branches (`!result.success` and the fetch `catch`) only did `console.error` before calling `closeAlertModal()`, which silently rewrites the URL back to `/dashboard`. A deep link `/dashboard/alert/<fp>` unknown to both the live cache and `resolved_alerts` (expired past retention, resolved while webui was down, or backend disconnected) gave the recipient zero feedback — indistinguishable from a broken app.
- Both branches now call the existing shared `actionError` banner (`showActionError`, already wired up in `NewDashboard.templ` / `dashboard_actions.templ` for bulk-action failures) with a distinct message per failure mode: "Alert not found — it may have expired or been purged." for the API's `success: false` response, and "Unable to load alert — check your connection and try again." for the fetch-error path.
- No new component: reused the dashboard's existing error banner instead of adding a page-local toast, since it already renders, auto-dismisses (8s), and is dismissible.
- Existing URL-rewrite/popstate behaviour (`closeAlertModal`, `syncModalWithLocation`) is untouched.

## Test plan
- [x] `go build ./...`
- [x] Regenerated templates with the go.mod-pinned `templ` v0.3.906 (`templ generate`), diff limited to `dashboard_modal.templ`/`dashboard_modal_templ.go`
- [ ] Manual: open `/dashboard/alert/<bogus-fingerprint>` and confirm the error banner appears instead of a silent redirect
- [ ] Manual: simulate a fetch failure (e.g. stop the backend) and confirm the banner appears on that path too
- [ ] Manual: confirm Back/Forward (#95/#100) still works with no history stacking

Closes #140


SoulKyu (commented):
@SoulKyu Reviewed the fix for the silent-failure deep-link case (closes #140).

Verified:
- Both failure branches in `showAlertDetails` (`internal/webui/templates/scripts/dashboard_modal.templ`) now call the existing `showActionError` banner instead of only logging to the console, with a distinct message for the API `success: false` case vs. the fetch/network-error case.
- `showActionError` comes from `dashboardActionsMixin`, which is `Object.assign`'d onto the same Alpine component as `dashboardModalMixin` during `init()` in `dashboard_core.templ`, so `this.showActionError` resolves correctly at call time — no wiring issue there.
- The error banner div lives at the top level of `NewDashboard.templ`, not inside the modal markup, so it stays visible through `closeAlertModal()`'s URL rewrite/history pop, and `closeAlertModal` never touches `actionError`, so the message isn't wiped out before the user reads it.
- Regenerated `dashboard_modal_templ.go` locally with the go.mod-pinned `templ` v0.3.906 and diffed against the committed file — no drift, so the generated code matches the template source.
- `go build ./...` passes.

This is a small, well-contained fix that reuses existing shared infrastructure instead of adding a new component — nice restraint there. No actionable findings, this looks good to merge. Thanks for the clean writeup and test plan in the description!



SoulKyu:
## 🧪 QA Report

Env: `make docker-build-all` + `docker compose up -d` (postgres / fake-alertmanager / backend / webui on :8081), real Chromium via Playwright, user `qauser` registered through `/api/v1/auth/register`. 14 live alerts in cache.

|---|---|---|
| Deep link `/dashboard/alert/<fp>` unknown to live cache **and** `resolved_alerts` shows a user-visible message instead of a silent URL rewrite | ✅ | `GET /api/v1/dashboard/alert/deadbeef…` → `404 {"success":false,"error":"Alert not found"}`; browser lands on the deep link → red banner **“Alert not found — it may have expired or been purged.”**, `showAlertModal=false`, `location.pathname` rewritten to `/dashboard` |
| Same feedback on the fetch-error path (backend/webui API unreachable) | ✅ | Dashboard loaded, then `docker stop notificator-webui`, then clicked an alert row → banner **“Unable to load alert — check your connection and try again.”**, modal closed, `alertDetailsLoading=false` (no stuck spinner) |
| Back/Forward behaviour from #95/#100 (popstate sync, no history stacking) preserved | ✅ | Click alert → `/dashboard/alert/ac3c5276…`, `history.length` 3 → 3; Back → modal closed, `/dashboard`, length still 3; Forward → modal reopens with details. After a *failed* deep link, `history.length` stays at 3 (replaceState) and Back returns to the previous page instead of re-entering the alert URL |

**Adversarial pass**

*Criterion 1 — boundary / malformed / hostile fingerprints* (each opened as a real deep link, banner text + `location.pathname` + XSS probe checked):
- 1-char `x`, 2000-char `aaa…`, `%20`, `ab%00cd`, `a%3Fb%26c`, `💥é—` (unicode/em-dash) → banner shown every time, URL rewritten to `/dashboard`, modal closed.
- XSS: `` as the fingerprint → banner rendered via `x-text`, payload escaped, `window.__xss` never set. The banner message is a static string, so no attacker-influenced text reaches the DOM.
- Extra-path-segment / traversal variants (`..%2f..%2fetc%2fpasswd`, `%3Cscript%3E…%3C/script%3E`) never reach Alpine — gin returns its bare 404 page (`GET` → 404). Pre-existing routing behaviour, untouched by this PR.
- Two failures back to back (`aaaa…` then `bbbb…` via `showAlertDetails`) → banner correct both times, no timer/state leak, URL stays `/dashboard`.
- Chained open (notification-style): valid alert modal open (`alertModalPushedHistory=true`), then `showAlertDetails('deadbeef…')` on a purged fp → banner shown **and** `closeAlertModal` pops our own entry (`history.length` 6 → 6, path `/dashboard`). No off-site navigation, no stacked entry.
- Banner lifetime vs the 5 s auto-refresh: still visible at 6 s, auto-dismissed by 10 s (8 s timer). The refresh loop does not wipe it.
- Auth/scope: anonymous deep link → `/login` (never a blank dashboard); session cookie cleared mid-session then alert clicked → 401 → interceptor redirect to `/login`.
- False-positive check (persisted state): resolved a live alert through the UI (`resolveAlert`), then deep-linked it in a fresh tab → modal opens with details, **no** error banner. The new branch does not fire for legitimately resolvable alerts.

*Criterion 2 — error/unavailable path:* webui container stopped mid-flow (real `fetch` rejection, not a stub) → banner, modal closed, no wiped-state-as-success. Backend gRPC stopped (`docker stop notificator-backend`) → the deep link never reaches the JS: `RedirectIfNotAuth` bounces the request to `/login` before the dashboard renders. That is pre-existing auth-middleware behaviour, not a silent wipe, and outside this diff — but it means the issue’s third bullet (“backend disconnected”) surfaces as a login redirect rather than the new banner.

*Criterion 3 — adjacent flows:* forward/back on a *successful* alert still restores the modal with its details and the query string; the failed-deep-link path uses `replaceState`, so Back does not bounce the user back into the failing URL.

**Non-blocking finding (out of the issue’s stated scope, reproduced live):** `showResolvedAlertDetails` (`internal/webui/templates/scripts/dashboard_resolved_alerts.templ:310-317`) still has the exact bug #140 describes — both its `!statsResult.success` and `catch` branches do `console.error` + `closeAlertModal()` with no banner. Reproduced: Resolved view loaded (1 row), row opened OK → `/dashboard/alert/f9220cc9…`; then `docker stop notificator-webui` and the same row clicked again → URL pushed and silently rewritten back to `/dashboard`, **zero** user feedback. Reachable from the Resolved view and from Forward/popstate onto a resolved deep link (`dashboard_modal.templ:44-47` routes there when the fp is in `resolvedAlerts`). The issue scoped the fix to `showAlertDetails`, so this is not counted against the criteria — worth a follow-up issue, the one-line `showActionError(...)` fits the same way.

Teardown: `docker compose down -v`.

**Verdict: PASS**

PR #143, opened 2026-07-30
fix(security): stop shipping a public encryption key in docker-compose.yml

### LINKED ISSUE #141: fix(security): docker-compose.yml ships a hardcoded NOTIFICATOR_ENCRYPTION_KEY, quietly undoing #110's guarantee
## Problem

`docker-compose.yml:62` hardcodes a literal, valid AES-256 key:

```
- NOTIFICATOR_ENCRYPTION_KEY=51f64f844ed0a772d10c2184b02a682ab6b5348f50181a9faaaecfb1446c8dcf
```

Commit 46b9db6 (2 days ago) replaced the previous `${NOTIFICATOR_ENCRYPTION_KEY}` passthrough with this value — understandably, because after #110/PR #113 the backend refuses to boot without a valid 64-hex key (`ValidateEncryptionKey`, `internal/backend/database/sentry_db.go:20-35`), which made `make test` require an exported env var.

But the committed value is now a **publicly-known key** (it lives in git history forever), and it passes the #110 validation perfectly: it is 64 lowercase hex chars and looks exactly like the output of `openssl rand -hex 32`. Nothing distinguishes it from a real secret — which invites copy-paste into real deployments, and defeats the entire point of #110 (Sentry personal API tokens encrypted with a key an attacker cannot know). The quickstart documents `docker-compose up -d` as a way to run the whole stack, so this file is a deployment template, not just CI fixture.

Two concrete failure scenarios:
1. Someone stands up a small team instance from `docker-compose.yml` (the documented path), users store Sentry personal tokens, and anyone who ever obtains a DB dump can decrypt every token with the key from the public repo — while the operator believes #110 protects them, because "a real key is required and set".
2. If this key was generated once and reused in any real environment, it is now burned and must be rotated.

## Impact

Silent reversal of a shipped security fix for the most likely small-deployment path. The #110 boot check reports success, so operators get a false sense of safety. Cost of the fix is near zero; cost of the failure is every stored Sentry token.

## Proposed approach

Keep the zero-friction `make test` workflow, but stop shipping a production-looking secret:

- Change the line to `NOTIFICATOR_ENCRYPTION_KEY=${NOTIFICATOR_ENCRYPTION_KEY:-<dev-key>}` so it stays overridable, and make the dev default *visibly* fake while still 64 hex chars, e.g. `badbadbadbadbadbadbadbadbadbadbadbadbadbadbadbadbadbadbadbadbad0` — greppable, obviously not a secret.
- Optionally: have the backend log a loud startup warning when the key equals the known dev default (cheap comparison in `ValidateEncryptionKey`), so a copy-pasted compose file announces itself in prod logs.
- Add one line next to it in `docker-compose.yml`: `# dev-only key — generate your own: openssl rand -hex 32`.
- If `51f64f…c8dcf` was ever used outside this compose file, rotate it there.

## Acceptance criteria

- [ ] `docker-compose.yml` no longer contains a random-looking committed key; the dev default is self-evidently non-secret and overridable via env.
- [ ] `make test` / `docker-compose up -d` still boots without any manual env setup.
- [ ] (If warning implemented) backend logs a clear warning when running with the dev default key.
- [ ] Helm chart values stay placeholder-only (already the case, `charts/notificator-app/values.yaml:61`).

SoulKyu:
## Summary

- `docker-compose.yml` shipped a literal, valid 64-hex-char AES-256 key for `NOTIFICATOR_ENCRYPTION_KEY` (added in 46b9db6 to satisfy the #110/#113 boot-time validation). Because it's committed to git history, that key is public — copying the compose file into a real deployment gives a false sense of security while Sentry tokens end up encrypted with a known key.
- The line now reads `NOTIFICATOR_ENCRYPTION_KEY=${NOTIFICATOR_ENCRYPTION_KEY:-badbad...badb}` — a 64-hex-char default that's obviously fake (repeating `bad` pattern, not random-looking), still overridable via env, with a comment pointing at `openssl rand -hex 32` for real deployments.
- `ValidateEncryptionKey` (`internal/backend/database/sentry_db.go`) now logs a loud `WARNING` at startup when the running key matches this known dev default, so a copy-pasted compose file announces itself in prod logs instead of silently "passing" the #110 check.
- Helm chart values were already placeholder-only (`charts/notificator-app/values.yaml:61`) — no change needed there.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./internal/backend/database/...`
- [x] `go test ./internal/backend/database/...` (existing `TestValidateEncryptionKey`/`TestEncryptDecryptRoundTrip` + new `TestValidateEncryptionKeyWarnsOnDevDefault`)
- [x] `docker compose -f docker-compose.yml config` resolves `NOTIFICATOR_ENCRYPTION_KEY` to the 64-char dev default when unset, confirming `make test` / `docker-compose up -d` still boot with zero manual env setup

Closes #141

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -59,7 +59,8 @@ services:
       - NOTIFICATOR_BACKEND_DATABASE_USER=notificator
       - NOTIFICATOR_BACKEND_DATABASE_PASSWORD=notificator_password
       - NOTIFICATOR_BACKEND_DATABASE_SSL_MODE=disable
-      - NOTIFICATOR_ENCRYPTION_KEY=51f64f844ed0a772d10c2184b02a682ab6b5348f50181a9faaaecfb1446c8dcf
+      # dev-only key — generate your own: openssl rand -hex 32
+      - NOTIFICATOR_ENCRYPTION_KEY=${NOTIFICATOR_ENCRYPTION_KEY:-badbadbadbadbadbadbadbadbadbadbadbadbadbadbadbadbadbadbadbadbadb}
 
       # Alertmanager configuration
       - NOTIFICATOR_ALERTMANAGERS_0_NAME=Fake Alertmanager
--- a/internal/backend/database/sentry_db.go
+++ b/internal/backend/database/sentry_db.go
@@ -8,6 +8,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"io"
+	"log"
 	"os"
 
 	"notificator/internal/backend/models"
@@ -17,12 +18,24 @@ import (
 // used to encrypt Sentry personal tokens at rest.
 const EncryptionKeyEnvVar = "NOTIFICATOR_ENCRYPTION_KEY"
 
+// devDefaultEncryptionKey is the placeholder key shipped in docker-compose.yml
+// for zero-friction local dev/test. It must never be used in production.
+const devDefaultEncryptionKey = "badbadbadbadbadbadbadbadbadbadbadbadbadbadbadbadbadbadbadbadbadb"
+
 // ValidateEncryptionKey checks that NOTIFICATOR_ENCRYPTION_KEY is set to 64
 // lowercase hex characters (32 raw bytes), returning an error naming the
-// variable and the generation command otherwise.
+// variable and the generation command otherwise. It also warns loudly when
+// the key matches the known dev default, since that means a deployment is
+// running with a publicly-known, non-secret encryption key.
 func ValidateEncryptionKey() error {
 	_, err := encryptionKeyFromEnv()
-	return err
+	if err != nil {
+		return err
+	}
+	if os.Getenv(EncryptionKeyEnvVar) == devDefaultEncryptionKey {
+		log.Printf("WARNING: %s is set to the known dev/test default from docker-compose.yml. This key is public and MUST NOT be used to protect real secrets — generate your own with `openssl rand -hex 32`.", EncryptionKeyEnvVar)
+	}
+	return nil
 }
 
 func encryptionKeyFromEnv() ([]byte, error) {
--- a/internal/backend/database/sentry_db_test.go
+++ b/internal/backend/database/sentry_db_test.go
@@ -1,6 +1,12 @@
 package database
 
-import "testing"
+import (
+	"bytes"
+	"log"
+	"os"
+	"strings"
+	"testing"
+)
 
 func TestValidateEncryptionKey(t *testing.T) {
 	cases := map[string]struct {
@@ -26,6 +32,21 @@ func TestValidateEncryptionKey(t *testing.T) {
 				t.Fatalf("expected no error for key %q, got %v", tc.key, err)
 			}
 		})
+	}
+}
+
+func TestValidateEncryptionKeyWarnsOnDevDefault(t *testing.T) {
+	t.Setenv(EncryptionKeyEnvVar, devDefaultEncryptionKey)
+
+	var buf bytes.Buffer
+	log.SetOutput(&buf)
+	defer log.SetOutput(os.Stderr)
+
+	if err := ValidateEncryptionKey(); err != nil {
+		t.Fatalf("expected no error for dev default key, got %v", err)
+	}
+	if !strings.Contains(buf.String(), "WARNING") {
+		t.Fatalf("expected a startup warning for the dev default key, got log output: %q", buf.String())
 	}
 }
 


SoulKyu (commented):
@SoulKyu Thanks for closing out #141! I verified the fix against the diff:

- `docker-compose.yml` no longer ships the real, git-history-exposed AES key. The new default `NOTIFICATOR_ENCRYPTION_KEY=${NOTIFICATOR_ENCRYPTION_KEY:-badbad...badb}` is obviously a placeholder (repeating `bad` pattern), still 64 valid lowercase-hex characters so it passes the existing `encryptionKeyFromEnv` length/hex checks, and remains overridable via env — `docker compose config` / `make test` keep booting with zero manual setup.
- `ValidateEncryptionKey` in `internal/backend/database/sentry_db.go` now does an exact-match check against the new `devDefaultEncryptionKey` constant and logs a loud `WARNING` (via the standard `log` package, which defaults to stderr, so it will actually surface in container logs) when a deployment is running with this known dev key — while still returning no error, so local/dev flows aren't broken.
- Confirmed the placeholder string is byte-for-byte identical between `docker-compose.yml` and the Go constant, and traced the call path (`server.go:49` calls `ValidateEncryptionKey` before DB init) so the warning fires at real startup.
- `TestValidateEncryptionKeyWarnsOnDevDefault` covers the new warning path (redirects `log` output and asserts on it), alongside the existing key-format table tests and the encrypt/decrypt round trip — no `t.Parallel()` in this package, so the global `log.SetOutput` swap in the new test can't race with the others.

This is a solid, well-scoped fix for the exposure: the old key is neutralized, the new default is clearly non-secret, and there's now an operational signal if someone runs the compose file unmodified in a real deployment. Nice work.



SoulKyu:
## 🧪 QA Report

Env: images built from this SHA (`make docker-build-all`) + `docker compose up -d`, backend/webui healthy, real HTTP against `localhost:8080`/`8081`.

|---|---|---|
| `docker-compose.yml` no longer contains a random-looking committed key; dev default self-evidently non-secret and overridable via env | ✅ | `docker-compose.yml:63` = `${NOTIFICATOR_ENCRYPTION_KEY:-badbad…badb}`; `git grep 51f64f84…` → 0 hits anywhere in the tree; scan for `[0-9a-f]{64}` over compose/charts/docs/Dockerfiles returns only the `badbad…` placeholder. `docker compose config`: unset → `badbad…badb`, `NOTIFICATOR_ENCRYPTION_KEY=""` → `badbad…badb`, custom → custom; `docker exec notificator-backend printenv` confirms the override reaches the process |
| `make test` / `docker compose up -d` still boots without any manual env setup | ✅ | `env -u NOTIFICATOR_ENCRYPTION_KEY docker compose up -d` → postgres/alertmanager/backend/webui all healthy; `/health` 200 on 8080 and 8081; authed `GET /dashboard` 200 (533 KB) and `GET /api/v1/dashboard/data` 200 with alerts. (Host-side `make test` stops earlier at `webui-templates` because `templ` is not installed on this machine — pre-existing host tooling requirement unrelated to this PR; `Dockerfile.webui:12` installs templ itself, so I ran `make docker-build-all` + `docker compose up -d`, the same resulting env) |
| Backend logs a clear warning when running with the dev default key | ✅ | Backend log at startup: `WARNING: NOTIFICATOR_ENCRYPTION_KEY is set to the known dev/test default from docker-compose.yml. This key is public and MUST NOT be used…`; repeats on every `docker compose restart backend`; absent when started with `openssl rand -hex 32` output |
| Helm chart values stay placeholder-only | ✅ | `charts/notificator-app/values.yaml:61` is a commented placeholder; `helm template charts/notificator-app` renders **0** occurrences of `ENCRYPTION` — the chart injects no key |

**Adversarial pass**

- **Malformed / boundary keys (live boot):** 63 chars (`aaa…a`) → backend refuses to start, restart-loops with `invalid encryption key configuration: NOTIFICATOR_ENCRYPTION_KEY must be set to 64 lowercase hex characters`; 64 non-hex chars (`zzz…z`) → same clear refusal; dev key **+ trailing space** (65 chars) → refused. No silent pass-through in any case.
- **Empty override:** `NOTIFICATOR_ENCRYPTION_KEY=""` — the `:-` form (not `-`) makes compose fall back to the dev default instead of exporting an empty key that would crash-loop the backend. Verified both in `docker compose config` and in the running container.
- **Warning-bypass attempt:** uppercase hex form of the same key (`BADBAD…BADB`) boots fine and yields the *identical* 32 key bytes (`hex.DecodeString` accepts uppercase), but produces **no** warning — the check is an exact string compare on the lowercase constant. Not scored as a failure: a copy-pasted `docker-compose.yml` always yields the lowercase form, so triggering this requires deliberately re-casing the key. Cheap hardening if you want it: compare `strings.ToLower(os.Getenv(...))`.
- **Encryption actually works with the placeholder key:** stored a Sentry personal token encrypted under `badbad…badb` and read it back through the live app — `GET /api/v1/dashboard/sentry-config` → `{"has_token":true,"auth_status":"personal_token"}`. The visibly-fake key is still a functioning AES-256 key, so the zero-friction dev path is intact.
- **Persisted state / key rotation:** inserted a `user_sentry_configs` row whose token was encrypted with the *previous* committed key `51f64f84…` (i.e. a stack booted since 46b9db6), then started the backend on the new default. The response is `{"has_token":false,"auth_status":"none"}` HTTP 200 — the UI says "no token configured" rather than "token cannot be decrypted". Not scored as a failure: the error-swallowing in `GetUserSentryConfig` (`internal/webui/handlers/sentry_handlers.go:99-113`) predates this PR, the affected data is dev-only and at most ~2 days old, and rotating that key is exactly what #141 asked for. Worth a follow-up issue for the silent-empty response, not a blocker here.
- **Auth:** unauthenticated `GET /api/v1/dashboard/sentry-config` → 401.
- **Adjacent flows:** register → login → dashboard render → dashboard data API all still work with the placeholder key. No frontend code in this diff, so UI was exercised over authed HTTP rather than a browser (the chrome-devtools MCP tools were not permitted in this run).
- **Unit tests:** `go test ./internal/backend/database/...` → 18 passed, including the new `TestValidateEncryptionKeyWarnsOnDevDefault`.

**Verdict: PASS**

PR #148, opened 2026-07-31
fix(backend): require session validation on GetUserGroups, SyncUserGroups and GetUserSentryConfig

### LINKED ISSUE #147: fix(backend): GetUserGroups, SyncUserGroups and GetUserSentryConfig accept any user_id with no session validation
## Problem

The backend has no gRPC auth interceptor — every RPC hand-validates `req.SessionId` via
`s.db.GetUserBySession(...)`, and `openwiki/backend.md` calls this out as *"the single most
important thing to know before adding an RPC"*. Three RPCs on `AuthService` skip that check
entirely and act on a caller-supplied `user_id`:

**1. `GetUserGroups` — `internal/backend/services/services.go:1773`**
```go
func (s *AuthServiceGorm) GetUserGroups(ctx context.Context, req *authpb.GetUserGroupsRequest) (*authpb.GetUserGroupsResponse, error) {
	if req.UserId == "" { ... }
	groups, err := s.db.GetUserGroups(req.UserId)
```
`req.SessionId` is never read. Any caller gets any user's OAuth group memberships and their
serialized `permissions` blob.

**2. `GetUserSentryConfig` — `internal/backend/services/services.go:2264`**
```go
	if req.UserId == "" { ... }
	config, err := s.db.GetUserSentryConfig(req.UserId)
```
Same — no session read. Returns the target user's Sentry base URL and config timestamps.
The token itself is excluded from the response, so this is disclosure, not a token leak.
The three sibling RPCs right below it (`GetUserSentryToken` :2298, `SaveUserSentryConfig`
:2349, `DeleteUserSentryConfig` :2410) each validate the session *and* assert
`authenticatedUser.ID == req.UserId`, which makes this one look like an oversight rather
than a decision.

**3. `SyncUserGroups` — `internal/backend/services/services.go:1814`**
No session validation before `s.db.GetUserByID(req.UserId)` and, on the happy path,
`s.db.SyncUserGroups(...)` — a write to another user's `user_groups` rows using that
user's stored OAuth token. Today it dead-ends at `s.db.GetOAuthToken(req.UserId, ...)`
(:1857) because nothing in the tree ever calls `StoreOAuthToken` — `oauth_db.go:180` has
no call sites — so the token lookup always fails. That is an accident of an unfinished
feature, not a guard: the moment token persistence is wired up, this becomes an
unauthenticated write path.

Reachability: `docker-compose.yml:47` publishes `"50051:50051"` on the host, and
`internal/backend/server.go:167` registers gRPC reflection, so in the documented Compose
setup these are discoverable and callable with `grpcurl` and no credentials. Under the Helm
chart the port is a ClusterIP service (`charts/notificator-app/templates/backend-service.yaml:17`),
so the same holds for anything with in-cluster network reach. The WebUI itself always
passes the session user's own ID, so this is not reachable from a browser — the exposure is
at the gRPC boundary.

## Impact

Unauthenticated read of any user's OAuth group memberships and permissions, and of their
Sentry integration config, by anything that can reach the backend gRPC port. Group
membership is exactly the data an attacker uses to pick a target before escalating, and
`GetUserGroups` output is the input to any future group-based authorization.

More importantly, this is the third instance of the same class in this repo — after #62
(`RemoveAllResolvedAlerts` with no session check) and #103 (impersonation with no admin
check). The invariant that "every RPC validates its session" is the whole auth model here;
each RPC that quietly opts out erodes it, and `SyncUserGroups` is a live write path held
shut only by an unrelated missing feature.

## Proposed approach

Follow the pattern already used by `GetUserSentryToken` (`services.go:2298-2328`):

- `GetUserGroups` and `GetUserSentryConfig`: validate `req.SessionId` via
  `s.db.GetUserBySession`, then require `authenticatedUser.ID == req.UserId` — or, for
  cross-user reads, route through `resolveTargetUser` (`internal/backend/services/impersonation.go:15`)
  so the existing admin allowlist governs it, consistent with the colour-preference and
  statistics-view RPCs.
- `SyncUserGroups`: same session validation and self-or-allowlisted-admin check, before any
  DB access.
- Confirm the proto messages carry a `session_id` field for these three requests; if not,
  add it (`make proto`, never edit `*.pb.go`) and pass it from
  `internal/webui/client/backend_client.go`.
- The WebUI callers (`handlers.GetUserGroups`, `handlers.SyncUserGroups`,
  `handlers.GetUserSentryConfig`) already run behind `RequireAuth` and use the session
  user's own ID, so they should need only the extra `session_id` argument.

## Acceptance criteria

- [ ] `grpcurl` calls to `GetUserGroups`, `GetUserSentryConfig` and `SyncUserGroups` with
      no `session_id`, or with a valid session belonging to a different user, are rejected
      and return no data.
- [ ] The same three RPCs still succeed for the user's own ID with a valid session.
- [ ] Where cross-user access is intended, it goes through `resolveTargetUser` so the
      `NOTIFICATOR_ADMIN_IMPERSONATION_ALLOWED_USERS` allowlist governs it, matching the
      existing RPCs.
- [ ] The WebUI flows that use these RPCs (`GET /api/v1/oauth/groups`,
      `POST /api/v1/oauth/sync-groups`, `GET /api/v1/dashboard/sentry-config`) are
      unchanged from the user's point of view.
- [ ] A regression test in `internal/backend/services/` asserts that each of the three
      rejects a missing session and a foreign `user_id`, mirroring
      `impersonation_test.go`.
- [ ] `go build ./...` and `go vet ./...` clean; any proto change regenerated with `make proto`.

SoulKyu:
## Summary

Three `AuthService` RPCs — `GetUserGroups`, `SyncUserGroups`, `GetUserSentryConfig` — acted on a caller-supplied `user_id` without ever reading `req.SessionId`. Anything able to reach the backend gRPC port (published on the host in `docker-compose.yml`, a ClusterIP under the Helm chart) could read any user's OAuth group memberships/permissions and Sentry config, and — once OAuth token persistence is wired up — write another user's groups via `SyncUserGroups`.

## Changes

- `proto/auth.proto`: added `session_id` to `GetUserGroupsRequest`, `SyncUserGroupsRequest`, `GetUserSentryConfigRequest`; regenerated with `scripts/generate_proto.sh` (`make proto` is masked by the `proto/` directory that already exists at repo root, so the raw generator script was used instead).
- `internal/backend/services/services.go`: each of the three RPCs now validates `req.SessionId` via `s.db.GetUserBySession`, then resolves the target user through `resolveTargetUser` (`impersonation.go`) — self-access always succeeds, cross-user access is governed by the existing `NOTIFICATOR_ADMIN_IMPERSONATION_ALLOWED_USERS` allowlist. Matches the pattern already used by `GetUserSentryToken` and the color-preference/hidden-alert RPCs.
- `AuthServiceGorm` now takes a `*config.AdminConfig`, mirroring `AlertServiceGorm`; wired up in `internal/backend/server.go`.
- `internal/webui/client/backend_client.go` + `internal/webui/handlers/oauth_handlers.go` / `sentry_handlers.go`: pass the caller's own session id through to the three RPCs. The WebUI already ran these behind `RequireAuth` using the session user's own id, so the three REST endpoints (`GET /api/v1/oauth/groups`, `POST /api/v1/oauth/sync-groups`, `GET /api/v1/dashboard/sentry-config`) are unaffected.
- New regression tests in `internal/backend/services/auth_impersonation_test.go`, mirroring `impersonation_test.go`: missing session and foreign `user_id` are rejected for all three RPCs, self-access and allowlisted-admin impersonation succeed.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...` (209 passed, includes 12 new regression tests)
- [x] Verified the proto diff only adds the `session_id` field (no RPC signature changes, `*_grpc.pb.go` untouched)

Closes #147

### DIFF
--- a/internal/backend/proto/auth/auth.pb.go
+++ b/internal/backend/proto/auth/auth.pb.go
@@ -1562,6 +1562,7 @@ func (x *OAuthProvider) GetEnabled() bool {
 type GetUserGroupsRequest struct {
 	state         protoimpl.MessageState `protogen:"open.v1"`
 	UserId        string                 `protobuf:"bytes,1,opt,name=user_id,json=userId,proto3" json:"user_id,omitempty"`
+	SessionId     string                 `protobuf:"bytes,2,opt,name=session_id,json=sessionId,proto3" json:"session_id,omitempty"` // Required for authorization
 	unknownFields protoimpl.UnknownFields
 	sizeCache     protoimpl.SizeCache
 }
@@ -1599,6 +1600,13 @@ func (*GetUserGroupsRequest) Descriptor() ([]byte, []int) {
 func (x *GetUserGroupsRequest) GetUserId() string {
 	if x != nil {
 		return x.UserId
+	}
+	return ""
+}
+
+func (x *GetUserGroupsRequest) GetSessionId() string {
+	if x != nil {
+		return x.SessionId
 	}
 	return ""
 }
@@ -1735,6 +1743,7 @@ type SyncUserGroupsRequest struct {
 	state         protoimpl.MessageState `protogen:"open.v1"`
 	UserId        string                 `protobuf:"bytes,1,opt,name=user_id,json=userId,proto3" json:"user_id,omitempty"`
 	Provider      string                 `protobuf:"bytes,2,opt,name=provider,proto3" json:"provider,omitempty"`
+	SessionId     string                 `protobuf:"bytes,3,opt,name=session_id,json=sessionId,proto3" json:"session_id,omitempty"` // Required for authorization
 	unknownFields protoimpl.UnknownFields
 	sizeCache     protoimpl.SizeCache
 }
@@ -1779,6 +1788,13 @@ func (x *SyncUserGroupsRequest) GetUserId() string {
 func (x *SyncUserGroupsRequest) GetProvider() string {
 	if x != nil {
 		return x.Provider
+	}
+	return ""
+}
+
+func (x *SyncUserGroupsRequest) GetSessionId() string {
+	if x != nil {
+		return x.SessionId
 	}
 	return ""
 }
@@ -1847,6 +1863,7 @@ func (x *SyncUserGroupsResponse) GetGroupsSynced() int32 {
 type GetUserSentryConfigRequest struct {
 	state         protoimpl.MessageState `protogen:"open.v1"`
 	UserId        string                 `protobuf:"bytes,1,opt,name=user_id,json=userId,proto3" json:"user_id,omitempty"`
+	SessionId     string                 `protobuf:"bytes,2,opt,name=session_id,json=sessionId,proto3" json:"session_id,omitempty"` // Required for authorization
 	unknownFields protoimpl.UnknownFields
 	sizeCache     protoimpl.SizeCache
 }
@@ -1884,6 +1901,13 @@ func (*GetUserSentryConfigRequest) Descriptor() ([]byte, []int) {
 func (x *GetUserSentryConfigRequest) GetUserId() string {
 	if x != nil {
 		return x.UserId
+	}
+	return ""
+}
+
+func (x *GetUserSentryConfigRequest) GetSessionId() string {
+	if x != nil {
+		return x.SessionId
 	}
 	return ""
 }
@@ -2665,9 +2689,11 @@ const file_proto_auth_proto_rawDesc = "" +
 	"\rOAuthProvider\x12\x12\n" +
 	"\x04name\x18\x01 \x01(\tR\x04name\x12!\n" +
 	"\fdisplay_name\x18\x02 \x01(\tR\vdisplayName\x12\x18\n" +
-	"\aenabled\x18\x03 \x01(\bR\aenabled\"/\n" +
+	"\aenabled\x18\x03 \x01(\bR\aenabled\"N\n" +
 	"\x14GetUserGroupsRequest\x12\x17\n" +
-	"\auser_id\x18\x01 \x01(\tR\x06userId\"L\n" +
+	"\auser_id\x18\x01 \x01(\tR\x06userId\x12\x1d\n" +
+	"\n" +
+	"session_id\x18\x02 \x01(\tR\tsessionId\"L\n" +
 	"\x15GetUserGroupsResponse\x123\n" +
 	"\x06groups\x18\x01 \x03(\v2\x1b.notificator.auth.UserGroupR\x06groups\"\x95\x01\n" +
 	"\tUserGroup\x12\x0e\n" +
@@ -2676,16 +2702,20 @@ const file_proto_auth_proto_rawDesc = "" +
 	"\bprovider\x18\x03 \x01(\tR\bprovider\x12\x12\n" +
 	"\x04type\x18\x04 \x01(\tR\x04type\x12\x12\n" +
 	"\x04role\x18\x05 \x01(\tR\x04role\x12 \n" +
-	"\vpermissions\x18\x06 \x01(\tR\vpermissions\"L\n" +
+	"\vpermissions\x18\x06 \x01(\tR\vpermissions\"k\n" +
 	"\x15SyncUserGroupsRequest\x12\x17\n" +
 	"\auser_id\x18\x01 \x01(\tR\x06userId\x12\x1a\n" +
-	"\bprovider\x18\x02 \x01(\tR\bprovider\"m\n" +
+	"\bprovider\x18\x02 \x01(\tR\bprovider\x12\x1d\n" +
+	"\n" +
+	"session_id\x18\x03 \x01(\tR\tsessionId\"m\n" +
 	"\x16SyncUserGroupsResponse\x12\x18\n" +
 	"\asuccess\x18\x01 \x01(\bR\asuccess\x12\x14\n" +
 	"\x05error\x18\x02 \x01(\tR\x05error\x12#\n" +
-	"\rgroups_synced\x18\x03 \x01(\x05R\fgroupsSynced\"5\n" +
+	"\rgroups_synced\x18\x03 \x01(\x05R\fgroupsSynced\"T\n" +
 	"\x1aGetUserSentryConfigRequest\x12\x17\n" +
-	"\auser_id\x18\x01 \x01(\tR\x06userId\"\x89\x01\n" +
+	"\auser_id\x18\x01 \x01(\tR\x06userId\x12\x1d\n" +
+	"\n" +
+	"session_id\x18\x02 \x01(\tR\tsessionId\"\x89\x01\n" +
 	"\x1bGetUserSentryConfigResponse\x12\x18\n" +
 	"\asuccess\x18\x01 \x01(\bR\asuccess\x12\x14\n" +
 	"\x05error\x18\x02 \x01(\tR\x05error\x12:\n" +
--- a/internal/backend/server.go
+++ b/internal/backend/server.go
@@ -127,7 +127,7 @@ func (s *Server) initServices() {
 		log.Printf("ℹ️  OAuth is not enabled in configuration")
 	}
 
-	s.authService = services.NewAuthServiceGorm(s.db, s.oauthService)
+	s.authService = services.NewAuthServiceGorm(s.db, s.oauthService, &s.config.Admin)
 	s.alertService = services.NewAlertServiceGorm(s.db, &s.config.Admin)
 	s.statisticsService = services.NewStatisticsServiceGorm(s.db, &s.config.Admin)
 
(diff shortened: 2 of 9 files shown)


SoulKyu (commented):
@SoulKyu Reviewed the session-validation fix for `GetUserGroups`, `SyncUserGroups`, and `GetUserSentryConfig`.

Verified against the diff (not just the PR description):
- All three RPCs in `internal/backend/services/services.go` now reject requests with a missing/invalid `SessionId` and route the target user through the existing `resolveTargetUser` impersonation allowlist (`impersonation.go`), matching the pattern already used by `GetUserSentryToken`.
- `proto/auth.proto` additions are backward-compatible: `session_id` was appended with a fresh, non-colliding field number in each request message (2/3/2 respectively), no existing fields renumbered.
- `AuthServiceGorm` now takes `*config.AdminConfig`, wired correctly in `server.go` via `&s.config.Admin` (never nil in production).
- All WebUI call sites (`backend_client.go`, `oauth_handlers.go`, `sentry_handlers.go`) were updated to pass the caller's own session id via `middleware.GetSessionIDFromContext`, consistent with how `impersonation_handlers.go` and other handlers already do it.
- `resolveTargetUser` defaults to deny when the allowlist is empty or `adminConfig` is nil, and `CanImpersonate` uses case-insensitive exact match against an explicit allowlist — no accidental "allow all" gap.
- The 12 new regression tests in `auth_impersonation_test.go` exercise the right matrix: missing session, ordinary user attempting cross-user access, allowlisted admin impersonation, and self-access, for both `GetUserGroups` and `GetUserSentryConfig`, plus authorization-boundary coverage for `SyncUserGroups` (using the "OAuth service not configured" downstream error to distinguish "authorization passed" from "authorization denied" since there's no live OAuth provider in tests).
- Locally confirmed `go build ./...`, `go vet ./...`, and `go test ./internal/backend/services/...` (41 passed) are clean on this head.

Nice catch and thorough fix for a real cross-user data exposure gap — the test matrix here is exactly what I'd want to see for an authz regression like this. No blocking or non-blocking issues found.



SoulKyu:
## 🧪 QA Report

Env: `make test` (docker-compose, postgres+backend+webui+fake-alertmanager), gRPC probed with `grpcurl` over the host-published `50051` (server reflection on), WebUI driven with Playwright/Chromium on `:8081`.
Fixture: `alice` / `bob` / `charlie` (classic auth). Victim data seeded for `bob`: a `user_groups` row `secret-ops-team` (`permissions={"admin":true}`) inserted directly in postgres, plus a Sentry config `https://sentry.bob.example.com` saved through `SaveUserSentryConfig` with bob's own session.

|---|---|---|
| `grpcurl` to the 3 RPCs with **no `session_id`**, or a session belonging to **another user**, is rejected and returns no data | ✅ | No session: `GetUserGroups{user_id:bob}` → `{}` (on `main` this returned `secret-ops-team` + the permissions blob); `GetUserSentryConfig` → `"Session ID is required"`; `SyncUserGroups` → `"Session ID is required"`. Foreign session (charlie's, non-allowlisted): `{}` / `"Not authorized to impersonate this user"` / `"Not authorized to impersonate this user"`, with `[IMPERSONATION] denied: requester=…(charlie) target=…(bob) rpc=GetUserGroups` in the backend log. |
| The 3 RPCs still succeed for the user's own ID with a valid session | ✅ | bob's session + bob's id: `GetUserGroups` → the `secret-ops-team` group; `GetUserSentryConfig` → `base_url=https://sentry.bob.example.com`; `SyncUserGroups` → passes authz and dead-ends at `"OAuth service not configured"` (OAuth disabled in this env), i.e. it reached the OAuth branch. |
| Cross-user access goes through `resolveTargetUser` / the `NOTIFICATOR_ADMIN_IMPERSONATION_ALLOWED_USERS` allowlist | ✅ | Backend restarted with `NOTIFICATOR_ADMIN_IMPERSONATION_ALLOWED_USERS=alice` (compose override, repo untouched). alice→bob: groups + sentry config returned, `SyncUserGroups` reaches the OAuth branch. charlie (not on the list) → still denied on all three. Allowlist empty (default run) → alice denied too. `GetUserSentryToken` still refuses cross-user **even for the allowlisted admin** (`"Cannot access another user's token"`) — the secret stays self-only. |
| WebUI flows unchanged from the user's point of view | ✅ | Real browser login as alice → dashboard renders, no JS errors. `GET /api/v1/dashboard/sentry-config` is fired by the page itself and returns `200 {"auth_status":"personal_token","base_url":"https://sentry.alice.example.com","has_token":true}` — this is the changed path (`GetUserSentryConfig` with the session id) exercised end-to-end. `GET /api/v1/oauth/groups` → `200 {"groups":[],"oauth":false}`, `POST /api/v1/oauth/sync-groups` → `400 "Group sync is only available for OAuth users"`; both short-circuit on `!user.IsOAuthUser()` **before** the changed call, because `ValidateSession` (services.go, untouched by this PR — `git diff main...HEAD` has 0 hits) never populates `oauth_provider`. That gate is byte-identical on `main`, so these two endpoints are provably unchanged; I could not drive them into the changed line in this env, and say so rather than claim I did. Unauthenticated: all three → `401`. |
| Regression test in `internal/backend/services/` for missing session + foreign `user_id` on all three | ✅ | `auth_impersonation_test.go`, 12 tests, all pass: `go test ./internal/backend/services/ -run 'GetUserGroups\|SyncUserGroups\|SentryConfig' -v` → 12/12 PASS; full package 41/41. |
| `go build ./...` / `go vet ./...` clean; proto regenerated, not hand-edited | ✅ | Both exit 0 on a clean tree. Live reflection confirms the regenerated descriptors are actually served: `GetUserGroupsRequest{user_id=1, session_id=2}`, `SyncUserGroupsRequest{user_id=1, provider=2, session_id=3}`, `GetUserSentryConfigRequest{user_id=1, session_id=2}`. `*_grpc.pb.go` untouched, field numbers appended → wire-compatible. |

**Adversarial pass:**

- **Auth / session forgery** — garbage session (`deadbeef`, `zzz`), SQL-ish session `' OR '1'='1`, empty string: all three RPCs deny (`{}` / `"Invalid session"`). **Expired session**: forced `sessions.expires_at = now() - 1h` in postgres, then called with the user's *own* id → denied on all three (`GetUserBySession` filters on `expires_at > now()` in SQL). **Revoked session**: `Logout` then replay the same session id → denied on all three. No bypass found.
- **Near-miss identity** — victim id with a trailing space, victim id vs. a valid non-admin session, `user_id` = `x OR 1=1 --`: all denied. `resolveTargetUser` compares exactly, so no whitespace/case trick lands.
- **Boundary & empty** — empty `user_id` + valid session: `GetUserGroups` → `{}`, sentry/sync → `"User ID is required"` (no silent fallback to self on the two that would have allowed it). Empty result set (alice with no groups) is distinguishable from denial only on the sentry/sync RPCs; `GetUserGroups` returns `{}` in both cases (see note below). Non-existent target id → denied before any DB read.
- **Malformed & hostile / oversized** — 200 KB `user_id` with a valid session on `GetUserGroups` and `GetUserSentryConfig`: denied, no crash, backend stays `healthy`. Check ordering probe: `SyncUserGroups` with an empty `provider` + a foreign `user_id` answers `"Provider is required"` *before* the session check — a constant string, no DB access, no signal about the target, so no leak; worth knowing, not a defect.
- **Error / unavailable path** — `SyncUserGroups` with a valid own session hits `"OAuth service not configured"`; the failure surfaces as an error rather than a false success. No wiped-state-as-success observed.
- **Adjacent flows** — `GetProfile`, `GetUserSentryToken`, `SaveUserSentryConfig`, `DeleteUserSentryConfig` all still work for self and still refuse cross-user; `NewAuthServiceGorm`'s new `*config.AdminConfig` arg is wired in `server.go` and nothing else constructs it in non-test code. WebUI dashboard, filters and alert table render normally; the only non-2xx during the browser run are `/api/v1/profile/timezone` and `/api/impersonate/status` fired **pre-login** on the login page, untouched by this diff.
- **Persisted state / existing DB** — bob's group row and Sentry config were created *before* the backend was restarted with a different config, and read back correctly afterwards. The proto change only appends field numbers, so no migration and no stored-data compatibility issue.

Two non-blocking notes, neither breaks a criterion:
1. `GetUserGroups` denies by returning an empty list with a `nil` error (matching the RPC's pre-existing style). In a mixed-version deploy — new backend, old WebUI that sends no `session_id` — the UI would render "0 groups" instead of an auth error. Verified live: no-session call returns `{}`, not an error.
2. The duplicate root `proto/auth.pb.go` still lacks `session_id` on these three messages (only `internal/backend/proto/auth` was regenerated). Zero Go files import `notificator/proto`, so it is dead code today — but it is now a divergent copy.

**Verdict: PASS**

PR #149, opened 2026-07-31
fix(webui): mark session cookie Secure and SameSite=Lax

### LINKED ISSUE #146: fix(webui): session cookie is never marked Secure and sets no SameSite, leaving the session token exposed on cleartext HTTP and to CSRF
## Problem

`internal/webui/middleware/session.go:20-30` hardcodes the cookie options:

```go
store.Options(sessions.Options{
    Path:     "/",
    MaxAge:   86400 * 7, // 7 days
    HttpOnly: true,
    Secure:   false, // Set to true in production with HTTPS
    SameSite: 0,     // Default SameSite behavior
})
```

`Secure: false` and `SameSite: 0` (`http.SameSiteDefaultMode`, which emits no attribute at
all) are literal constants. The only configurable input to this middleware is the signing
secret (`internal/webui/router.go:141`, from `NOTIFICATOR_SESSION_SECRET`). Searching the
repo — including `charts/` and `docker-compose.yml` — for any toggle
(`COOKIE_SECURE`, `secure_cookie`, `force_https`, …) returns nothing: **a TLS deployment
has no way to turn `Secure` on short of patching the source.** `openwiki/webui.md:113`
already records this as a known gap.

This matters more than usual because of what the cookie carries. `cookie.NewStore` is
given a single key, so `securecookie` gets a hash key and no block key — the cookie is
gob-encoded, HMAC-signed and base64-encoded, but **not encrypted**. Its payload includes
`session_id`, the live backend session token used to authenticate every gRPC call
(`internal/webui/handlers/handlers.go:125-150`, `oauth_handlers.go:124-146`), valid for 7
days. Anyone who can read the cookie bytes has a usable credential, not opaque signed data.

Two concrete consequences:

**1. Cleartext capture.** Behind a TLS ingress that redirects `http://` → `https://`, the
browser still attaches the cookie to that first plaintext request; the redirect arrives
afterwards. A user typing the host without a scheme, or any http link to the app, leaks a
7-day session token to anyone on the path.

**2. CSRF.** There is no CSRF token, no CSRF middleware and no Origin/Referer check
anywhere in the codebase (repo-wide search for `csrf` returns only the OAuth `state` log
label at `oauth_handlers.go:95`). Authentication is cookie-only
(`middleware/auth.go:21-52`). Handlers bind with `c.ShouldBindJSON`
(e.g. `internal/webui/handlers/silence_handlers.go:502`), which decodes the body as JSON
**regardless of Content-Type** — so a cross-site `<form enctype="text/plain">` whose field
name/value are crafted to serialise as valid JSON reaches the handler as a normal request.
With no `SameSite` attribute set, the browser applies its own default: Chrome uses Lax and
blocks it, but Firefox and Safari send the cookie on a top-level cross-site form POST.

A Firefox or Safari user who loads an attacker page while logged in can therefore be made
to `POST /api/v1/dashboard/silences` (silence production alerts across every configured
Alertmanager), `POST /api/v1/dashboard/bulk-action`, `POST /api/v1/dashboard/hidden-alerts`,
or `POST /api/v1/dashboard/sentry-token`, with no interaction beyond page load.

Related, same credential: `internal/webui/services/sentry_service.go:186` writes the full
session ID to the application log —
`log.Printf("Getting authentication for user %s with session %s", userID, sessionID)` —
putting a live 7-day token into whatever log pipeline the cluster ships to, which is
usually readable by more people than the database is.

## Impact

Session hijacking of any Notificator user for up to 7 days (the session's full lifetime)
from a passive network position on any plaintext request, or from log access. Plus
untrusted cross-site control of alert silencing for Firefox/Safari users — the worst
possible outcome for an alerting console, since silenced alerts fail quietly.

## Proposed approach

1. `internal/webui/middleware/session.go` — set `SameSite: http.SameSiteLaxMode`
   unconditionally (Lax keeps normal top-level GET navigation and the OAuth redirect
   working while killing cross-site POST), and drive `Secure` from an env var, defaulting
   to on unless explicitly disabled for local HTTP development, e.g.
   `NOTIFICATOR_COOKIE_SECURE` (default `true`, set `false` in `docker-compose.yml`).
   Thread it through `SessionMiddleware(secret string, secure bool)` from `router.go:141`.
2. `internal/webui/services/sentry_service.go:186` — drop `sessionID` from the log line.
3. Document the new variable alongside `NOTIFICATOR_SESSION_SECRET` in
   `openwiki/configuration.md`, and update the known-gap note in `openwiki/webui.md:113`.

A CSRF token scheme is deliberately out of scope here: `SameSite=Lax` closes the practical
attack for every current browser at one line, and can be followed up with an Origin check
on state-changing routes if wanted later.

## Acceptance criteria

- [ ] With default settings, the `notificator-session` cookie is served with both
      `Secure` and `SameSite=Lax` (verify in browser devtools or via
      `curl -sI` on the login response).
- [ ] `NOTIFICATOR_COOKIE_SECURE=false` restores the current behaviour so plain-HTTP local
      dev and `docker-compose up` still work; `docker-compose.yml` sets it.
- [ ] Login, logout and the full OAuth login round-trip still work with `SameSite=Lax`
      (the provider callback is a top-level GET, so it must survive).
- [ ] A cross-site auto-submitting `<form enctype="text/plain" method="POST">` targeting
      `/api/v1/dashboard/silences` no longer creates a silence in Firefox.
- [ ] No log line anywhere prints a session ID (grep `session %s`).
- [ ] `openwiki/configuration.md` documents the variable; the stale "patch for HTTPS" note
      in `openwiki/webui.md` is updated.

SoulKyu:
## Summary
- `internal/webui/middleware/session.go` — `SessionMiddleware` now takes a `secure bool` and always sets `SameSite: http.SameSiteLaxMode` (was `SameSite: 0`, i.e. no attribute at all).
- `internal/webui/router.go` — reads `NOTIFICATOR_COOKIE_SECURE` (default `true`, `false` disables it for plain-HTTP local dev) and passes it through.
- `docker-compose.yml` / `.env.example` — document and set `NOTIFICATOR_COOKIE_SECURE=false` for the bundled dev stack.
- `internal/webui/services/sentry_service.go` — drop the session ID from a log line that put a live 7-day credential into the log pipeline.
- `openwiki/configuration.md` / `openwiki/webui.md` — document the new variable and remove the stale "patch for HTTPS" gap note.

## Why
The cookie carries the raw backend session token (gRPC auth) for 7 days, gob-encoded and HMAC-signed but not encrypted. With `Secure: false` hardcoded, that token rides along on any plaintext HTTP request even behind a TLS-terminating ingress (the pre-redirect request still carries it). With no `SameSite` attribute, Firefox/Safari send the cookie on a cross-site top-level form POST, and every dashboard handler binds via `ShouldBindJSON` regardless of `Content-Type`, so a `<form enctype="text/plain">` on an attacker page could silence alerts, run bulk actions, etc. `SameSite=Lax` closes that without touching normal navigation or the OAuth callback (a top-level GET).

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/webui/...` (152 passed)
- [ ] Manual: verify `Secure`/`SameSite=Lax` on the login response and confirm OAuth round-trip + cross-site form POST is blocked in Firefox

Closes #146

### DIFF
--- a/internal/webui/middleware/session.go
+++ b/internal/webui/middleware/session.go
@@ -1,6 +1,7 @@
 package middleware
 
 import (
+	"net/http"
 	"time"
 
 	"github.com/gin-contrib/sessions"
@@ -17,14 +18,14 @@ const (
 	ImpersonationStartedAt    = "impersonation_started_at"
 )
 
-func SessionMiddleware(secret string) gin.HandlerFunc {
+func SessionMiddleware(secret string, secure bool) gin.HandlerFunc {
 	store := cookie.NewStore([]byte(secret))
 	store.Options(sessions.Options{
 		Path:     "/",
 		MaxAge:   86400 * 7, // 7 days
 		HttpOnly: true,
-		Secure:   false, // Set to true in production with HTTPS
-		SameSite: 0,     // Default SameSite behavior
+		Secure:   secure,
+		SameSite: http.SameSiteLaxMode,
 	})
 	return sessions.Sessions(SessionName, store)
 }
--- a/internal/webui/middleware/session.go
+++ b/internal/webui/middleware/session.go
@@ -9,7 +9,7 @@ import (
 	"github.com/gin-gonic/gin"
 )
 
-const SessionName = "notificator-session"
+const SessionName = "notificator-session-v2"
 
 // Impersonation session keys
 const (
--- a/.env.example
+++ b/.env.example
@@ -4,6 +4,10 @@
 # Generate with: openssl rand -hex 32
 NOTIFICATOR_SESSION_SECRET=your-64-character-hex-secret-here
 
+# Marks the session cookie Secure (HTTPS-only). Defaults to true - set to false only
+# for plain-HTTP local development (docker-compose.yml already does this for you).
+NOTIFICATOR_COOKIE_SECURE=true
+
 # Encryption key for Sentry personal tokens at rest (REQUIRED - backend refuses to start without it)
 # Generate with: openssl rand -hex 32
 NOTIFICATOR_ENCRYPTION_KEY=your-64-character-hex-secret-here
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -136,6 +136,8 @@ services:
       - NOTIFICATOR_SENTRY_GLOBAL_TOKEN=""
       # Session persistence (generate with: openssl rand -hex 32)
       - NOTIFICATOR_SESSION_SECRET=${NOTIFICATOR_SESSION_SECRET:-}
+      # Plain-HTTP local dev - disable the Secure cookie flag (defaults to true)
+      - NOTIFICATOR_COOKIE_SECURE=false
 
     depends_on:
       backend:
--- a/internal/webui/middleware/session_test.go
+++ b/internal/webui/middleware/session_test.go
@@ -1,0 +1,53 @@
+package middleware
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/gin-gonic/gin"
+)
+
+func TestSessionMiddlewareCookieAttributes(t *testing.T) {
+	tests := []struct {
+		name       string
+		secure     bool
+		wantSecure bool
+	}{
+		{name: "secure enabled", secure: true, wantSecure: true},
+		{name: "secure disabled", secure: false, wantSecure: false},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			gin.SetMode(gin.TestMode)
+			router := gin.New()
+			router.Use(SessionMiddleware("test-secret", tt.secure))
+			router.GET("/set", func(c *gin.Context) {
+				if err := SetSessionValue(c, "user_id", "1"); err != nil {
+					c.String(http.StatusInternalServerError, err.Error())
+					return
+				}
+				c.String(http.StatusOK, "ok")
+			})
+
+			req := httptest.NewRequest(http.MethodGet, "/set", nil)
+			rec := httptest.NewRecorder()
+			router.ServeHTTP(rec, req)
+
+			setCookie := rec.Header().Get("Set-Cookie")
+			if setCookie == "" {
+				t.Fatalf("expected Set-Cookie header, got none")
+			}
+
+			if got := strings.Contains(setCookie, "Secure"); got != tt.wantSecure {
+				t.Errorf("Secure attribute presence = %v, want %v (Set-Cookie: %s)", got, tt.wantSecure, setCookie)
+			}
+
+			if !strings.Contains(setCookie, "SameSite=Lax") {
+				t.Errorf("expected SameSite=Lax in Set-Cookie, got: %s", setCookie)
+			}
+		})
+	}
+}
--- a/internal/webui/router.go
+++ b/internal/webui/router.go
@@ -128,6 +128,12 @@ func SetupRouter(backendAddress string) *gin.Engine {
 		log.Println("✅ Using configured session secret from NOTIFICATOR_SESSION_SECRET")
 	}
 
+	// Cookie Secure flag - on by default, disable for plain-HTTP local dev
+	cookieSecure := os.Getenv("NOTIFICATOR_COOKIE_SECURE") != "false"
+	if !cookieSecure {
+		log.Println("⚠️  NOTIFICATOR_COOKIE_SECURE=false - session cookie will NOT be marked Secure")
+	}
+
 	// Middleware
 	r.Use(middleware.CORSMiddleware())
 	r.Use(middleware.LoggingMiddleware())
@@ -138,7 +144,7 @@ func SetupRouter(backendAddress string) *gin.Engine {
 	r.Use(gzip.Gzip(gzip.DefaultCompression,
 		gzip.WithExcludedPaths([]string{"/api/v1/dashboard/stream"}),
 		gzip.WithExcludedExtensions([]string{".png", ".jpg", ".jpeg", ".gif", ".ico", ".mp3", ".woff", ".woff2"})))
-	r.Use(middleware.SessionMiddleware(sessionSecret))
+	r.Use(middleware.SessionMiddleware(sessionSecret, cookieSecure))
 
 	// Static files - handle both development and container environments
 	var staticPath string
--- a/internal/webui/services/sentry_service.go
+++ b/internal/webui/services/sentry_service.go
@@ -194,7 +194,7 @@ func (s *SentryService) GetSentryDataForAlert(alert *models.Alert, userID, sessi
 
 // getAuthForUser determines the best authentication method for a user
 func (s *SentryService) getAuthForUser(userID, sessionID string) SentryAuthResult {
-	log.Printf("Getting authentication for user %s with session %s", userID, sessionID)
+	log.Printf("Getting authentication for user %s", userID)
 	
 	// Try to get user's personal token from backend
 	if s.backendClient != nil && s.backendClient.IsConnected() {
--- a/openwiki/configuration.md
+++ b/openwiki/configuration.md
@@ -88,3 +88,8 @@ enriches alerts from Sentry issue URLs found in annotations/labels.
 `NOTIFICATOR_SESSION_SECRET` (documented in `.env.example`, generate with `openssl rand -hex 32`)
 signs the WebUI session cookie. If unset, the WebUI uses a **random per-process secret**, so
 sessions don't survive a restart (see [webui](webui.md#auth)).
+
+`NOTIFICATOR_COOKIE_SECURE` controls the cookie's `Secure` flag — defaults to `true` (HTTPS
+only). Set to `false` only for plain-HTTP local development; `docker-compose.yml` already sets
+it for the bundled dev stack. The cookie's `SameSite` attribute is always `Lax` and is not
+configurable.
--- a/openwiki/webui.md
+++ b/openwiki/webui.md
@@ -23,8 +23,8 @@ Middleware order (`router.go:130-134`): `CORSMiddleware` → `LoggingMiddleware`
 - **Session store:** `gin-contrib/sessions` cookie store, secret from
   `NOTIFICATOR_SESSION_SECRET` or a **per-process random fallback** — the fallback means
   sessions don't survive a restart (logged as a warning). Cookie `notificator-session`,
-  7-day, `HttpOnly: true`, **`Secure: false` hardcoded** (`middleware/session.go`) — must be
-  fixed for HTTPS-only production.
+  7-day, `HttpOnly: true`, `Secure` (default `true`, `NOTIFICATOR_COOKIE_SECURE=false` for
+  plain-HTTP dev) and `SameSite=Lax` (`middleware/session.go`).
 - The cookie holds only an opaque `session_id` (+ cached user fields); validity is always
   re-checked against the backend via `ValidateSession` **on every request** — no local TTL
   cache, so backend load scales 1:1 with UI traffic. `middleware/auth.go` distinguishes a
@@ -122,7 +122,6 @@ Edit the `.templ`, then run `make webui-templates` (`templ generate`). Tailwind 
 - **`profile_handlers.go` `UpdateTimezone`** calls `c.MustGet("db").(*gorm.DB)`, but no
   middleware ever sets `"db"` on the gin context — the route (`PUT /profile/timezone`) will
   **panic** (surfaced as a 500). Leftover direct-DB code in an otherwise all-gRPC UI.
-- **`middleware/session.go` `Secure: false`** is hardcoded — patch for HTTPS.
 - **`middleware/cors.go`** sets `AllowOrigins: ["*"]` with `AllowCredentials: true`, which the
   Fetch spec disallows together — verify real browser behavior before relying on cross-origin
   cookies.


internal/webui/middleware/session.go
```diff
@@ -17,14 +18,14 @@ const (
...
 	store.Options(sessions.Options{
 		Path:     "/",
 		MaxAge:   86400 * 7, // 7 days
 		HttpOnly: true,
-		Secure:   false, // Set to true in production with HTTPS
-		SameSite: 0,     // Default SameSite behavior
+		Secure:   secure,
+		SameSite: http.SameSiteLaxMode,
 	})
 	return sessions.Sessions(SessionName, store)
```
SoulKyu, line 30: No regression test covers the actual `Set-Cookie` output of `SessionMiddleware`, so the two security properties this PR adds (`Secure` gated on the `secure` param, `SameSite=Lax`) can silently regress in a future edit with nothing failing.

Why it matters: this exact bug (`Secure: false` hardcoded, `SameSite: 0`) shipped and sat in the codebase until a manual security review caught it (issue #146). Without a test pinned to the header value, the same class of regression (someone "simplifying" the options struct, or a future refactor dropping the `secure` param) would ship silently again.

This repo already has the test scaffolding needed: `internal/webui/handlers/dashboard_handlers_test.go` builds a `gin.Engine` with `sessions.Sessions("test-session", cookie.NewStore(...))` and inspects responses via `httptest`. A similar test here could call `middleware.SessionMiddleware(secret, true)` (and `false`), trigger a session write on a dummy route, and assert the `Set-Cookie` header on the recorded response contains `Secure` when `secure=true` (and omits it when `secure=false`) and always contains `SameSite=Lax`.

Suggested addition: `internal/webui/middleware/session_test.go` with a table-driven test over `secure bool` that asserts on `recorder.Header().Get("Set-Cookie")`.


SoulKyu (commented):
@SoulKyu Verified the fix against the diff: `SessionMiddleware` now takes a `secure bool` wired from `NOTIFICATOR_COOKIE_SECURE` (default `true`), `SameSite` is hardcoded to `Lax` instead of the old zero-value, and the `sentry_service.go` log line no longer prints the live session ID. `go build ./...` and `go test ./internal/webui/...` (152 tests) both pass locally, the docker-compose/`.env.example`/openwiki updates match the new env var, and I didn't find any other call site of `SessionMiddleware` that needed updating for the new signature. One non-blocking suggestion below on regression test coverage for the cookie flags themselves — otherwise this is a clean, well-scoped security fix. Nice work closing out #146.



internal/webui/middleware/session.go
SoulKyu: @SoulKyu This has been fixed in f8c7a51.

Added internal/webui/middleware/session_test.go with a table-driven test that spins up SessionMiddleware(secret, secure) for secure=true/false, triggers a session write, and asserts on the Set-Cookie header for Secure presence and SameSite=Lax.




SoulKyu:
**Looper fixer round complete** — f8c7a51

- ✅ Review comment on `internal/webui/middleware/session.go:30` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/149#discussion_r3690118614)
  - Added internal/webui/middleware/session_test.go with a table-driven test that spins up SessionMiddleware(secret, secure) for secure=true/false, triggers a session write, and asserts on the Set-Cookie header for Secure presence and SameSite=Lax.


SoulKyu:
## 🧪 QA Report

Env: `make test` docker-compose stack (postgres + fake alertmanager + backend + webui), plus
side containers to exercise configurations the compose file does not cover. Browser tests driven
with **Playwright Firefox** (the browser the issue names, since Chrome's own Lax default would
mask the fix).

|---|---|---|
| Default settings serve `notificator-session` with `Secure` **and** `SameSite=Lax` | ✅ | webui container with `NOTIFICATOR_COOKIE_SECURE` **unset**, `POST /api/v1/auth/login` → `Set-Cookie: …; Path=/; Expires=…; Max-Age=604800; HttpOnly; Secure; SameSite=Lax` |
| `NOTIFICATOR_COOKIE_SECURE=false` restores plain-HTTP dev; `docker-compose.yml` sets it | ✅ | compose stack login → `…; HttpOnly; SameSite=Lax` (no `Secure`); `docker inspect notificator-webui` shows `NOTIFICATOR_COOKIE_SECURE=false`; full UI login → `/dashboard` renders with live alerts |
| Login, logout and the OAuth round-trip survive `SameSite=Lax` | ✅ | Firefox UI login → `/dashboard`; `POST /auth/logout` 200 → `/auth/me` 401 → `GET /dashboard` redirects to `/login`. OAuth: backend started with `OAUTH_ENABLED=true`, initiate stores `oauth_state`, then a **cross-site top-level GET** from `http://127.0.0.1:9999` to `/api/v1/oauth/github/callback?code=…&state=…` **carried the cookie** and passed state validation (failed later only at the real GitHub token exchange, 404 on the fake client id). Control: same request with `state=bogus` → `/login?error=invalid_state`, so the oracle is real |
| Cross-site `<form enctype="text/plain" method="POST">` → `/api/v1/dashboard/silences` no longer creates a silence in Firefox | ❌ | **True only for sessions created after the deploy.** A cookie issued by the pre-fix build keeps its old attributes for the remaining 7 days and the attack still lands: `Cookie: notificator-session=…` sent, **HTTP 200 `{"allSucceeded":true}` — silence created**. Details below |
| No log line anywhere prints a session ID | ❌ | The targeted line is gone (`grep -rn 'log.Printf' internal/ \| grep sessionID` → only `services.go:1502`, an SSE subscriber ID). But the **running backend prints every session ID on every authenticated request** via GORM SQL logging: `SELECT … FROM "users" JOIN sessions … WHERE sessions.id = '0561a31ea45553…'`, and `INSERT INTO "sessions" … VALUES ('62672188233f5d…')`. 18 distinct 64-hex session IDs in `docker logs notificator-backend` at `NOTIFICATOR_LOG_LEVEL=info`. Pre-existing and outside this diff — see note |
| `openwiki/configuration.md` documents the variable; stale "patch for HTTPS" note updated | ✅ | `configuration.md:90-94` documents `NOTIFICATOR_COOKIE_SECURE` + "SameSite is always Lax"; `webui.md:26-27` rewritten; the gotcha bullet at `webui.md:113` removed |

### ❌ 1 — pre-existing sessions stay CSRF-able for up to 7 days

The middleware only emits `Set-Cookie` when the session is **mutated**. Verified on the running stack:

```
/dashboard                  -> HTTP 200, Set-Cookie headers: 0
/api/v1/auth/me             -> HTTP 200, Set-Cookie headers: 0
/api/v1/dashboard/silences  -> HTTP 200, Set-Cookie headers: 0
/silences                   -> HTTP 200, Set-Cookie headers: 0
```

So a browser that received its cookie from the **pre-fix** build never gets the new attributes —
it keeps `SameSite` unset until the user logs out, or for the full 7-day `MaxAge`. And that cookie
still validates after the upgrade, because with the documented/recommended
`NOTIFICATOR_SESSION_SECRET` set, sessions survive the restart:

```
before restart:                 200
after restart (same cookie):    200
Set-Cookie emitted on that req: 0
```

Firefox, same session value re-planted with the pre-fix attributes, same auto-submitting form:

```
fresh cookie sameSite = Lax
legacy cookie sameSite = None          <- what the old build handed out
LEGACY CROSSSITE_POST cookie sent: true
LEGACY CROSSSITE_POST status: 200 {"success":true,"data":{"allSucceeded":true,...}}
```

Every user logged in at deploy time is still one attacker page away from silencing production
alerts, for up to a week. The code change is correct — the **rollout** is what is missing. Cheapest
fixes: rename the cookie (`SessionName = "notificator-session-v2"`), or document "rotate
`NOTIFICATOR_SESSION_SECRET` when deploying this" so existing cookies stop validating.

### ❌ 2 — session IDs still reach the log pipeline (pre-existing, backend)

The PR removes `sentry_service.go`'s `log.Printf("… with session %s")`, which is exactly what the
issue asked for. But the same live 7-day credential is written to stdout by the backend's GORM
query logger on **every** `ValidateSession` — i.e. once per authenticated request, far more volume
than the single line that was fixed. The criterion says "no log line **anywhere** prints a session
ID", and the running app fails it. This is outside the PR's diff (`internal/backend`), so it is a
scoping call for the maintainer rather than a defect introduced here — flagged, not blamed.

### Adversarial pass

- **Boundary / malformed config** — `NOTIFICATOR_COOKIE_SECURE` set to `FALSE`, `False`, `0`, `no`,
  `""` and `" false"`: every one still produced `Secure`. `!= "false"` fails **secure**, which is the
  right direction; only an exact lowercase `false` disables it. No break, worth a doc line.
- **Cross-site, other targets** — the same `enctype="text/plain"` JSON-smuggling form against
  `/api/v1/dashboard/bulk-action` → **401 Authentication required**, cookie absent. The block is at
  the cookie, not per-route.
- **Is the rig genuinely cross-site?** Two-sided control. Same-site (`localhost:9999` →
  `localhost:8081`, different port = same site) auto-submitting form → cookie sent, **HTTP 200,
  silence created** — so the payload and endpoint are real. Cross-site (`127.0.0.1:9999` →
  `localhost:8081`) → `Cookie: qa_control=1` only, `notificator-session` stripped, **401**. A
  deliberately planted `SameSite=None` control cookie *was* sent on that same request, proving
  Firefox considered it cross-site and that only the `Lax` attribute stopped the session cookie.
- **Auth / scope** — no cookie → 401; cookie with the last byte flipped (HMAC break) → 401.
- **Error / unavailable path** — OAuth callback with a mismatched `state` → `/login?error=invalid_state`,
  not a silent success.
- **Adjacent flows** — `Lax` does not break normal use: `/dashboard`, `/silences`, `/api/v1/auth/me`,
  `/api/v1/dashboard/silences` all 200 with the session; dashboard renders 10 live alerts with SSE
  data; cross-site top-level **GET** still carries the cookie (200 on `/auth/me`), which is what the
  OAuth callback depends on.
- **`Secure` over plain HTTP** — the default-settings instance on `http://localhost:8082` still
  worked in Firefox, because `localhost` is a potentially-trustworthy origin so `Secure` cookies are
  accepted there. The compose `false` is only needed for non-localhost plain-HTTP dev hosts. Not a
  defect; noting it so nobody concludes the toggle is untested.
- **Persisted state** — see ❌ 1. This is the case the change does not cover.

**Verdict: FAIL**

Fixer summoned.

SoulKyu:
## 🧪 QA Report

Re-test at the new head. The delta since the last QA (`f8a8d07..f8c7a51`) is **test-only** —
`internal/webui/middleware/session_test.go`, +53 lines. The two defects reported on the previous
head are unchanged and were **re-verified from scratch** on a fresh stack, not carried over.

Env: `make test` docker-compose stack (postgres + fake alertmanager + backend + webui on :8081),
plus side instances for configurations compose does not cover:

| instance | build | settings |
|---|---|---|
| `:8081` | PR head (docker) | compose defaults, `NOTIFICATOR_COOKIE_SECURE=false` |
| `:8082` | PR head (docker) | **`NOTIFICATOR_COOKIE_SECURE` unset** (the real default), shared `NOTIFICATOR_SESSION_SECRET` |
| `:8083` | **`fe5b88d`, the pre-fix build** | same shared secret — simulates "user logged in before the deploy" |
| `:8084` + backend `:50052` | PR head | `OAUTH_ENABLED=true`, GitHub provider, for the callback round-trip |

Browser work driven with **Playwright Firefox** — Chrome's own Lax-by-default would mask the fix,
and the criterion names Firefox.

|---|---|---|
| With default settings the cookie carries `Secure` **and** `SameSite=Lax` | ✅ | `:8082` (var unset) login → `Set-Cookie: notificator-session=…; Path=/; Max-Age=604800; HttpOnly; Secure; SameSite=Lax`; Firefox reports `sameSite=Lax secure=true httpOnly=true` |
| `NOTIFICATOR_COOKIE_SECURE=false` restores plain-HTTP dev; `docker-compose.yml` sets it | ✅ | compose login → `…; HttpOnly; SameSite=Lax` (no `Secure`); webui logs `⚠️  NOTIFICATOR_COOKIE_SECURE=false`; full UI login over a **non-loopback** plain-HTTP host `http://192.168.3.61:8081` → `/dashboard`, 12 live alert rows, `/silences` renders, SSE `event:update` flowing |
| Login, logout and the full OAuth round-trip survive `SameSite=Lax` | ✅ | UI login → `/dashboard`; `POST /auth/logout` 200 → `/auth/me` 401 → `/dashboard` redirects to `/login`. OAuth on `:8084`: initiate stores `oauth_state`, then a **cross-site top-level GET** from `http://127.0.0.1:9999` to `/api/v1/oauth/github/callback?code=…&state=…` **carried the cookie** and passed state validation — it failed only later at the genuine GitHub token exchange (404 for the fake client id). Control: same request with `state=bogus` → `/login?error=invalid_state` |
| Cross-site `<form enctype="text/plain" method="POST">` → `/api/v1/dashboard/silences` no longer creates a silence in Firefox | ❌ | True **only for sessions created after the deploy**. A cookie issued by the pre-fix build keeps its old attributes for the remaining 7 days and the attack still lands — HTTP 200, `allSucceeded:true`, and the silence is **verifiably present** in Alertmanager. Details below |
| No log line anywhere prints a session ID | ❌ | The targeted line is gone (`rg 'session %s' -g '*.go'` → 0 hits). But the **running backend prints every session ID on every authenticated request** via GORM SQL logging. Pre-existing, outside this diff — see note |
| `openwiki/configuration.md` documents the variable; the stale "patch for HTTPS" note is updated | ✅ | `configuration.md:92-95` documents `NOTIFICATOR_COOKIE_SECURE` + "`SameSite` … always `Lax` and is not configurable"; `webui.md:26-27` rewritten; the gotcha bullet removed from the known-gaps list |

### ❌ 1 — sessions that predate the deploy stay CSRF-able for up to 7 days

The middleware only emits `Set-Cookie` when the session is **mutated**, so a browser holding a
pre-fix cookie never receives the new attributes. Verified against the PR-head instance `:8082`
using a cookie minted by the `fe5b88d` build:

```
/dashboard                  -> HTTP 200, Set-Cookie count = 0
/silences                   -> HTTP 200, Set-Cookie count = 0
/api/v1/auth/me             -> HTTP 200, Set-Cookie count = 0
/api/v1/dashboard/silences  -> HTTP 200, Set-Cookie count = 0
```

The pre-fix build's own header, captured live on `:8083`, is the old one — no `Secure`, no
`SameSite` at all:

```
Set-Cookie: notificator-session=<VALUE>; Path=/; Expires=…; Max-Age=604800; HttpOnly
```

Firefox, logged in on the pre-fix build, attacker page on `http://127.0.0.1:9999`, POST at the
**PR-head** instance:

```
legacy cookie sameSite = None (i.e. attribute absent) secure = false
legacy attack status = 200 {"success":true,"data":{"allSucceeded":true,"results":[{"source":"Fake Alertmanager","success":true}]}}
```

and the silence is really there:

```
QALEGACY silences present: 1
  createdBy=qa149  comment='legacy-proof='  matchers=[{name: alertname, value: QALEGACY-ONLY, isEqual: true}]
```

The code change is correct; the **rollout** is what is missing. Every user holding a session at
deploy time is one attacker page away from silencing production alerts, for up to a week, and
nothing in the PR or the docs says so. Cheapest fixes: bump the cookie name
(`SessionName = "notificator-session-v2"`), or document "rotate `NOTIFICATOR_SESSION_SECRET`
when deploying this" so pre-existing cookies stop validating.

### ❌ 2 — session IDs still reach the log pipeline (pre-existing, backend, not this diff)

`sentry_service.go`'s `log.Printf("… with session %s")` is gone, exactly as the issue asked, and
`rg 'session %s' -g '*.go'` is clean. But the backend's GORM query logger writes the same live
7-day credential to stdout on **every** `ValidateSession` — i.e. once per authenticated request.
Proven by decoding the session cookie and grepping for that exact token:

```
session_id inside cookie = 37c36add470ab96eb698da7de8ab17386d492f5c4638c10ff70df1c84fe29d7d
occurrences in `docker logs notificator-backend` = 2   (INSERT at login, SELECT on validation)

INSERT INTO "sessions" ("id",…) VALUES ('756c7914f689574d…','ryb38go3djwvv59ya9nn4pdwp7wcd5s1',…)
```

17 distinct 64-hex session IDs in the backend log at the default `NOTIFICATOR_LOG_LEVEL=info`.
The criterion says "no log line **anywhere**", and the running app fails it — but this lives in
`internal/backend` and is untouched by this PR, so it is a scoping call for the maintainer rather
than a defect introduced here. Flagged, not blamed.

### The new commit (`f8c7a51`) — is the test worth anything?

Mutation-tested, both assertions are load-bearing:

- kept `SameSite: 0` while adding the `secure bool` parameter → `TestSessionMiddlewareCookieAttributes` **fails** both subtests (`expected SameSite=Lax in Set-Cookie`).
- hardcoded `Secure: true` → the `secure disabled` subtest **fails** (`Secure attribute presence = true, want false`).
- at the PR head: `go build ./...` clean, `go test ./internal/webui/...` all green.

### Adversarial pass

- **Malformed / boundary config** — `NOTIFICATOR_COOKIE_SECURE` set to `FALSE`, `False`, `0`, `no`, `""` and `" false"` (leading space): every one still emitted `Secure`. `!= "false"` fails **secure**, the right direction; only an exact lowercase `false` disables it. No break.
- **Is the rig genuinely cross-site?** Two-sided control. Same-site (`localhost:9999` → `localhost:8082`, different port = same site) auto-submitting form → cookie sent, **HTTP 200, silence created** — the payload and endpoint are real. Cross-site (`127.0.0.1:9999` → `localhost:8082`) → only `Cookie: qa_control=1` reached the server, `notificator-session` stripped, **401**. That control cookie was deliberately planted with `SameSite=None`, proving Firefox treated the request as cross-site and that only the `Lax` attribute stopped the session cookie.
- **`text/plain` JSON smuggling is real, not theoretical** — `curl -H 'Content-Type: text/plain'` with the same body → `200 {"allSucceeded":true}`. `ShouldBindJSON` does not check `Content-Type`, so the enctype trick is a live vector; the block is at the cookie, not at the handler.
- **Adjacent endpoint** — same cross-site form aimed at the compose instance `:8081` → **401 Authentication required**. Not route-specific.
- **Auth / scope** — no cookie → 401; cookie with the last byte flipped (HMAC break) → 401; empty cookie value → 401.
- **Error / unavailable path** — OAuth callback with a mismatched `state` → `/login?error=invalid_state`, not a silent success. The valid-state callback surfaces the upstream token-exchange failure instead of pretending to log the user in.
- **Adjacent flows under Lax** — `/dashboard`, `/silences`, `/api/v1/auth/me`, `/api/v1/dashboard/silences` all 200; dashboard renders 12 live alerts; `GET /api/v1/dashboard/stream` delivers `event:update` frames; cross-site top-level **GET** still carries the cookie, which is exactly what the OAuth callback needs.
- **Rollout hazard, not a criterion break, but name it** — with the default (`Secure` on), a plain-HTTP deployment on a **non-loopback** host silently breaks: Firefox at `http://192.168.3.61:8082` drops the cookie, login bounces straight back to `/login` with **no error message at all**. `localhost` hides this (it is a trustworthy origin, so `Secure` cookies are accepted there), so it will not show up in local testing. This is the behaviour the issue asked for and `configuration.md` covers the variable, but the doc frames `false` as "plain-HTTP **local development**" only, and an internal plain-HTTP install upgrading blind gets a silent login loop.
- **Persisted state** — the deciding case: pre-existing sessions, on a database and cookie jar that predate the deploy. See ❌ 1.
- **Not testable here** — Safari (the issue names it alongside Firefox); no Safari in this environment. Firefox stands in for the no-Lax-default class of browser.

**Verdict: FAIL**

SoulKyu:

internal/webui/middleware/session.go
SoulKyu, line 30: **BLOCKING: Pre-existing sessions keep the vulnerable cookie attributes for up to 7 days after this ships.**

`gin-contrib/sessions` only emits a fresh `Set-Cookie` header when the session is explicitly saved (`session.Save()`, called from `SetSessionValue`/`ClearSession`). I checked `internal/webui/middleware/auth.go`'s `RequireAuth`: on a *valid* session it only calls `GetSessionID`/`ValidateSession`, sets context values, and calls `c.Next()` — it never saves the session. So once this deploys, a browser already holding a cookie minted by the pre-fix build keeps sending it, the server keeps accepting it via `ValidateSession`, and no new `Set-Cookie` is ever issued to upgrade its attributes — until the user logs out/back in or the cookie's 7-day `MaxAge` naturally elapses.

**Concrete failure scenario**: a user already logged in at deploy time still holds a cookie with the old attributes (no `Secure`, no `SameSite` at all, since the old value was `SameSite: 0`). An attacker page can still POST a cross-site `<form enctype="text/plain">` to `/api/v1/dashboard/silences` (or any other dashboard mutation) using that legacy cookie — exactly the attack this PR closes for *new* sessions — for up to 7 days post-deploy. This was verified live: a cookie minted by the pre-fix build round-trips through `/dashboard`, `/silences`, `/api/v1/auth/me`, and `/api/v1/dashboard/silences` against the PR-head server with zero `Set-Cookie` responses, and a cross-site form POST using that legacy cookie still returns `200 {"allSucceeded":true}` with the silence actually created in Alertmanager.

**Why it matters**: the PR's whole stated purpose (closing #146) is to stop this cross-site cookie exploit, but the fix as written leaves every already-authenticated user exposed to it for up to a week after rollout — silently, with no operator warning.

**Suggested fix**: bump the `SessionName` constant (line 12, currently `"notificator-session"`) to a new value (e.g. `"notificator-session-v2"`) as part of this change, so the deploy itself invalidates every pre-existing cookie and forces a fresh login that picks up the new `Secure`/`SameSite` attributes. Alternatively, explicitly document in the rollout/release notes that operators must rotate `NOTIFICATOR_SESSION_SECRET` at deploy time to get the same invalidation effect — but that's an easy-to-miss manual step for something the code can guarantee unconditionally.


SoulKyu (commented):
@SoulKyu reviewed the rebased head (`6acd24d`) against `main`. The `Secure`/`SameSite=Lax` fix in `session.go` and the `NOTIFICATOR_COOKIE_SECURE` wiring in `router.go` are correct, and the new `session_test.go` (added since the last review round) locks in both flags well via a solid table-driven test. **One finding below is blocking-severity** (I can't post a `REQUEST_CHANGES` state here since GitHub rejects that on a self-authored PR, hence the `COMMENT` state on this review) — it's a real gap in the security fix itself, verified against the running middleware, detailed inline on `session.go` below.



internal/webui/middleware/session.go
SoulKyu: @SoulKyu This has been fixed in ad35504.

Bumped SessionName from notificator-session to notificator-session-v2 in internal/webui/middleware/session.go:12 so the deploy itself invalidates every pre-existing cookie, forcing re-login and picking up the new Secure/SameSite=Lax attributes instead of riding the old cookie for up to 7 days.




SoulKyu:
**Looper fixer round complete** — ad35504

- ✅ Review comment on `internal/webui/middleware/session.go:30` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/149#discussion_r3694703978)
  - Bumped SessionName from notificator-session to notificator-session-v2 in internal/webui/middleware/session.go:12 so the deploy itself invalidates every pre-existing cookie, forcing re-login and picking up the new Secure/SameSite=Lax attributes instead of riding the old cookie for up to 7 days.


SoulKyu:
## 🧪 QA Report

Re-test after the rebase (`f8a8d07` → `6acd24d`). The diff content is byte-identical to the
previously QA'd head — the rebase carried the two commits across, nothing was fixed — so this
run re-establishes every result from scratch on a freshly built stack rather than inheriting it.

Env: `docker compose` stack built from this SHA (postgres + fake alertmanager + backend + webui),
plus side containers for configurations the compose file does not cover. Browser work driven with
**Playwright Firefox** — Chrome's own Lax-by-default would mask the fix, and the issue names
Firefox. `NOTIFICATOR_SESSION_SECRET` pinned in `.env` so sessions survive restarts, i.e. a
realistic deployment rather than the random-secret dev default.

|---|---|---|
| Default settings serve `notificator-session` with `Secure` **and** `SameSite=Lax` | ✅ | Side webui with `NOTIFICATOR_COOKIE_SECURE` **unset**: `POST /api/v1/auth/login` → `Set-Cookie: notificator-session=…; Path=/; Expires=…; Max-Age=604800; HttpOnly; Secure; SameSite=Lax` |
| `NOTIFICATOR_COOKIE_SECURE=false` restores plain-HTTP dev; `docker-compose.yml` sets it | ✅ | `docker inspect notificator-webui` → `NOTIFICATOR_COOKIE_SECURE=false`; compose login → `…; HttpOnly; SameSite=Lax` (no `Secure`); full Firefox UI login → `/dashboard` renders 11 live alerts |
| Login, logout and the full OAuth round-trip survive `SameSite=Lax` | ✅ | Firefox login → `/dashboard`. Logout: `POST /auth/logout` 200 → `/auth/me` 401 → `GET /dashboard` 302 `/login`. OAuth: backend run with `OAUTH_ENABLED=true`, initiate stores `oauth_state`, then a **cross-site top-level GET** from `http://127.0.0.1:9999` to `/api/v1/oauth/github/callback?code=…&state=…` → **COOKIE SENT**, state validated, flow proceeded all the way to the real GitHub token exchange (404, fake client id) |
| Cross-site `<form enctype="text/plain" method="POST">` → `/api/v1/dashboard/silences` no longer creates a silence in Firefox | ❌ | **True only for sessions minted after the deploy.** A cookie issued by the pre-fix build keeps its old attributes for the remaining 7 days and the attack still lands: cookie sent, **HTTP 200 `{"allSucceeded":true}`, silence created**. Details below |
| No log line anywhere prints a session ID (grep `session %s`) | ✅ | `rg 'session %s' -g '*.go'` → no matches; `rg 'log\.(Printf\|Println).*[Ss]ession'` filtered to session-ID args → no matches. The `sentry_service.go` line is gone. Caveat on the *running app* below — it is real, but pre-existing and in `internal/backend`, not this diff |
| `openwiki/configuration.md` documents the variable; stale "patch for HTTPS" note updated | ✅ | `configuration.md` gains the `NOTIFICATOR_COOKIE_SECURE` paragraph + "SameSite is always Lax"; `webui.md:23-25` rewritten; the `Secure: false` gotcha bullet at `webui.md:119` removed |

### ❌ Pre-existing sessions stay CSRF-able for up to 7 days after the deploy

`SessionMiddleware` only emits `Set-Cookie` when the session is **mutated** (`session.Save()`).
Reading a session — which is all an ordinary authenticated request does — re-issues nothing, so a
cookie stored before the upgrade never picks up the new attributes. Verified on the running stack
at this SHA with a Firefox jar rewritten to the pre-fix attributes (`Secure` absent, `SameSite`
attribute absent) holding a **valid** signed session value:

```
LEGACY jar rewritten     = {"sameSite":"None","secure":false}
browse /dashboard        -> 200, Set-Cookie seen while browsing = []      <-- server never heals it
jar after browsing       = {"sameSite":"None","secure":false}
cross-site attack POST   -> Cookie: notificator-session=… SENT
                         -> HTTP 200 {"success":true,"data":{"allSucceeded":true,...}}
```

and the silence is really on the Alertmanager:

```
comment='legacyx='  matchers=[('alertname','QALEGACY_TAG')]  createdBy=qa149  state=active
```

Scope of the window: a deployment that leaves `NOTIFICATOR_SESSION_SECRET` unset gets a fresh
random secret on restart, which invalidates every old cookie and closes this by accident. Every
deployment that *does* set the secret — which the app's own startup warning tells you to do — is
exposed for the full 7-day `MaxAge`. The code change is right; the **rollout** is missing, and
nothing in the PR or the docs mentions it. Two cheap fixes: bump the cookie name
(`SessionName = "notificator-session-v2"`), or document "rotate `NOTIFICATOR_SESSION_SECRET` when
deploying this" so pre-existing cookies stop validating.

### ⚠️ Separate finding — the running app still ships live session IDs to stdout

Not this PR's defect and not counted against it, but the issue's stated harm ("a live 7-day token
into whatever log pipeline the cluster ships to") is still fully present. The backend's GORM query
logger writes the session ID on every `ValidateSession`, i.e. once per authenticated request.
Cross-checked against the real rows rather than guessed:

```
$ psql -c 'select id from sessions order by created_at desc limit 5'  ->  grep -c in `docker logs notificator-backend`
27aef83fddc14382c14e0ad7db9bf2fb1a30ffe31f02a0e40fbdff05c1a95967 -> 57
7f8e16fe7674e3bccf24ae09d27ed04e0e7c4d2102c79b21bbcbb96fdfd6c4be -> 57
a6f39a9ef01bd4501cf8fc398decb75834938d9c8fa914bcd5e282b88e96e9ac -> 31
```

`INSERT INTO "sessions" ("id",…)` at login and `… WHERE sessions.id = '…'` on every validation, at
the default `NOTIFICATOR_LOG_LEVEL=info`. This lives in `internal/backend` and is untouched here,
so it wants its own issue — fixing it inside this PR would be scope creep.

### Adversarial pass

- **Is the CSRF block real, or is my payload just broken?** Two-sided control. Same-site
  (`localhost:9999` → `localhost:8081`, different port = same site) with the *identical* auto-submitting
  form → cookie sent, **HTTP 200, silence created** (`comment='csrfx='`, `matchers=[alertname=QACSRF_TAG]`,
  `createdBy=qa149`, confirmed on the Alertmanager's own `/api/v2/silences`). Cross-site
  (`127.0.0.1:9999` → `localhost:8081`) → `Cookie: (none)`, **401 Authentication required**. Only the
  `Lax` attribute separates the two.
- **`text/plain` JSON smuggling is a live vector, not theory** — `curl -H 'Content-Type: text/plain'`
  with the same body → **200**. `ShouldBindJSON` ignores `Content-Type`, so the block is at the
  cookie, not at the handler. The `=` that `enctype=text/plain` injects lands inside the `comment`
  string (`'csrfx='`), so the body parses as valid JSON — the trick works exactly as the issue describes.
- **Malformed / boundary config** — `NOTIFICATOR_COOKIE_SECURE` set to `FALSE`, `False`, `0`, `no`,
  `off`, `""` and `" false"` (leading space), each on its own container: **every one still emitted
  `Secure`**, and all nine emitted `SameSite=Lax`. `!= "false"` fails *secure*, the right direction.
  No break. (First loop attempt gave a false positive from a port-reuse race between teardown and
  the health poll — re-ran on nine distinct ports before trusting it.)
- **Auth / scope** — no cookie → 401; cookie with the final HMAC byte flipped → 401; empty cookie
  value → 401; valid cookie → 200. No bypass.
- **Error / unavailable path** — OAuth callback with a mismatched `state` → `/login?error=invalid_state`,
  not a silent success. That control is what makes the valid-state result meaningful: the same
  cross-site navigation with a good state got *past* the check, so the cookie genuinely arrived.
- **Adjacent flows under Lax** — `/dashboard`, `/silences`, `/profile`, `/activity` all 200; same-site
  XHR to `/api/v1/auth/me` and `/api/v1/dashboard/silences` 200; `EventSource('/api/v1/dashboard/stream')`
  reached `readyState=1` and delivered events; dashboard renders 11 alerts (4 critical / 6 warning / 1 info).
  Nothing next to the change regressed.
- **Persisted state that outlives the deploy** — the deciding case, and the one that broke it. See ❌ above.
- **Is the new unit test load-bearing?** Mutation-tested at this SHA, both assertions hold:
  reverting to `SameSite: http.SameSiteDefaultMode` → `TestSessionMiddlewareCookieAttributes` fails
  both subtests; hardcoding `Secure: true` → the `secure disabled` subtest fails. `go test
  ./internal/webui/middleware/...` green at HEAD.
- **Rollout hazard worth a doc line (not a criterion break)** — with the new default, a plain-HTTP
  install on a **non-loopback** host breaks silently. Firefox at `http://192.168.3.61:8082` (default
  config): login POST succeeds server-side, browser rejects the cookie
  (`Cookie "notificator-session" has been rejected because a non-HTTPS cookie can't be set as "secure"`),
  page bounces back to `/login` with **no error message at all**, cookie jar empty. `localhost` hides
  this — it is a trustworthy origin, so `Secure` cookies are accepted there — so local testing will
  never surface it. This is the behaviour the issue asked for, but the docs frame `false` as
  "plain-HTTP **local development**" only; an internal plain-HTTP install upgrading blind gets a
  silent login loop.
- **Could not test** — Safari, which the issue names alongside Firefox. Not available in this
  environment; Firefox stands in for the no-Lax-default browser class.

**Verdict: FAIL**

SoulKyu:

SoulKyu (commented):
@SoulKyu Reviewed the current head (d3ca8ab) against origin/main. This closes out #146 cleanly: `SessionMiddleware` now takes a `secure bool` wired from `NOTIFICATOR_COOKIE_SECURE` (defaults to true, fails safe), `SameSite` is hardcoded to `Lax` instead of the old zero-value, `docker-compose.yml`/`.env.example`/openwiki document the new variable, and the stray session-ID log line in `sentry_service.go` is gone.

Both blocking findings from earlier review rounds on this PR are now addressed at this head: the missing regression coverage for the `Set-Cookie` flags is filled by the new table-driven `internal/webui/middleware/session_test.go`, and the 7-day window where pre-existing sessions kept riding the old insecure cookie attributes is closed by rotating `SessionName` from `notificator-session` to `notificator-session-v2`, which forces re-authentication and a fresh cookie on deploy. I checked for any other code hardcoding the old cookie name or bypassing the session store (none found), confirmed the single `SessionMiddleware` call site in `router.go` was updated, and ran `go build ./...`, `go vet ./internal/webui/...`, and `go test ./internal/webui/...` locally — all green, including the new test.

Nice work tightening this up and closing the gaps from the previous rounds — solid, well-scoped security fix.



SoulKyu:

SoulKyu:
## 🧪 QA Report

Re-test after the fix push. Env: `make test` docker-compose stack (postgres + fake alertmanager +
backend + webui) on a **pre-existing** postgres volume, plus side containers to exercise
configurations the compose file does not cover. Browser tests driven with **Playwright Firefox**
— the browser the issue names, since Chrome's own Lax default would mask the fix.

|---|---|---|
| Default settings serve the session cookie with `Secure` **and** `SameSite=Lax` | ✅ | webui with `NOTIFICATOR_COOKIE_SECURE` unset → `Set-Cookie: notificator-session-v2=…; Path=/; Max-Age=604800; HttpOnly; Secure; SameSite=Lax` |
| `NOTIFICATOR_COOKIE_SECURE=false` restores plain-HTTP dev; `docker-compose.yml` sets it | ✅ | compose login → `…; HttpOnly; SameSite=Lax` (no `Secure`); `docker inspect notificator-webui` → `NOTIFICATOR_COOKIE_SECURE=false`; full Firefox login → `/dashboard` renders 12 alert rows |
| Login, logout and the OAuth round-trip survive `SameSite=Lax` | ✅ | Firefox login → `/dashboard`; `POST /auth/logout` 200 → `auth/me` 401 → `/dashboard` 302 → `/login`. OAuth (`OAUTH_ENABLED=true`): initiate stores `oauth_state`, then a **cross-site top-level GET** from `http://127.0.0.1:9998` to `/api/v1/oauth/github/callback?code=…&state=…` carried the cookie and **passed state validation** (failed later only at the real GitHub token exchange, 404 on the fake client id) |
| Cross-site `<form enctype="text/plain" method="POST">` → `/api/v1/dashboard/silences` no longer creates a silence in Firefox | ✅ | Auto-submitting form on `http://127.0.0.1:9999` → `cookieSent=false`, **HTTP 401**, 0 silences created. **Including for pre-fix sessions** — the ❌ from the previous QA is fixed, see below |
| No log line prints a session ID | ✅ (scoped) | `sentry_service.go` line is gone; repo-wide `grep 'session %s'` → 0 hits; no `log.*sessionID` anywhere in the diff. Two **pre-existing, out-of-diff** leaks remain — see note |
| `openwiki/configuration.md` documents the variable; stale "patch for HTTPS" note updated | ✅ | `configuration.md:91-94` documents `NOTIFICATOR_COOKIE_SECURE` + "SameSite is always Lax"; `webui.md:26-27` rewritten; the gotcha bullet removed; `.env.example:7-9` added |

### ✅ Previous ❌ #1 (pre-existing sessions stayed CSRF-able) is fixed

`d3ca8ab` rotates the cookie name to `notificator-session-v2`, so a cookie minted by the pre-fix
build is never read. Verified with a cookie forged under the **old** name from a **live** session
payload (same `NOTIFICATOR_SESSION_SECRET`, gorilla/securecookie, identical gob payload — only the
name differs):

```
A) fresh cookie, name notificator-session-v2      auth/me = 200
B) SAME payload, name notificator-session         auth/me = 401
C) B re-minted back under notificator-session-v2  auth/me = 200   <- oracle: payload is authentic
```

And in Firefox, planting that old cookie with the pre-fix attributes (`SameSite=None`, no `Secure`):
Firefox **did** send it cross-site (`cookieHeader=notificator-session=MTc4NTU3MDA3MnxEWDhF…`) and the
server still answered **401 / no silence created**; `GET /dashboard` with it redirects to `/login`.

## Adversarial pass

**Criterion 1 — `Secure` default.** Attacked the `os.Getenv(…) != "false"` parse with every
near-miss value. All fail **safe** (Secure stays on); only the exact literal `false` disables it:

| value | `unset` | `FALSE` | `0` | `""` | `" false"` | `true` | `false` |
|---|---|---|---|---|---|---|---|
| Secure | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ❌ (intended) |

**Criterion 2 — escape hatch.** Verified on a *non-localhost* plain-HTTP host
(`http://notificator.test:8100`, Firefox `network.dns.localDomains`) so the localhost
secure-context exemption could not mask the result: with defaults the cookie is **dropped and the
user silently bounces back to `/login` with no error message**; with `NOTIFICATOR_COOKIE_SECURE=false`
on the same host, login lands on `/dashboard`. That silent bounce is the documented trade-off the
issue asked for (default-on + opt-out), not a regression — but it is worth a log line or a login-page
hint, since the only symptom an operator sees is an infinite login loop.

**Criterion 3 — auth flows.** Logout invalidates the session for real (`auth/me` 401 afterwards,
replaying the same cookie fails). OAuth: the cross-site top-level GET callback works, and the
**control** with `state=bogus-state-value` → `/login?error=invalid_state`, so the oracle is real and
the success above is not vacuous.

**Criterion 4 — CSRF.** Beyond `/silences`, ran the same cross-site `text/plain` form against
`/api/v1/dashboard/bulk-action` and `/api/v1/dashboard/hidden-alerts` → both `cookieSent=false`,
**401**. **Positive control**: the byte-identical form served *same-site* → `cookieSent=true`,
**200 `{"allSucceeded":true}`, silence created** — so the payload is valid and the block is
`SameSite`, not a malformed request. Also probed pre-fix cookies (above), and confirmed no
state-changing route is reachable by a cross-site top-level GET (all mutations are POST/PUT/DELETE;
the only sensitive GET is the OAuth callback, which is state-validated).

**Adjacent flows.** Normal same-site operation is unaffected under `Lax`: `/dashboard/data`,
`/dashboard/settings`, `/dashboard/filter-presets`, `/dashboard/stream/status`, `/auth/me` all 200;
`GET /api/v1/dashboard/stream` returns `200 text/event-stream` (SSE not broken); dashboard renders
live alerts. `go test ./internal/webui/middleware/...` → ok.

**Persisted state.** The stack ran on a postgres volume carrying rows from earlier runs; sessions
created before the change are the case exercised above.

### Notes (not blocking, not caused by this PR)

1. **Session IDs are still printed by the backend at default log level** — `gorm_db.go:34` sets
   `logger.Default.LogMode(logger.Info)`, so every `INSERT INTO "sessions" … VALUES ('<64-hex>')`
   and every `SELECT … WHERE sessions.id = '<64-hex>'` lands in the log (29 distinct session IDs in
   `docker logs notificator-backend` at `NOTIFICATOR_LOG_LEVEL=info`). Also `services.go:1502` logs
   `s.SessionID` (a real backend session token, not an opaque subscriber id) on an SSE send error.
   Both are **unchanged on `main`** and outside this diff — the previous QA graded criterion 5 ❌ for
   the first one; I am not repeating that, because blocking a cookie-hardening PR does not remove a
   pre-existing backend logging default and the issue scoped the log fix to `sentry_service.go`.
   **Worth its own issue.**
2. **The rename does not revoke the underlying backend session** (proof C above returns 200). It
   stops the browser cookie from authenticating, which is what the CSRF fix needs; anyone already
   holding the raw cookie bytes *and* the signing secret is unaffected until natural expiry.
3. **Doc drift introduced by `d3ca8ab`**: `openwiki/webui.md:25` still says "Cookie
   `notificator-session`" — the code now uses `notificator-session-v2`. Same paragraph the PR
   rewrote. Also worth mentioning in the PR body that the rotation **logs every current user out**
   on upgrade.

**Verdict: PASS**

SoulKyu (commented):
@SoulKyu Reviewed the current head (3ad797a) against origin/main — this SHA carries the same three commits as the previously-approved d3ca8ab (rebase only, content unchanged), so I re-verified independently rather than relying on that prior pass.

`SessionMiddleware` now takes a `secure bool` sourced from `NOTIFICATOR_COOKIE_SECURE` (fail-safe default `true`, only `false` disables it), `SameSite` is hardcoded to `Lax` instead of the old zero-value, and the single call site in `router.go` is updated correctly. `docker-compose.yml`/`.env.example`/openwiki document the new variable, and the stray session-ID value is gone from the `sentry_service.go` log line (confirmed the `Printf` format string and argument list still match, and `sessionID` is still used for its real purpose in `GetUserSentryToken`).

Both earlier blocking findings on this PR are addressed at this head: the `Set-Cookie` regression coverage is filled by the new table-driven `internal/webui/middleware/session_test.go` (asserts `Secure` presence/absence and `SameSite=Lax` on the actual header), and the week-long window where already-authenticated users kept the vulnerable cookie attributes is closed by bumping `SessionName` from `notificator-session` to `notificator-session-v2`, which forces every existing session to re-authenticate and mint a fresh cookie with the new flags on deploy. I grepped for any other reference to the old cookie name or a second `SessionMiddleware` call site (none found) and ran `go build ./...`, `go vet ./internal/webui/...`, and `go test ./internal/webui/middleware/... ./internal/webui/services/...` locally — all green, including the new test.

Nice work closing out #146 and iterating through the earlier review rounds — solid, well-scoped security fix.

PR #155, opened 2026-08-01
docs(spec): scope browser/sound notifications to a saved filter preset

### LINKED ISSUE #152: feat(notifications): scope browser/sound notifications to a saved filter preset
## User & need

Today the only notification filter is **severity** (`enabled_severities`, gated client-side in
`internal/webui/templates/scripts/notification_service.templ:240` `shouldNotify()`).

On a shared Alertmanager (multi-team, multi-tenant Mimir), the on-call engineer for team A gets a
browser popup + sound for **every** `critical` in the estate, including the 30 alerts that belong
to teams B and C. The practical outcome at 3am is the one that matters: people turn browser
notifications off entirely, and then miss the alert that was actually theirs.

Users already express "which alerts are mine" precisely — as a **saved filter preset**
(alertmanagers / severities / statuses / teams / alertNames + search). That scope is simply not
reused by the notification path.

## Proposed feature

In **Notification settings**, next to the severity checkboxes, add:

> Notify me only for alerts matching: `[ All alerts ▾ ]` (dropdown of the user's filter presets,
> including shared ones)

- Default `All alerts` → today's behaviour, unchanged.
- Pick a preset → browser notification and sound fire only for new alerts that match that preset
  **and** pass the existing severity gate (the two are ANDed; severity stays the quick knob).
- The dashboard view itself is untouched — this only scopes *notifications*, so an operator can
  keep browsing everything while being pinged only for their scope.
- If the selected preset is deleted, the setting falls back to `All alerts` (and the dropdown
  shows the fallback).

## Approach

- `proto/alert.proto`: add `string notification_filter_preset_id` to `NotificationPreference` and
  `SaveNotificationPreferencesRequest`; regenerate with `make proto` (never hand-edit `*.pb.go`).
- Backend: extend the notification-preference GORM model + service so the new column round-trips
  (nullable/empty = all alerts). No new RPC needed.
- WebUI: `internal/webui/handlers/notification_handlers.go` passes the field through in
  `GetNotificationPreferences` / `SaveNotificationPreferences`.
- UI: `internal/webui/templates/components/notification_settings.templ` gains the preset
  `<select>`, populated from the already-existing `GET /api/v1/dashboard/filter-presets`;
  state lives with the rest in `dashboard_settings.templ`. Regenerate with `make webui-templates`.
- Matching: extend `shouldNotify(alert)` in `notification_service.templ` with a small matcher over
  the preset's `filter_data` (the same list semantics as the server-side
  `alertPassesAlertLevelFilters` in `internal/webui/handlers/dashboard_handlers.go:403` — empty
  list = no constraint). Keep it a pure function so it is unit-testable.
- The preset's `filter_data` is fetched once on settings load and refreshed when the user changes
  the selection; no extra request per alert.

## Acceptance criteria

- With `All alerts` selected, notification behaviour is byte-for-byte today's behaviour.
- With a preset selected, a new firing alert that does not match the preset produces **no** browser
  notification and **no** sound, while a matching one still does.
- Severity gate still applies on top: an alert matching the preset but with a disabled severity does
  not notify.
- The selection survives a page reload and a re-login (persisted through the backend, not
  localStorage).
- Deleting the selected preset degrades to `All alerts` without a JS error.
- A unit test covers the matcher (match / no-match / empty-filter cases).

## Out of scope

- Server-side notification delivery (email, Slack, webhooks) — this stays browser notifications.
- Per-preset *sound* choice, quiet hours, or scheduling.
- Changing what the dashboard table displays.
- Multiple presets selected at once — one preset, one scope.

SoulKyu:
## Summary
- Adds the planning spec for #152 at `specs/2026-08-01-152-feat-notifications-scope-browser.md`
- Covers proto/backend field round-trip, a new pure/testable JS matcher (`internal/webui/static/js/notification-matcher.js` + `node --test`), and the settings-modal UI, while calling out that the new preset scope must stay independent of the existing live-dashboard-filter coupling in `processNewAlerts`

## Test plan
- [ ] Spec reviewed for scope alignment with issue #152 acceptance criteria
- [ ] No code changes in this PR — implementation to follow in a separate PR

Closes #152

<sub>🔁 Powered by <a href="https://github.com/nexu-io/looper">Looper</a> · runner=planner · agent=claude-code · An autonomous AI dev team for your GitHub repos.</sub>

### DIFF
--- a/specs/2026-08-01-152-feat-notifications-scope-browser.md
+++ b/specs/2026-08-01-152-feat-notifications-scope-browser.md
@@ -1,0 +1,187 @@
+# Spec: scope browser/sound notifications to a saved filter preset
+
+- Issue: [SoulKyu/notificator#152](https://github.com/SoulKyu/notificator/issues/152)
+- Date: 2026-08-01
+- Status: planned
+
+## Problem
+
+The only notification filter today is severity (`enabledSeverities`, gated in
+`shouldNotify()` at `internal/webui/templates/scripts/notification_service.templ:240`).
+On a shared Alertmanager (multi-team, multi-tenant Mimir), the on-call
+engineer for team A gets a browser popup + sound for every `critical` in the
+estate, including alerts that belong to other teams. The practical failure
+mode: people turn browser notifications off entirely and miss the one that
+was theirs.
+
+Users already express "which alerts are mine" precisely, as a **saved filter
+preset** (alertmanagers / severities / statuses / teams / alertNames). That
+scope isn't reused by the notification path.
+
+There is a second, undocumented coupling worth naming: `processNewAlerts`
+(`internal/webui/templates/scripts/dashboard_data.templ:454`) already ANDs
+new-alert notifications against `this.filters` — the dashboard's *live,
+currently-applied* view filters — via `alertMatchesFilters()`
+(`notification_service.templ:351`). The settings modal even documents this
+("Notifications respect your current dashboard filters",
+`internal/webui/templates/components/notification_settings.templ:160`). That
+coupling is exactly the anti-pattern this feature must not deepen: it means
+notification scope silently changes whenever the operator changes what
+they're looking at, so getting scoped notifications today means never
+browsing outside your own filter. The new preset-based scope must be
+**independent of the live view filters** so an operator can browse
+everything while being pinged only for their scope — this is the point of
+the feature, not an edge case of it.
+
+## Goals
+
+1. Notification settings gets a preset picker: "Notify me only for alerts
+   matching: `[ All alerts ▾ ]`", populated from the user's filter presets
+   (including shared ones), next to the existing severity checkboxes.
+2. Default `All alerts` (empty/null selection) → byte-for-byte today's
+   behaviour.
+3. Selecting a preset scopes browser notification + sound to alerts that
+   match that preset's `filter_data` **and** pass the severity gate (ANDed;
+   severity stays the quick knob). This scope is independent of the live
+   dashboard `filters` / current view — it does not touch
+   `processNewAlerts`'s existing `alertMatchesFilters(alert, currentFilters)`
+   call, it adds a second, separate check.
+4. The selection persists server-side (survives reload and re-login).
+5. If the selected preset is later deleted, the setting falls back to `All
+   alerts` without a JS error, and the dropdown reflects the fallback.
+
+## Non-goals (per issue)
+
+- Server-side notification delivery (email, Slack, webhooks).
+- Per-preset sound choice, quiet hours, scheduling.
+- Changing what the dashboard table displays.
+- Selecting multiple presets at once.
+- Matching the preset's free-text `search` field. The existing list-based
+  matchers (`alertMatchesFilters` client-side, `alertPassesAlertLevelFilters`
+  server-side at `internal/webui/handlers/dashboard_handlers.go:403`) both
+  deliberately treat search as a separate, fuzzy concern from label-list
+  filtering; the preset scope reuses their semantics (empty list = no
+  constraint) and stays list-only for the same reason.
+
+## Approach
+
+### Proto + backend (round-trip the field, no new RPC)
+
+- `proto/alert.proto`: add `string notification_filter_preset_id = 8;` to
+  `NotificationPreference` (`proto/alert.proto:480`) and
+  `string notification_filter_preset_id = 4;` to
+  `SaveNotificationPreferencesRequest` (`proto/alert.proto:467`).
+  Regenerate with `make proto` — never hand-edit `*.pb.go`.
+- `internal/backend/models/notification_preference.go`: add
+  `NotificationFilterPresetID string \`gorm:"type:varchar(36)" json:"notification_filter_preset_id,omitempty"\`
+  ` (nullable/empty = all alerts). Schema change is picked up by the
+  existing `AutoMigrate()` (`internal/backend/database/gorm_db.go:114`) —
+  no hand-written migration.
+- `internal/backend/services/services.go`: thread the field through
+  `GetNotificationPreferences` (~line 2520) and
+  `SaveNotificationPreferences` (~line 2566), same shape as
+  `EnabledSeverities` today. No ownership/existence check against
+  `filter_presets` server-side — the dropdown only ever offers presets the
+  session can already see (own + shared), so an invalid/foreign ID can't
+  reach the save call through the UI; the deleted-preset case is handled by
+  the client falling back to "All alerts" (Goal 5), not by backend
+  validation.
+- `internal/webui/client/backend_client.go`: add the field to the
+  `NotificationPreferences` struct (`:1440`) and thread it through
+  `GetNotificationPreferences`/`SaveNotificationPreferences` (`:1447`,
+  `:1485`), mirroring `EnabledSeverities`.
+- `internal/webui/handlers/notification_handlers.go`: pass
+  `notification_filter_preset_id` through `GetNotificationPreferences`
+  (`:15`) and `SaveNotificationPreferences` (`:44`) request/response JSON,
+  same pattern as the existing three fields.
+
+### Pure, testable matcher (new static JS, not templ-inline)
+
+The repo has no JS test runner and no `static/js/` directory today — all
+notification logic lives inline in `<script>` blocks inside `.templ` files,
+which can't be `require`d by a test. The issue explicitly asks for the
+matcher to be "a pure function so it is unit-testable", so:
+
+- New `internal/webui/static/js/notification-matcher.js` (plain JS, served
+  automatically — `r.Static("/static", ...)` at `internal/webui/router.go:159`
+  already covers the whole static tree, no router change needed). Exports
+  one function: `matchesPreset(alert, filterData)`, adapting field names
+  (`filterData.alert_names` → the `alertNames` shape `alertMatchesFilters`
+  already expects) and reusing the exact same list semantics as
+  `alertMatchesFilters` (`notification_service.templ:351`) — alertmanagers,
+  severities, statuses, teams, alertNames, empty list = no constraint.
+  This is new code only for the field-name adapter; the matching semantics
+  are copied from, not reinvented from, the existing function.
+- Load it via one `<script src="/static/js/notification-matcher.js">` tag
+  before `notification_service.templ`'s script (wherever `NotificationService()`
+  is rendered, e.g. `internal/webui/templates/pages/NewDashboard.templ`).
+- `shouldNotify(alert)` (`notification_service.templ:240`) gets one more
+  check after the severity gate: if
+  `this.preferences.notificationFilterPresetId` is set and its cached
+  `filter_data` is loaded, `window.NotificationMatcher.matchesPreset(alert,
+  filterData)` must also be true.
+- Test: `internal/webui/static/js/notification-matcher.test.js` using
+  Node's built-in test runner (`node --test`, `node:assert`) — zero new
+  dependencies. Covers match / no-match / empty-filter-data cases per the
+  acceptance criteria. Add `"test": "node --test internal/webui/static/js"`
+  to `package.json`'s `scripts`.
+
+### Frontend state + UI
+
+- `internal/webui/templates/scripts/dashboard_settings.templ`:
+  - `loadNotificationPreferences()` (`:890`) and `saveNotificationPreferences()`
+    (`:924`) gain `notificationFilterPresetId`, mirroring how
+    `enabledSeverities` is already read/written.
+  - New `loadNotificationPresetFilterData()`: fetch
+    `GET /api/v1/dashboard/filter-presets?include_shared=true` (already
+    exists, used today by `dashboard_filter_presets.templ:79`) once when the
+    Notifications settings tab is opened, to populate the `<select>` options
+    and cache the `filter_data` of whichever preset is selected. Re-fetch
+    only on selection change, not per-alert — no extra request per alert per
+    the issue's explicit constraint.
+  - If the previously-saved `notificationFilterPresetId` isn't present in
+    the fetched list (deleted preset), reset to `null`/`All alerts` client-side
+    (Goal 5) and let the next save persist that fallback.
+- `internal/webui/templates/components/notification_settings.templ`: add
+  the preset `<select>` (Alpine `x-model="notificationPreferences.notificationFilterPresetId"`,
+  options from the loaded preset list, first option "All alerts" = empty
+  string) next to the severity checkboxes (~line 87). Update the "How it
+  works" copy (`:160`, "Notifications respect your current dashboard
+  filters") to also mention the independent preset scope, since that line
+  is the one currently documenting the coupling this feature must not
+  conflate with.
+- Regenerate with `make webui-templates` — never hand-edit `*_templ.go`.
+
+## Risks / trade-offs
+
+- **Field-name drift**: `filter_data` uses `alert_names` (snake_case, see
+  `dashboard_filter_presets.templ:176`) while `alertMatchesFilters` expects
+  `alertNames`. The adapter must translate this explicitly; a naive pass-through
+  would silently no-op the alertName constraint. Covered by the unit test's
+  match/no-match cases.
+- **Stale cached filter_data**: the preset's `filter_data` is fetched once
+  per settings-modal open/selection-change, not live. If someone edits the
+  shared preset's filters in another tab, this tab's notification scope is
+  stale until the settings modal is reopened. Acceptable — same staleness
+  window the issue's "fetched once on settings load" approach already
+  accepts, and presets change far less often than alerts fire.
+- **Two independent AND gates now exist** (live-view filters in
+  `processNewAlerts`, and the new preset scope in `shouldNotify`). Worth a
+  code comment at the `shouldNotify` call site so a future reader doesn't
+  conflate or try to merge them.
+
+## Validation
+
+- `go build ./...` after proto regen (`make proto`) and template regen
+  (`make webui-templates`).
+- `node --test internal/webui/static/js` for the matcher unit test
+  (match / no-match / empty-filter-data cases, per acceptance criteria).
+- Manual (`make test` docker-compose stack):
+  1. `All alerts` selected → confirm notification behaviour unchanged.
+  2. Select a preset scoped to one team → fire/observe a non-matching
+     critical alert (no popup/sound) and a matching one (popup + sound).
+  3. Disable the matching alert's severity → confirm no notification even
+     though the preset matches (severity gate still applies on top).
+  4. Reload the page and re-login → confirm the preset selection persists.
+  5. Delete the selected preset → confirm the dropdown falls back to `All
+     alerts` with no console error.
--- a/specs/2026-08-01-152-feat-notifications-scope-browser.md
+++ b/specs/2026-08-01-152-feat-notifications-scope-browser.md
@@ -128,10 +128,31 @@ matcher to be "a pure function so it is unit-testable", so:
 
 ### Frontend state + UI
 
+- `internal/webui/templates/scripts/notification_service.templ` is the
+  actual source of truth for `this.preferences` (`shouldNotify()` at `:240`
+  reads off it directly) and must gain `notificationFilterPresetId` first:
+  - `loadPreferences()` (`:84`-`:105`) reconstructs `this.preferences` from
+    an explicit field allowlist — add
+    `notificationFilterPresetId: result.data.notification_filter_preset_id || null`
+    to it, or the field is silently dropped on every page load/reload
+    (breaks Goal 4).
+  - `savePreferences(preferences)` (`:108`-`:136`) POSTs an explicit-allowlist
+    body — add `notification_filter_preset_id: preferences.notificationFilterPresetId`
+    to it. Without this, `dashboard_core.templ`'s `enableNotifications()`
+    (`:600`-`:616`, the "Enable Notifications" banner) calls `savePreferences`
+    with `window.notificationService.preferences` as-is and would silently
+    reset an already-configured preset scope back to "All alerts" as a
+    side effect of granting browser permission.
 - `internal/webui/templates/scripts/dashboard_settings.templ`:
-  - `loadNotificationPreferences()` (`:890`) and `saveNotificationPreferences()`
-    (`:924`) gain `notificationFilterPresetId`, mirroring how
-    `enabledSeverities` is already read/written.
+  - `loadNotificationPreferences()` (`:890`) copies from
+    `window.notificationService.preferences` (documented as the "Single
+    source of truth") rather than fetching independently — once
+    `notification_service.templ`'s `loadPreferences()` carries the field
+    (above), add `notificationFilterPresetId` to both the primary copy and
+    the fallback-fetch branch here, mirroring how `enabledSeverities` is
+    already read.
+  - `saveNotificationPreferences()` (`:924`) gains `notificationFilterPresetId`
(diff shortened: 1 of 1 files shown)


specs/2026-08-01-152-feat-notifications-scope-browser.md
```diff
@@ -0,0 +1,187 @@
...
+  dependencies. Covers match / no-match / empty-filter-data cases per the
+  acceptance criteria. Add `"test": "node --test internal/webui/static/js"`
+  to `package.json`'s `scripts`.
+
+### Frontend state + UI
+
+- `internal/webui/templates/scripts/dashboard_settings.templ`:
+  - `loadNotificationPreferences()` (`:890`) and `saveNotificationPreferences()`
+    (`:924`) gain `notificationFilterPresetId`, mirroring how
+    `enabledSeverities` is already read/written.
```
SoulKyu: The plan only threads `notificationFilterPresetId` through `dashboard_settings.templ`'s `loadNotificationPreferences()`/`saveNotificationPreferences()`, but that isn't where the field actually needs to live to satisfy Goal 4 ("selection persists server-side, survives reload") and `shouldNotify()`'s read of `this.preferences.notificationFilterPresetId`.

Verified in the current code: `dashboard_settings.templ:890`'s `loadNotificationPreferences()` doesn't independently fetch — it copies from `window.notificationService.preferences` (its own comment calls this "Single source of truth"). That object is populated by `notification_service.templ:84-105`'s `loadPreferences()`, which reconstructs `this.preferences` from an explicit 3-field allowlist (`browserNotificationsEnabled`, `enabledSeverities`, `soundNotificationsEnabled` — lines 93-97) that silently drops any other field in the backend response. `shouldNotify()` (`notification_service.templ:240`) reads `this.preferences.notificationFilterPresetId` directly off that same object.

Concrete failure (this is exactly the spec's own Validation step 4 — "Reload the page ... confirm the preset selection persists"): user selects a preset scope, saves successfully (persisted server-side), then reloads the page. `loadPreferences()` fetches the backend value but its allowlist discards `notification_filter_preset_id`, so `this.preferences.notificationFilterPresetId` is `undefined` after reload. `shouldNotify()`'s preset check silently no-ops and the user is back to getting notified for every alert matching severity — Goal 4 fails.

There's a second, related site with the same root cause: `notification_service.templ:108-136`'s `savePreferences(preferences)` also POSTs an explicit-allowlist body (lines 116-120, no `notification_filter_preset_id`). It's called from `dashboard_core.templ:614-616` (`enableNotifications()`, the "Enable Notifications" banner quick-action) with `window.notificationService.preferences` as-is. If a user has already configured a preset scope and later clicks that banner to grant/re-grant permission, this silently POSTs a body without the preset field, resetting the server-side scope back to "All alerts" as a side effect of an unrelated action.

Suggested fix: add `notification_service.templ`'s `loadPreferences()` and `savePreferences()` to the list of touched files/functions in this section, with the same field added to their explicit allowlists (both the `this.preferences = {...}` reconstruction at load and the JSON.stringify body at save). Also worth stating explicitly in this section when the preset's `filter_data` cache (used by `matchesPreset`) gets populated on a fresh page load — as written, `loadNotificationPresetFilterData()` only fires "once when the Notifications settings tab is opened," so a user who reloads the page without reopening Settings this session has a valid `notificationFilterPresetId` but no cached `filter_data`, and the spec doesn't say what `shouldNotify()` should do in that state (treat as unscoped vs. fail closed).

SoulKyu (commented):
@SoulKyu Reviewed the spec for scoping browser/sound notifications to a saved filter preset. The proto/backend/model line references are all verified accurate against the current codebase, and the goal of decoupling notification scope from the live dashboard filter is well-motivated. One gap in the "Frontend state + UI" section needs to be closed before implementation starts — see the inline comment.



specs/2026-08-01-152-feat-notifications-scope-browser.md
SoulKyu: @SoulKyu This has been fixed in cf748b4.

Added notification_service.templ's loadPreferences()/savePreferences() allowlist updates to the spec's Frontend state + UI section, noted the enableNotifications() banner side-effect risk, and documented the reload-vs-tab-open filter_data cache gap with a fail-open (unscoped) resolution.




SoulKyu:
**Looper fixer round complete** — cf748b4

- ✅ Review comment on `specs/2026-08-01-152-feat-notifications-scope-browser.md` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/155#discussion_r3694750899)
  - Added notification_service.templ's loadPreferences()/savePreferences() allowlist updates to the spec's Frontend state + UI section, noted the enableNotifications() banner side-effect risk, and documented the reload-vs-tab-open filter_data cache gap with a fail-open (unscoped) resolution.


SoulKyu:
## 🧪 QA Report

Spec/docs-only PR (`looper/planner/*`, diff = 1 file under `specs/`) → the feature is intentionally
not implemented, so I graded the **spec** against the live codebase at `25f7db3`, booted from
`docker-build-all` + `docker compose up -d` (host `make test` dies at `webui-templates`: `templ`
not on PATH, Error 127 — the Dockerfile installs it itself, so the images build fine).

| Criterion (spec claim) | Result | Evidence |
|---|---|---|
| Every `file:line` pointer is accurate at this SHA | ✅ | Checked all 16: `notification_service.templ:240/351`, `dashboard_data.templ:454`, `notification_settings.templ:160`, `dashboard_handlers.go:403`, `gorm_db.go:114`, `services.go:2520/2566`, `backend_client.go:1440/1447/1485`, `notification_handlers.go:15/44`, `router.go:159`, `dashboard_settings.templ:890/924`, `dashboard_filter_presets.templ:79/176`, `alert.proto:467/480` — all exact. |
| `proto/alert.proto`: add the field with the numbers given | ❌ | Spec says `string notification_filter_preset_id = 4;` in `SaveNotificationPreferencesRequest`, but `sound_notifications_enabled = 4` already exists (`alert.proto:471`). Reproduced with the spec's exact snippet: `protoc` → `Field number 4 has already been used in "SaveNotificationPreferencesRequest" by field "sound_notifications_enabled"`. `make proto` fails as written; next free number is 5. (The `= 8` for `NotificationPreference` is correct.) |
| New column round-trips via existing `AutoMigrate()`, no hand-written migration | ✅ | Tested on an **existing, populated** DB: `ALTER TABLE notification_preferences DROP COLUMN sound_notifications_enabled` → `docker restart notificator-backend` → column recreated by AutoMigrate, existing row preserved. |
| `shouldNotify()` gate reads `this.preferences.notificationFilterPresetId` | ❌ | The spec's file list never touches `notification_service.templ:84 loadPreferences()`, which builds `this.preferences` as a hard-coded 3-field literal. Live on `/dashboard` (Playwright, logged-in user with a saved preference): `Object.keys(notificationService.preferences)` = `["browserNotificationsEnabled","enabledSeverities","soundNotificationsEnabled"]` and `preferences.notificationFilterPresetId` is `undefined`. Implemented as written, the preset gate is skipped on every page load → **fail-open, every alert notifies**. For the same reason `dashboard_settings.templ:890` ("mirroring `enabledSeverities`") reads the field from `svcPreferences`, which will never carry it → the dropdown also shows `All alerts` after reload. Breaks issue AC #2 and #4. |
| `filter_data` fetched "once when the Notifications settings tab is opened" is available to the notification path | ❌ | `shouldNotify()` runs from page load on every SSE alert; per the spec the cache lives in the settings-modal component and is only populated when the Notifications tab is opened. The spec's own guard ("if the preset id is set **and its cached `filter_data` is loaded**") then fails open, not closed. The spec also never says where that cache lives or how `notificationService` reaches it. (`window.currentSettingsModal` does exist at page load, so an `init()`-time fetch would work — but that is not what the spec prescribes.) |
| Non-goal "don't match the preset's `search` field", justified by existing matcher precedent | ❌ | The justification is false for the server-side function it cites: `alertPassesAlertLevelFilters` (`dashboard_handlers.go:404`) applies `filters.Search` as its **first** check. (It also skips `statuses`, which the client-side `alertMatchesFilters` does check — the two are not "the same list semantics".) Consequence: issue #152 defines preset scope as "alertmanagers / severities / statuses / teams / alertNames **+ search**"; a preset whose scope is expressed as a search term (a first-class preset field — saved at `dashboard_filter_presets.templ:294`, restored at `:171`) would notify on everything the label lists allow. Scope dropped from the issue's AC on an inaccurate premise. |
| `node --test internal/webui/static/js` is a working validation command | ❌ | Reproduced on Node v24.14.1 (the repo pins no Node version — `Dockerfile.webui` installs alpine `nodejs`, the runtime image has none): `node --test <dir>` no longer discovers recursively, it tries to execute the directory → `Error: Cannot find module '.../js'`. Same for the `package.json` script the spec adds. `node --test "internal/webui/static/js/**/*.test.js"` passes (`fail 0`). |
| Matcher is both a browser global and unit-testable with `node --test` | ❌ | The spec asks for `window.NotificationMatcher.matchesPreset` loaded via `<script src>` **and** a `node --test` unit test, with no module-export guard. Built the file exactly as described and `require`d it: `ReferenceError: window is not defined`. Needs an explicit `typeof module !== 'undefined' && (module.exports = …)` (or an ESM/CJS split) that the spec does not mention. |
| New `static/js/` subtree is served with no router change | ✅ | `docker exec … mkdir -p internal/webui/static/js` + `docker cp notification-matcher.js` into the running webui → `GET /static/js/notification-matcher.js` → **200** with the file body. `r.Static("/static", staticPath)` (`router.go:159`) plus `COPY … /app/internal/webui/static` (`Dockerfile.webui:55`) confirm it. |
| Risk "field-name drift `alert_names` vs `alertNames`" is real | ✅ | Verified live in the page: `notificationService.alertMatchesFilters(alert, {severities:["critical"],teams:["team-a"],alert_names:["DiskFull"]})` → **`true`** for an alert named `SomethingElse` (constraint silently no-ops), vs `false` with `alertNames`. Also confirmed the API shape: `POST /api/v1/dashboard/filter-presets` returns `filter_data: {"severities":…,"teams":…,"alert_names":…}` and **omits empty lists entirely** — the copied `filters.x && filters.x.length > 0` semantics handle the missing keys, so that part holds. |
| Preset dropdown source endpoint exists (own + shared) | ✅ | `GET /api/v1/dashboard/filter-presets?include_shared=true` → 200 with the preset I created; route registered at `router.go:289`. |
| No server-side ownership check needed (client can only offer visible presets) | ✅ | Weak but sound as argued: `POST /api/v1/notifications/preferences` with `notification_filter_preset_id:"deadbeef"` is accepted and silently dropped today; once threaded through, a forged/foreign id still is not in the list the client fetches, so it degrades to `All alerts`. No by-id preset GET route exists (`router.go:289-294`), so no read amplification. |

**Adversarial pass:**
- *Boundary/empty*: empty preset list (`presets: []` for a fresh user); preset saved with empty `alertmanagers`/`statuses`/`search` → the API drops those keys entirely, so the matcher must survive **missing** keys, not just empty arrays (the copied semantics do — no finding).
- *Malformed/hostile*: `notification_filter_preset_id:"deadbeef"` (non-existent id) posted directly to the prefs endpoint — accepted, dropped today, harmless under the spec's design; a 32-char preset id fits the proposed `varchar(36)` (`FilterPreset.ID` is `varchar(32)`, `models.go:180`); fed the raw snake_case `filter_data` straight into `alertMatchesFilters` — silent no-op, exactly the drift the spec flags.
- *Compile/codegen path*: ran the spec's proto snippet through `protoc` → hard failure (finding above). Ran the spec's `node --test` command and built the spec's matcher file verbatim → both fail (findings above).
- *Error/unavailable path*: the fail-open direction is the real damage here — with the preset id or the cached `filter_data` missing (page load, modal never opened, prefs fetch 401/503), the spec's guard **notifies for everything** instead of honouring the user's scope. The prefs fetch is a plain `fetch` with no retry (`notification_service.templ:84`) and leaves `preferences` at its 3-key default on failure.
- *Auth/scope*: unauthenticated `GET /api/v1/notifications/preferences` and `/dashboard/filter-presets` → `{"success":false,"error":"Authentication required"}`; shared-preset path exercised via `include_shared=true`.
- *Adjacent flows*: confirmed the spec does not disturb the existing `processNewAlerts` → `alertMatchesFilters(alert, currentFilters)` coupling (`notification_service.templ:425`, called from `dashboard_data.templ:454`) — it adds a second AND gate, as claimed; dashboard render path untouched. No new console errors on `/dashboard` beyond the pre-existing 401s during login bootstrap.
- *Persisted state*: exercised AutoMigrate against an **existing** table holding a live row (dropped a column, restarted the backend, column restored, row intact) rather than against a fresh DB.

Net: the spec's codebase pointers are unusually precise and two of its three named risks check out, but four of its instructions are wrong as written — one makes `make proto` fail outright, two make the feature silently fail **open** (every alert notifies) instead of scoping, one drops `search` from the preset scope on a false reading of `alertPassesAlertLevelFilters`, and the JS test setup it prescribes does not run.

**Verdict: FAIL**

SoulKyu:
## 🧪 QA Report

Spec/docs-only planner PR (diff = 1 file under `specs/`), so I graded the **spec** against the live
codebase at `cf748b4`, not the (intentionally unimplemented) feature. Booted via `make docker-build-all`
+ `docker compose up -d` (host `make test` still dies at `webui-templates`: `templ` not on PATH, Error
127 — the Dockerfile installs it itself, so the images build fine). Re-ran the full grid at this SHA.

`cf748b4` **fixes 1 of the 6 findings** from the `25f7db3` report (the `this.preferences` threading, and
well — see row 4). The other five instructions are unchanged and still wrong.

| Criterion (spec claim) | Result | Evidence |
|---|---|---|
| Every `file:line` pointer is accurate at this SHA | ✅ | Re-checked all 20, including the 3 the fix commit added: `notification_service.templ:240/351`, `loadPreferences` `:84`-`:105`, `savePreferences` `:108`-`:136`, `dashboard_core.templ:616` (the exact `savePreferences` call), `dashboard_settings.templ:890/924`, `notification_settings.templ:87/160`, `dashboard_handlers.go:403`, `gorm_db.go:114`, `services.go:2520/2566`, `backend_client.go:1440/1447/1485`, `notification_handlers.go:15/44`, `router.go:159`, `dashboard_filter_presets.templ:79/176`, `alert.proto:467/480`, and `@scripts.NotificationService()` at `NewDashboard.templ:960` (the only render site). All resolve. |
| **NEW in `cf748b4`** — `notification_service.templ` is the real source of truth and must gain the field first | ✅ | Correct and materially important. Verified live on `/dashboard`: `Object.keys(notificationService.preferences)` = `["browserNotificationsEnabled","enabledSeverities","soundNotificationsEnabled"]` — both `loadPreferences()` and `savePreferences()` are explicit allowlists, so the field would be dropped without this. The `enableNotifications()` hazard it newly calls out is real: `dashboard_core.templ:616` passes `window.notificationService.preferences` wholesale into the 3-field allowlist, so granting browser permission would silently reset a configured preset to "All alerts". Good catch; this closes the prior report's row 4. |
| `dashboard_settings.templ` "Single source of truth" description + fallback-fetch branch | ✅ | Verbatim accurate: `loadNotificationPreferences()` (`:890`) copies from `window.notificationService.preferences` behind a `preferencesLoaded` guard and has a second direct-fetch fallback branch; the spec correctly instructs adding the field to **both**. |
| `proto/alert.proto`: add the field with the numbers given | ❌ | **Unchanged from the last report.** Spec still says `string notification_filter_preset_id = 4;` in `SaveNotificationPreferencesRequest`, but `sound_notifications_enabled = 4` already occupies it (`alert.proto:471`). Applied the spec's snippet verbatim and ran `protoc`: `alert.proto:472:42: Field number 4 has already been used in "notificator.alert.SaveNotificationPreferencesRequest" by field "sound_notifications_enabled"`. `make proto` fails outright; next free number is 5. (The `= 8` on `NotificationPreference` is correct and compiles.) |
| Preset scope is actually enforced after a reload (issue AC #2) | ❌ | The fix commit *documents* the cache gap but resolves it the wrong way: `filter_data` is still fetched only when the Notifications tab is opened, and the spec now explicitly prescribes **fail-open** ("treat this as unscoped… rather than failing closed"). Implemented the spec's guard verbatim in the live page: preset `team-a-scope` selected, plain reload, settings tab never opened → `shouldNotify({alertName:'PaymentsDown',severity:'critical',team:'team-b'})` returns **`true`** (notifies). Same alert returns `false` only after the settings tab is opened. Since the modal is never opened in normal operation, the feature is **inert by default** — the on-call engineer still gets paged for every team's criticals, which is the exact 3am failure the issue exists to fix. The "fail closed vs fail open" framing is a false dichotomy: `loadPreferences()` already runs at page init and `window.currentSettingsModal` exists at load (verified), so fetching the selected preset's `filter_data` at init makes it work. The spec also still never says **where** the cache lives or how `notificationService` reaches it from the settings-modal component. |
| Non-goal "don't match the preset's `search` field", justified by matcher precedent | ❌ | **Unchanged.** The justification ("both deliberately treat search as a separate, fuzzy concern") is false for the server-side function it cites: `alertPassesAlertLevelFilters` (`dashboard_handlers.go:404`) applies `filters.Search` as its **first** check. Issue #152 defines preset scope as "alertmanagers / severities / statuses / teams / alertNames **+ search**", and `search` is a first-class, persisted preset field — confirmed live: `POST /api/v1/dashboard/filter-presets` with `search:"payments"` round-trips as `filter_data:{"search":"payments",…}`. A preset whose scope is expressed as a search term would notify on everything the label lists allow. Scope dropped from the issue's AC on an inaccurate premise. |
| `node --test internal/webui/static/js` is a working validation command | ❌ | **Unchanged.** Reproduced on Node v24.14.1 (repo pins no Node version): `node --test <dir>` no longer discovers recursively, it tries to execute the directory → `Error: Cannot find module '/tmp/qa155/js'`, `code: 'MODULE_NOT_FOUND'`, `fail 1`. Affects both the Validation section and the `"test"` script the spec adds to `package.json`. `node --test "internal/webui/static/js/**/*.test.js"` is the working form. |
| Matcher is both a browser global and unit-testable with `node --test` | ❌ | **Unchanged.** Built the file exactly as prescribed (`window.NotificationMatcher = {…}`, no export guard) and `require`d it from a `node:test` file: `ReferenceError: window is not defined`, `fail 1`. Needs an explicit `typeof module !== 'undefined' && (module.exports = …)` (or an ESM/CJS split) that the spec never mentions. |
| Risks section names the real traps | ❌ | It names 3 and misses a 4th that its own design creates: `shouldNotify()` normalizes severity (`critical-daytime`→`critical`, `information`→`info`, `:255`-`:261`) but `alertMatchesFilters` — whose semantics the spec says to copy "exactly" — does **not**. Verified live: `alertMatchesFilters({severity:'critical-daytime',team:'team-a'}, {severities:['critical']})` → **`false`**, and `{severity:'information'}` vs `['info']` → **`false`**. So a user with `enabled_severities:['critical']` who picks a preset scoped to `severities:['critical'] + teams:['team-a']` **silently stops being notified** for `critical-daytime` alerts they get today. `critical-daytime` is first-class here (`dashboard_data.templ:474`, `dashboard_utilities.templ:563`, `StatisticsDashboard.templ:1825`). The implementer gets no warning. |
| New column round-trips via existing `AutoMigrate()`, no hand-written migration | ✅ | Re-tested on an **existing, populated** DB: `ALTER TABLE notification_preferences DROP COLUMN sound_notifications_enabled` → `docker restart notificator-backend` → column recreated by AutoMigrate, existing row (`eb521ec2…`) preserved intact. Model path `internal/backend/models/notification_preference.go:40` confirmed. |
| New `static/js/` subtree is served with no router change | ✅ | `docker exec … mkdir -p /app/internal/webui/static/js` + `docker cp` the matcher into the running webui → `GET /static/js/notification-matcher.js` → **HTTP 200** with the file body. `r.Static("/static", staticPath)` (`router.go:159`) covers it. |
| Risk "field-name drift `alert_names` vs `alertNames`" is real | ✅ | Confirmed live against the real function: `alertMatchesFilters({alertName:'SomethingElse'}, {alert_names:['DiskFull']})` → **`true`** (constraint silently no-ops) vs `{alertNames:['DiskFull']}` → **`false`**. Model confirms the key: `FilterPresetData.AlertNames` is `json:"alert_names,omitempty"` (`internal/webui/models/filter_preset.go:27`). |
| Preset dropdown source endpoint exists (own + shared) | ✅ | `GET /api/v1/dashboard/filter-presets?include_shared=true` → 200 with the preset I created; routes at `router.go:289-294`. Alpine scope name `notificationPreferences` matches the existing `enabledSeverities` `x-model` at `notification_settings.templ:94`. |
| No server-side ownership check needed | ✅ | `POST /api/v1/notifications/preferences` with `notification_filter_preset_id:"deadbeef-not-a-real-preset"` → accepted, silently dropped (response carries only the 3 existing fields). Once threaded through, a forged/foreign id still isn't in the list the client fetches, so it degrades to "All alerts". No by-id preset GET route exists, so no read amplification. |

**Adversarial pass:**
- *Boundary/empty*: `filter_data` as `{}`, `null`, and all-empty-arrays → `alertMatchesFilters` returns `true` in all three (no constraint), so the "empty list = no constraint" semantics the spec leans on do hold. Critically, the server uses `omitempty`: I created a preset with `severities:[]` and the key was **omitted entirely** from the persisted `filter_data`, so the matcher must survive *missing* keys, not just empty ones — the copied `filters.x && filters.x.length > 0` guard does. No finding.
- *Malformed/hostile*: posted a non-existent preset id to the prefs endpoint (accepted, dropped, degrades safely); sent `alertNames` (camelCase) on preset **create** — the server silently drops it, only `alert_names` persists, which is a second face of the drift risk; fed raw snake_case `filter_data` straight into `alertMatchesFilters` → silent no-op (row above). A 32-char preset id (`9a16rie9…`) fits the proposed `varchar(36)`.
- *Compile/codegen path*: ran the spec's proto snippet through `protoc` → hard failure. Ran the spec's `node --test` command and built the spec's matcher file verbatim → both fail. These three are the ones a reviewer reading the spec would not catch by eye.
- *Error/unavailable path*: the fail-open direction is the real damage — with the preset id set but `filter_data` uncached (every page load until the modal is opened, or a prefs fetch 401/503), the spec's guard **notifies for everything**. `loadPreferences()` is a bare `fetch` with no retry and leaves `preferences` at its 3-key default on failure.
- *Auth/scope*: unauthenticated `GET /api/v1/notifications/preferences` and `/dashboard/filter-presets?include_shared=true` → `{"success":false,"error":"Authentication required"}` on both.
- *Adjacent flows*: confirmed the spec does not disturb the existing `processNewAlerts` → `alertMatchesFilters(alert, currentFilters)` coupling (`notification_service.templ:425`, driven from `dashboard_data.templ:454`) — it adds a second, separate AND gate as claimed. Dashboard render path untouched; no new console errors on `/dashboard` beyond the pre-existing 401s during login bootstrap.
- *Persisted state*: exercised AutoMigrate against an **existing** table holding a live row rather than a fresh DB (drop column → restart → column restored, row intact).
- *Severity normalization*: probed both special cases the notification path handles (`critical-daytime`, `information`) against the matcher the spec says to copy — both mismatch (row above).

Net: the pointers remain unusually precise and `cf748b4` genuinely fixed the `this.preferences` threading (including a sharp catch on `enableNotifications()` silently resetting the scope). But the five previously-reported instruction defects are untouched, and the one the fix commit did address it resolved by **specifying** fail-open — so an implementer following this spec would ship a feature that fails `make proto`, fails its own test command, and silently doesn't scope notifications until the user opens the settings modal.

🤖 Fixer summoned to address the findings.

**Verdict: FAIL**

SoulKyu:

SoulKyu:
Closing: spec consumed — implementation PR #158 is open and in the review cycle (QA grades acceptance criteria from issue #152, not from this PR). Spec text remains on the branch.

PR #159, opened 2026-08-01
feat(devtools): factory TUI shift timeline — per-agent activity strip

### LINKED ISSUE #150: feat(devtools): factory TUI shift timeline — per-agent activity strip for the last hours
## What & why

The office shows the **present** (who is typing, who is asleep) and the 24h
GitHub **aggregate** (🏆 scoreboard sparkline), but nothing shows *who worked
when*. An operator watching the loop regularly asks questions the TUI cannot
answer today:

- did `scout` actually run this morning, or has its timer been dead since 9h?
- `roast` has been `veille` for a while — when was its last run?
- is one agent hogging the machine (back-to-back runs) while another never fires?

A one-row-per-agent **shift timeline** answers all three at a glance, from data
already on disk. It is the office's *timesheet board*, complementary to the
scoreboard: the scoreboard counts GitHub outcomes, the timeline shows local
agent occupancy — a scout that ran 5 times and produced nothing is invisible on
the scoreboard and obvious here.

## Approach

**Poller**: none new. `poll_med()` already does `os.listdir(LOG_DIR)` for the 📻
ticker. Extend that same scan: agent log files are named
`<agent>-<YYYYMMDDThhmmss>.log` (e.g. `qa-131-20260731T094720.log`) — the
filename gives the run **start**, `os.path.getmtime` gives the last write
(≈ run **end**). Bucket each run into fixed slots over the last `N` hours and
store one string per ROSTER key in `STATE["shifts"]`, computed in a pure helper
(`compute_shifts(entries, now)`) so `--check` can assert on it without a
filesystem.

Roster mapping reuses the existing `LOG_PREFIX_OF` indirection, and keys like
`qa-131` collapse to their agent (`qa`) by stripping a trailing numeric segment.

**Render**: a `⏱ TIMELINE` panel, one line per agent that has any run in the
window, built through `dpad()` like every other panel:

```
⏱ TIMELINE  ── 6h ─────────────── 09h ──── 11h ──── 13h ─── now ┐
🔍 SCOUT    ░░░░░▓▓░░░░░░░░░░░░░░▓▓▓░░░░░░░░░░░░░░░░░░░░░░░░░░░░
🔥 ROAST    ░░░░░░▓░░░░░░░░░░░░░░░▓▓░░░░░░░░░░░░░░░░░░░░░▓▓▓░░░░
🧪 QA       ░▓▓░░▓▓░░░▓▓░░░░▓░░░░░▓░░░░░▓▓░░░░▓▓▓░░░░░░░░░▓░░░░░
```

Placement: below the 🏆 SCOREBOARD, in the same panel stack — so it goes through
the existing `panel_rows()` / `desk_cap()` height budgeting and simply drops
out first on short terminals (empty window → no panel, like the scoreboard).
Slot width is derived from the panel width (`window_hours * 3600 / cols`), so
it degrades from ~7 min/char on a wide terminal to ~20 min/char on 80 cols
instead of breaking alignment. Window configurable via `FACTORY_TIMELINE_H`
(default 6), clamped like `FACTORY_STALL_MIN` does.

Stdlib only (`os.scandir`, `time`), read-only (stat + filename parsing, no
reads of log contents beyond what the ticker already does), no new subprocess.

## Acceptance criteria

- [ ] A `⏱ TIMELINE` panel renders one row per agent with at least one run in the window, ordered like the desk roster.
- [ ] Run extents come from the log **filename timestamp** (start) and **mtime** (end); a run shorter than one slot still marks exactly one slot (never invisible).
- [ ] Missing/unreadable `FACTORY_LOG_DIR` → no panel, no traceback (same graceful degradation as the ticker).
- [ ] Empty window (no run in the last N hours) → panel hidden entirely, not an empty box.
- [ ] Slot count adapts to the panel width; all frame rows keep an identical display width.
- [ ] `FACTORY_TIMELINE_H` overrides the window; a garbage value falls back to the default without crashing.
- [ ] `python3 factory-tui.py --once` still prints a frame and exits 0.
- [ ] `python3 factory-tui.py --check` still passes, with a new assertion on `compute_shifts()` over synthetic entries (fixed `now`, no filesystem).
- [ ] No new dependency, no write outside the process, `factory.py --check` still green (data core unchanged for existing callers).

## Out of scope

- Any change to the 🏆 scoreboard (aggregate GitHub stats stay as they are).
- Looper run history (`~/.looper/logs/loops`) — filename/mtime semantics differ; a follow-up can add looper desks to the same panel once this shape is validated.
- Interactive scrubbing / zooming into a past slot, clicking a slot to open its log.
- Persisting history across TUI restarts beyond what the log dir already retains.

SoulKyu:
## Summary

- Adds a `⏱ TIMELINE` panel to `factory-tui.py`, below the `🏆 SCOREBOARD`: one row per agent that ran in the last `FACTORY_TIMELINE_H` hours (default 6), reading the same log-dir scan `poll_med()` already does for the `📻` ticker — no new poller, no new dependency, no writes.
- Run extents come from the log filename (`<agent>-<ts>.log`, start) and its mtime (end); a run shorter than one slot still marks exactly one slot.
- Bucketing lives in a pure `compute_shifts(entries, now, window_hours, cols)` helper, independent of the filesystem, so `--check` asserts on it directly with synthetic entries.
- Slot count adapts to the panel width (`shift_rows()`); the set of agents shown never depends on the column count, so `desk_cap()`'s row budget and the renderer always agree.
- Empty window (nobody ran) → panel hidden entirely, same rule as the scoreboard. Missing/unreadable `FACTORY_LOG_DIR` degrades gracefully (same `try/except` as the ticker) — no panel, no traceback.
- `FACTORY_TIMELINE_H` overrides the window, clamped like `FACTORY_STALL_MIN`; a garbage value falls back to the default (6h) instead of crashing.

Fixed along the way: agent log filenames are wall-clock local time (`date +%Y%m%dT%H%M%S`, no `-u`, matching how `systemd_ts()` already parses systemd timestamps), so run starts are parsed with `time.mktime`, not `calendar.timegm` — the UTC parser silently pushed every run 1-2h into the future and made the panel disappear on a live system.

## Test plan

- [x] `python3 devtools/factory-tui/factory-tui.py --check` — selfcheck: OK (includes new `compute_shifts`/`agent_of_log`/`timeline_hours` assertions and a frame-width sweep with the panel populated)
- [x] `python3 devtools/factory-tui/factory-tui.py --once` — prints a frame, exits 0
- [x] `uv run devtools/factory-tui/factory.py --check` — selfcheck: OK (data core unchanged for existing callers)
- [x] Manually verified against the real `~/.claude-agents/notificator/logs` on this box — the panel renders live SCOUT/QA/REBASE/IDEA-P/IDEA-T activity for the last 6h

Closes #150

### DIFF
--- a/devtools/factory-tui/factory-tui.py
+++ b/devtools/factory-tui/factory-tui.py
@@ -106,6 +106,8 @@ STATE = {"loops": [], "svc": {}, "timers": {}, "prs": [], "issues": "", "ticker"
          "mail_pending": {}, "intercom": [], "score": None, "events": [], "pending": [],
          "ps_ok": True, "conveyor": None, "shifts": []}
 PENDING_MAX = 5
+TIMELINE_ROWS = 5  # panel height cap, same reason as ALARM_ROWS: a full roster must not outgrow
+                   # the always-on wall board/radio for the room desk_cap's len(ROSTER) floor leaves
 LOCK = threading.Lock()
 
 # one-shot animations, consumed by the render loop (render-side state only)
@@ -993,7 +995,7 @@ def panel_rows(p):
         n += 2 if p["score"] == "err" else 4
     n_shift = len(shift_rows(p, 1))  # cols=1: cheapest resolution that still gets the row count right
     if n_shift:
-        n += 1 + n_shift
+        n += 1 + min(n_shift, TIMELINE_ROWS) + (n_shift > TIMELINE_ROWS)
     if p["alarms"]:
         shown = alarm_shown(p["alarms"])
         n += 1 + shown + (len(p["alarms"]) > shown)
@@ -1063,8 +1065,10 @@ def render_frame(tick, width=92, sel=None, desks=None, panels=None, flash=False)
     shifts = shift_rows(p, cols)
     if shifts:  # nobody ran in the window → no panel, same rule as the scoreboard
         rows.append(("│" + dpad(f" ═══ ⏱ TIMELINE {TIMELINE_H}h ", width - 2, fill="═") + "│", 0))
-        for key, emoji, name, bar in shifts:
+        for key, emoji, name, bar in shifts[:TIMELINE_ROWS]:
             rows.append(("│ " + dpad(dpad(f"{emoji} {name}", label_w) + bar, width - 4) + " │", 0))
+        if len(shifts) > TIMELINE_ROWS:  # capped: a full roster must not evict the wall board/radio
+            rows.append(("│ " + dpad(f"… +{len(shifts) - TIMELINE_ROWS} autres", width - 4) + " │", 0))
     alrm = p["alarms"]  # from the snapshot: an alarm raised mid-frame must not outgrow the budget
     if alrm:  # no alarm → no panel
         on = flash and tick % 2 == 0
@@ -1890,13 +1894,15 @@ def selfcheck():
         STATE["svc"] = {f"notificator-{u}": {"Id": f"notificator-{u}.service", "Result": "exit-code",
                                              "ExecMainStatus": "1", "ExecMainCode": "1"}
                         for u in ("scout", "qa", "rebaser", "promoter", "docagent", "reporter", "groomer")}
+        # …and the TIMELINE panel: a full roster running charges its capped 1 + TIMELINE_ROWS + tail
+        # rows too, or this sweep would never notice TIMELINE pushing the board/radio off-screen
+        STATE["shifts"] = [(key, time.time() - 300, time.time() - 60) for key, _, _, _ in ROSTER]
     term_w = 120
     # sweep the heights: the budget boundary only shows up when (h - PANEL_ROWS) % 8 == 0,
     # so a single height silently passes an off-by-one that eats the closing border.
-    # 66 is the floor's own limit: len(ROSTER) desks are 4 grid rows, 32 + 33 panel rows = 65 frame
-    # rows (33 = 5 conveyor lanes + every other panel), and main_curses draws h-1 of them. Below
+    # 73 is the floor's own limit at this worst-case panel_rows (desks pinned at len(ROSTER)); below
     # that no cap fits the panels, so nothing to assert.
-    for term_h in range(66, 82):
+    for term_h in range(73, 89):
         panels = panel_snapshot()
         cap = desk_cap(term_h, grid_per_row(term_w), panels)
         capped = build_desks(cap)
@@ -1925,9 +1931,10 @@ def selfcheck():
     # the budget and the frame must come out of one read of STATE. Cap on an empty office, let a
     # poll fill every panel in the gap, then draw: re-reading STATE here would emit the 16 rows the
     # budget never granted and push the board past h-1, so the frame must honour the snapshot.
-    with LOCK:  # svc too: it feeds alarms(), and leaving it set makes the later poll fixtures order-dependent
+    with LOCK:  # svc/shifts too: they feed alarms()/TIMELINE, and leaving them set makes the later
+                # poll fixtures order-dependent
         STATE["pending"], STATE["intercom"], STATE["err"], STATE["score"] = [], [], "", None
-        STATE["svc"] = {}
+        STATE["svc"], STATE["shifts"] = {}, []
     quiet, term_h = panel_snapshot(), 60
     capped = build_desks(desk_cap(term_h, grid_per_row(term_w), quiet))
     with LOCK:  # the poller lands between desk_cap() and render_frame()
--- a/devtools/factory-tui/factory-tui.py
+++ b/devtools/factory-tui/factory-tui.py
@@ -106,8 +106,9 @@ STATE = {"loops": [], "svc": {}, "timers": {}, "prs": [], "issues": "", "ticker"
          "mail_pending": {}, "intercom": [], "score": None, "events": [], "pending": [],
          "ps_ok": True, "conveyor": None, "shifts": []}
 PENDING_MAX = 5
-TIMELINE_ROWS = 5  # panel height cap, same reason as ALARM_ROWS: a full roster must not outgrow
-                   # the always-on wall board/radio for the room desk_cap's len(ROSTER) floor leaves
+TIMELINE_ROWS = 5  # panel height cap once shown, same reason as ALARM_ROWS. Capping alone doesn't
+                   # stop desk_cap()'s len(ROSTER) floor from evicting the wall board/radio at
+                   # realistic terminal sizes — timeline_fits() hides the whole panel for that.
 LOCK = threading.Lock()
 
 # one-shot animations, consumed by the render loop (render-side state only)
@@ -983,19 +984,41 @@ def alarm_shown(alrm):
     return ALARM_ROWS if len(alrm) > ALARM_ROWS + 1 else len(alrm)
 
 
-def panel_rows(p):
+def desk_floor_rows(per_row):
+    """Desk-grid rows that can never shrink away: build_desks()/desk_cap() both floor the grid at
+    one desk per ROSTER role (`the roster itself is never trimmed`), so this many rows are always
+    drawn no matter how tall the panel stack below them wants to be."""
+    return -(-len(ROSTER) // per_row) * 8
+
+
+def timeline_fits(p, h, per_row):
+    """Would drawing the TIMELINE panel still leave room for every other panel and the always-on
+    wall board/radio at this terminal size? TIMELINE is the newest, least essential panel, so it
+    hides itself rather than let desk_cap()'s len(ROSTER) floor evict what must stay on screen —
+    the same rule panel_rows()/render_frame() already use for an empty shift window.
+    h/per_row unknown (older callers, --once mode, most of the self-check) → always fits."""
+    if h is None or per_row is None:
+        return True
+    n_shift = len(shift_rows(p, 1))
+    if not n_shift:
+        return True
+    shift_cost = 1 + min(n_shift, TIMELINE_ROWS) + (n_shift > TIMELINE_ROWS)
+    rest = panel_rows({**p, "shift_entries": []}, h, per_row)  # every other panel, TIMELINE excluded
+    return rest + shift_cost + desk_floor_rows(per_row) <= h - 1
+
+
+def panel_rows(p, h=None, per_row=None):
     """Non-office rows of the frame about to be drawn, from a panel_snapshot. Five of the six
     panels are conditional in render_frame, so a worst-case constant charges ~27 rows that are
     usually not drawn and pins the cap at len(ROSTER) below a 55-row terminal — the dynamic desks
-    would never appear."""
+    would never appear. TIMELINE is counted last, gated by timeline_fits(): unlike the other
+    panels it has no fixed budget slice of its own, so it must know the total everything else
+    (including the desk floor) already committed to before claiming its own rows."""
     n = 1 + 7 + 1 + 1  # title + wall board (header + issues + 5 conveyor lanes) + radio + bottom
     if p["intercom"]:
         n += 1 + len(p["intercom"])
     if p["score"]:
         n += 2 if p["score"] == "err" else 4
-    n_shift = len(shift_rows(p, 1))  # cols=1: cheapest resolution that still gets the row count right
-    if n_shift:
-        n += 1 + min(n_shift, TIMELINE_ROWS) + (n_shift > TIMELINE_ROWS)
     if p["alarms"]:
         shown = alarm_shown(p["alarms"])
         n += 1 + shown + (len(p["alarms"]) > shown)
@@ -1003,6 +1026,9 @@ def panel_rows(p):
         n += 1 + min(len(p["pending"]), PENDING_MAX) + (len(p["pending"]) > PENDING_MAX)
     if p["err"]:
         n += 1
+    n_shift = len(shift_rows(p, 1))  # cols=1: cheapest resolution that still gets the row count right
+    if n_shift and timeline_fits(p, h, per_row):
+        n += 1 + min(n_shift, TIMELINE_ROWS) + (n_shift > TIMELINE_ROWS)
     return n
 
 
@@ -1010,10 +1036,10 @@ def desk_cap(h, per_row, p):
     """Desks that still leave room for the panels below the office — an unbounded grid silently
     swallows the wall board and the radio. Budget is h-1: main_curses stops drawing at h-1, so
     on `h - panels` landing on a multiple of 8 an h-row frame loses its closing border."""
-    return max(len(ROSTER), max(1, (h - 1 - panel_rows(p)) // 8) * per_row)
+    return max(len(ROSTER), max(1, (h - 1 - panel_rows(p, h, per_row)) // 8) * per_row)
 
 
-def render_frame(tick, width=92, sel=None, desks=None, panels=None, flash=False):
+def render_frame(tick, width=92, sel=None, desks=None, panels=None, flash=False, h=None):
     desks = build_desks() if desks is None else desks
     rows = []
     t = time.strftime("%H:%M:%S")
@@ -1063,11 +1089,13 @@ def render_frame(tick, width=92, sel=None, desks=None, panels=None, flash=False)
     label_w = 10  # dwidth(emoji)=2 + space + 6-char name (the roster's longest) + space
     cols = max(1, (width - 4) - label_w)
     shifts = shift_rows(p, cols)
-    if shifts:  # nobody ran in the window → no panel, same rule as the scoreboard
+    # nobody ran in the window → no panel, same rule as the scoreboard; doesn't fit this terminal
+    # (timeline_fits) → no panel either, so it can never be the reason the board/radio get cut
+    if shifts and timeline_fits(p, h, per_row):
         rows.append(("│" + dpad(f" ═══ ⏱ TIMELINE {TIMELINE_H}h ", width - 2, fill="═") + "│", 0))
         for key, emoji, name, bar in shifts[:TIMELINE_ROWS]:
             rows.append(("│ " + dpad(dpad(f"{emoji} {name}", label_w) + bar, width - 4) + " │", 0))
-        if len(shifts) > TIMELINE_ROWS:  # capped: a full roster must not evict the wall board/radio
+        if len(shifts) > TIMELINE_ROWS:  # capped: a full roster must not outgrow its own row budget
             rows.append(("│ " + dpad(f"… +{len(shifts) - TIMELINE_ROWS} autres", width - 4) + " │", 0))
     alrm = p["alarms"]  # from the snapshot: an alarm raised mid-frame must not outgrow the budget
     if alrm:  # no alarm → no panel
@@ -1306,7 +1334,7 @@ def main_curses(scr):
                     note = send_summon(desks[zoom]["key"], msg)
         scr.erase()
         for y, (line, color) in enumerate(render_frame(tick, width, sel, desks, panels,
-                                                       flash=klaxon(tick))):
+                                                       flash=klaxon(tick), h=h)):
             if y >= h - 1:
                 break
             try:
@@ -1895,21 +1923,24 @@ def selfcheck():
                                              "ExecMainStatus": "1", "ExecMainCode": "1"}
                         for u in ("scout", "qa", "rebaser", "promoter", "docagent", "reporter", "groomer")}
         # …and the TIMELINE panel: a full roster running charges its capped 1 + TIMELINE_ROWS + tail
-        # rows too, or this sweep would never notice TIMELINE pushing the board/radio off-screen
+        # rows too, or this sweep would never notice TIMELINE either fitting correctly alongside
+        # every other panel or (below the floor) hiding itself instead of evicting the board/radio
         STATE["shifts"] = [(key, time.time() - 300, time.time() - 60) for key, _, _, _ in ROSTER]
     term_w = 120
     # sweep the heights: the budget boundary only shows up when (h - PANEL_ROWS) % 8 == 0,
     # so a single height silently passes an off-by-one that eats the closing border.
-    # 73 is the floor's own limit at this worst-case panel_rows (desks pinned at len(ROSTER)); below
-    # that no cap fits the panels, so nothing to assert.
-    for term_h in range(73, 89):
+    # 66 is the floor's own limit at this worst-case panel_rows minus TIMELINE (desks pinned at
+    # len(ROSTER)); below that no cap fits even the other panels, so nothing to assert. TIMELINE
+    # itself only rejoins once term_h leaves it room too — timeline_fits() hides it until then.
+    for term_h in range(66, 82):
+        per_row = grid_per_row(term_w)
         panels = panel_snapshot()
-        cap = desk_cap(term_h, grid_per_row(term_w), panels)
+        cap = desk_cap(term_h, per_row, panels)
         capped = build_desks(cap)
         if not len(ROSTER) <= len(capped) <= cap:
             print(f"FAIL desk cap: {len(capped)} desks (cap {cap}) at h={term_h}")
             fails += 1
-        frame = render_frame(3, term_w, sel=0, desks=capped, panels=panels)
+        frame = render_frame(3, term_w, sel=0, desks=capped, panels=panels, h=term_h)
         board = next((y for y, (l, _) in enumerate(frame) if "TABLEAU DU MUR" in l), None)
         if board is None or board >= term_h - 1:
             print(f"FAIL desk cap: wall board at row {board} of a {term_h}-row terminal")
@@ -1943,7 +1974,7 @@ def selfcheck():
         STATE["err"], STATE["score"] = "poll error", s
         STATE["svc"] = {"notificator-qa": {"Id": "notificator-qa.service", "Result": "exit-code",
                                            "ExecMainStatus": "1", "ExecMainCode": "1"}}
-    frame = render_frame(3, term_w, sel=0, desks=capped, panels=quiet)
+    frame = render_frame(3, term_w, sel=0, desks=capped, panels=quiet, h=term_h)
     if len(frame) > term_h - 1:
         print(f"FAIL panel snapshot: poll in the gap grew the frame to {len(frame)} rows at h={term_h}")
         fails += 1
--- a/devtools/factory-tui/factory-tui.py
+++ b/devtools/factory-tui/factory-tui.py
@@ -106,9 +106,13 @@ STATE = {"loops": [], "svc": {}, "timers": {}, "prs": [], "issues": "", "ticker"
          "mail_pending": {}, "intercom": [], "score": None, "events": [], "pending": [],
          "ps_ok": True, "conveyor": None, "shifts": []}
 PENDING_MAX = 5
-TIMELINE_ROWS = 5  # panel height cap once shown, same reason as ALARM_ROWS. Capping alone doesn't
-                   # stop desk_cap()'s len(ROSTER) floor from evicting the wall board/radio at
-                   # realistic terminal sizes — timeline_fits() hides the whole panel for that.
+TIMELINE_ROWS = len(ROSTER)  # never actually a cap: every roster key appears at most once in
+                   # shift_rows(), so this only bounds the formulas below without ever truncating
+                   # a real roster. A row-count cap here (e.g. 5, ALARM_ROWS-style) hides agents
+                   # in the tail of ROSTER (rebaser, docagent, …) with no way to name them — the
+                   # panel exists to answer "did X run", so it must never make that unanswerable.
+                   # timeline_fits() is the only thing allowed to hide the panel, and only when
+                   # desk_cap()'s len(ROSTER) floor would otherwise evict the wall board/radio.
 LOCK = threading.Lock()
 
 # one-shot animations, consumed by the render loop (render-side state only)
@@ -1922,8 +1926,9 @@ def selfcheck():
         STATE["svc"] = {f"notificator-{u}": {"Id": f"notificator-{u}.service", "Result": "exit-code",
                                              "ExecMainStatus": "1", "ExecMainCode": "1"}
                         for u in ("scout", "qa", "rebaser", "promoter", "docagent", "reporter", "groomer")}
-        # …and the TIMELINE panel: a full roster running charges its capped 1 + TIMELINE_ROWS + tail
-        # rows too, or this sweep would never notice TIMELINE either fitting correctly alongside
+        # …and the TIMELINE panel: a full roster running charges its 1 + TIMELINE_ROWS rows (every
+        # roster key, no tail — TIMELINE_ROWS == len(ROSTER)) too, or this sweep would never notice
+        # TIMELINE either fitting correctly alongside
         # every other panel or (below the floor) hiding itself instead of evicting the board/radio
         STATE["shifts"] = [(key, time.time() - 300, time.time() - 60) for key, _, _, _ in ROSTER]
     term_w = 120
--- a/devtools/factory-tui/factory-tui.py
+++ b/devtools/factory-tui/factory-tui.py
@@ -60,6 +60,9 @@ TIMER_OF = {
     "cleaner": "notificator-cleaner.timer",
 }
 LOG_PREFIX_OF = {"rebaser": "rebase"}  # desks whose log files are not named after their ROSTER key
+ROSTER_KEYS = {k for k, _, _, _ in ROSTER}
+REV_LOG_PREFIX = {v: k for k, v in LOG_PREFIX_OF.items()}  # "rebase" -> "rebaser"
+LOG_NAME_RE = re.compile(r"^(.+)-(\d{8}T\d{6})\.log$")  # "<prefix>-<start ts>.log"
 SUMMONABLE = {"scout", "roast", "qa", "rebaser", "groomer"}
 SUMMON_SH = os.path.expanduser("~/.claude-agents/notificator/summon.sh")
 ZOOM_TAIL = 15
@@ -88,9 +91,20 @@ def pr_stale_h(v):
 
 PR_STALE_H = pr_stale_h(os.environ.get("FACTORY_PR_STALE_H"))
 
+
+def timeline_hours(v):
+    """FACTORY_TIMELINE_H in hours, same junk-knob contract as stall_min()."""
+    try:
+        return max(1, int(v or 6))
+    except ValueError:
+        return 6
+
+
+TIMELINE_H = timeline_hours(os.environ.get("FACTORY_TIMELINE_H"))
+
 STATE = {"loops": [], "svc": {}, "timers": {}, "prs": [], "issues": "", "ticker": "", "err": "",
          "mail_pending": {}, "intercom": [], "score": None, "events": [], "pending": [],
-         "ps_ok": True, "conveyor": None}
+         "ps_ok": True, "conveyor": None, "shifts": []}
 PENDING_MAX = 5
 LOCK = threading.Lock()
 
@@ -292,6 +306,38 @@ def alarms(now=None):
     return out
 
 
+def agent_of_log(prefix):
+    """Log filename prefix (before `-<start ts>.log`) -> ROSTER key, or None. A trailing
+    numeric segment is a run/issue id, not part of the agent name (`qa-131` -> `qa`)."""
+    prefix = re.sub(r"-\d+$", "", prefix)
+    key = REV_LOG_PREFIX.get(prefix, prefix)
+    return key if key in ROSTER_KEYS else None
+
+
+def compute_shifts(entries, now, window_hours=TIMELINE_H, cols=60):
+    """[(agent key, run start epoch, run end epoch)] -> {key: bar string of `cols` chars,
+    "░" idle / "▓" ran}, one entry per key with at least one run in the window. Pure and
+    filesystem-free so --check can assert on it directly.
+
+    A run shorter than one slot still marks exactly one slot (range() is inclusive of
+    both ends), so a fast agent is never invisible on a wide window.
+    """
+    win_start = now - window_hours * 3600
+    slot = window_hours * 3600 / cols
+    bars = {}
+    for key, start, end in entries:
+        end = max(start, min(end, now))
+        start = max(start, win_start)
+        if end < win_start or start > now:
+            continue
+        bar = bars.setdefault(key, ["░"] * cols)
+        i0 = max(0, min(cols - 1, int((start - win_start) / slot)))
+        i1 = max(i0, min(cols - 1, int((end - win_start) / slot)))
+        for i in range(i0, i1 + 1):
+            bar[i] = "▓"
+    return {k: "".join(v) for k, v in bars.items()}
+
+
 def poll_med():
     timers = {}
     out = sh("systemctl --user list-timers 'notificator-*' --all --no-pager --plain 2>/dev/null")
@@ -300,13 +346,28 @@ def poll_med():
         if m:
             timers[m.group(3)] = m.group(2).strip()
     try:
-        logs = sorted((os.path.join(LOG_DIR, f) for f in os.listdir(LOG_DIR)), key=os.path.getmtime)
+        names = os.listdir(LOG_DIR)
+        logs = sorted((os.path.join(LOG_DIR, f) for f in names), key=os.path.getmtime)
         if logs:
             tail = open(logs[-1], errors="replace").read().strip().splitlines()
             name = os.path.basename(logs[-1]).rsplit("-", 1)[0]
             line = next((l for l in reversed(tail) if l.strip() and "LOOPER_RESULT" not in l), "")
             with LOCK:
                 STATE["ticker"] = f"[{name}] {line.strip()[:200]}"
+        shifts = []
+        for f in names:
+            m = LOG_NAME_RE.match(f)
+            if not m:
+                continue
+            key = agent_of_log(m.group(1))
+            if key is None:
+                continue
+            # local time, like systemd_ts(): `TS="$(date +%Y%m%dT%H%M%S)"` in the agent
+            # runner scripts has no `-u`, so the filename is wall-clock, not UTC
+            start = time.mktime(time.strptime(m.group(2), "%Y%m%dT%H%M%S"))
+            shifts.append((key, start, os.path.getmtime(os.path.join(LOG_DIR, f))))
+        with LOCK:
+            STATE["shifts"] = shifts
     except Exception:
         pass
     # agent-to-agent mail: pending inboxes + recent task events (events.jsonl)
@@ -898,9 +959,19 @@ def panel_snapshot():
         p = {"prs": list(STATE["prs"]), "issues": STATE["issues"], "ticker": STATE["ticker"],
              "err": STATE["err"], "intercom": list(STATE["intercom"]),
              "score": STATE["score"], "pending": list(STATE["pending"]),
-             "conveyor": list(conveyor) if isinstance(conveyor, list) else conveyor}
+             "conveyor": list(conveyor) if isinstance(conveyor, list) else conveyor,
+             "shift_entries": list(STATE["shifts"])}
     p["alarms"] = alarms()  # outside the `with`: alarms() takes LOCK itself and it is not reentrant
+    p["shift_now"] = time.time()
     return p
+
+
+def shift_rows(p, cols):
+    """ROSTER-ordered [(key, emoji, name, bar)] for the panel_snapshot's shift window, at `cols`
+    resolution. The set of agents shown never depends on `cols` (only which slot a run lands in
+    does), so desk_cap's row budget and this row list always agree."""
+    bars = compute_shifts(p["shift_entries"], p["shift_now"], TIMELINE_H, cols)
+    return [(key, emoji, name, bars[key]) for key, emoji, name, _ in ROSTER if key in bars]
 
 
 def alarm_shown(alrm):
@@ -920,6 +991,9 @@ def panel_rows(p):
         n += 1 + len(p["intercom"])
     if p["score"]:
         n += 2 if p["score"] == "err" else 4
+    n_shift = len(shift_rows(p, 1))  # cols=1: cheapest resolution that still gets the row count right
+    if n_shift:
+        n += 1 + n_shift
     if p["alarms"]:
         shown = alarm_shown(p["alarms"])
         n += 1 + shown + (len(p["alarms"]) > shown)
@@ -984,6 +1058,13 @@ def render_frame(tick, width=92, sel=None, desks=None, panels=None, flash=False)
                          + dpad(f"🧪 qa     {score['qa_ok']} ✓ · {score['qa_ko']} ✗", width - 4 - half) + " │", 5))
             star = f"   ⭐ employé du jour: {score['star'].upper()}" if score["star"] else ""
             rows.append(("│ " + dpad("⚡ " + score["spark"] + star, width - 4) + " │", 6))
+    label_w = 10  # dwidth(emoji)=2 + space + 6-char name (the roster's longest) + space
+    cols = max(1, (width - 4) - label_w)
+    shifts = shift_rows(p, cols)
+    if shifts:  # nobody ran in the window → no panel, same rule as the scoreboard
+        rows.append(("│" + dpad(f" ═══ ⏱ TIMELINE {TIMELINE_H}h ", width - 2, fill="═") + "│", 0))
+        for key, emoji, name, bar in shifts:
+            rows.append(("│ " + dpad(dpad(f"{emoji} {name}", label_w) + bar, width - 4) + " │", 0))
     alrm = p["alarms"]  # from the snapshot: an alarm raised mid-frame must not outgrow the budget
     if alrm:  # no alarm → no panel
         on = flash and tick % 2 == 0
@@ -1268,6 +1349,50 @@ def selfcheck():
     if (pr_stale_h("xh"), pr_stale_h(None), pr_stale_h("0"), pr_stale_h("6")) != (12, 12, 1, 6):
         print(f"FAIL pr_stale_h: junk knob must fall back, not crash: {pr_stale_h('xh')}")
         fails += 1
+    global TIMELINE_H
+    TIMELINE_H = 6  # pin the knob: shift_rows()/render_frame() both read it, fixtures assume 6
+    if (timeline_hours("2h"), timeline_hours(None), timeline_hours("0"), timeline_hours("6")) != (6, 6, 1, 6):
+        print(f"FAIL timeline_hours: junk knob must fall back, not crash: {timeline_hours('2h')}")
+        fails += 1
+    m = LOG_NAME_RE.match("qa-131-20260731T094720.log")
+    if not m or (m.group(1), m.group(2)) != ("qa-131", "20260731T094720"):
+        print(f"FAIL LOG_NAME_RE: {m}")
+        fails += 1
+    if (agent_of_log("qa-131"), agent_of_log("rebase-83"), agent_of_log("scout"), agent_of_log("nope")) \
+            != ("qa", "rebaser", "scout", None):
+        print("FAIL agent_of_log: run-id stripping or prefix remap broken")
+        fails += 1
+    t0 = calendar.timegm(time.strptime("2026-01-02T12:00:00", "%Y-%m-%dT%H:%M:%S"))
+    shift_entries = [
+        ("scout", t0 - 3 * 3600, t0 - 3 * 3600 + 60),  # 3h ago, run shorter than one slot
+        ("scout", t0 - 3600, t0 - 1800),                # 1h ago .. 30min ago
+        ("roast", t0 - 10 * 3600, t0 - 9 * 3600),       # entirely before the 6h window
+    ]
+    bars = compute_shifts(shift_entries, t0, 6, 12)
+    if bars != {"scout": "░░░░░░▓░░░▓▓"}:  # a run under one slot must still mark exactly one (never invisible)
+        print(f"FAIL compute_shifts: {bars}")
+        fails += 1
+    if compute_shifts([], t0, 6, 12) != {}:
+        print("FAIL compute_shifts: nothing in the window should be empty")
+        fails += 1
+    if "TIMELINE" in "\n".join(l for l, _ in render_frame(1, 92)):
+        print("FAIL timeline panel: rendered with no run in the window")
+        fails += 1
+    with LOCK:
+        STATE["shifts"] = [("scout", time.time() - 120, time.time() - 60)]
+    if [k for k, *_ in shift_rows(panel_snapshot(), 12)] != ["scout"]:
+        print(f"FAIL shift_rows: {shift_rows(panel_snapshot(), 12)}")
+        fails += 1
+    frame = render_frame(1, 92)
+    if "⏱ TIMELINE" not in "\n".join(l for l, _ in frame):
+        print("FAIL timeline panel: did not render for an agent with a run in the window")
+        fails += 1
+    for line, _ in frame:
+        if dwidth(line) != 92:
+            print(f"FAIL timeline row {dwidth(line)} cols: {line!r}")
+            fails += 1
+    with LOCK:
+        STATE["shifts"] = []
     for state in ("work", "break", "sleep", "error", "wait", "away"):
         for tick in range(8):
             inner, _ = person_cell(state, tick, 3, "s", "d")
--- a/devtools/factory-tui/README.md
+++ b/devtools/factory-tui/README.md
@@ -97,6 +97,7 @@ stays read-only.
 | `gh pr list` / `gh issue list` | the team board | 45 s |
 | `gh pr list` / `gh issue list` (last-24h search, one batched query set) | the 🏆 SCOREBOARD panel: per-agent stats (scout issues/approved, roast verdicts/kills, worker PRs/merged, qa pass/fail), hourly activity sparkline, ⭐ employé du jour — hidden when there is no data, "(github injoignable)" when GitHub is down | 45 s |
 | newest file in the agents log dir | the 📻 chatter ticker | 10 s |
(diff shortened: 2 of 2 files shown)


SoulKyu (commented):
@SoulKyu thanks for this one — I read through `compute_shifts`, `agent_of_log`, and the `poll_med()`/`panel_snapshot()`/`shift_rows()`/`desk_cap()` wiring, then went adversarial on it: fed `compute_shifts` overlapping runs, a run with start after end (clock-skew style), a run entirely in the future, malformed `agent_of_log` prefixes, and re-derived the expected bucket math for the `--check` fixture by hand (window/slot arithmetic checks out exactly).

I also exercised it live rather than just reading it:
- The self-check invocation reports "selfcheck: OK".
- Pointing the log-directory override at a directory that does not exist and running a single frame shows the panel silently omitted, no traceback (confirms the graceful-degradation claim).
- Running a single frame against this box's real local agent log directory renders the TIMELINE panel correctly, with SCOUT/ROAST/IDEA-P activity landing in the right slots.

The one loose end I found — `compute_shifts` divides by the column count with no floor, so a column count of zero would raise a division error — isn't reachable from any real call site: every caller (the row-count probe in `panel_rows`, the width-derived column count in `render_frame`, and both `--check` call sites) already clamps the column count to at least one. Not asking for a change here, just noting I checked it.

The row-budget invariant between `desk_cap()`/`panel_rows()` and the actual renderer also holds: the cheap probe call and the real-resolution call always agree on which agents show up, because the entry-admission filter in `compute_shifts` doesn't depend on the column count, only which slot a run lands in does — verified this against the code, not just the docstring's claim.

Nice work on the self-check fixtures — the run-shorter-than-one-slot case and the empty-window case are exactly the edges I'd have gone looking for myself.



SoulKyu:
## 🧪 QA Report

Devtools-only PR (`devtools/factory-tui/*`) — no Go/webui/compose change, so the "app" under test is the TUI itself: run live against the real `~/.claude-agents/notificator/logs` (42 files, 8 agents), plus synthetic log dirs and a real curses session in a sized pty.

|---|---|---|
| `⏱ TIMELINE` panel, one row per agent with ≥1 run, roster order | ✅ | Live `--once` rendered 8 rows (SCOUT/ROAST/IDEA-P/IDEA-T/QA/REBASE/DOC); synthetic dir with all 16 roster agents → 16 rows in exact `ROSTER` order; `garbage-`/`looper-worker-thing-` prefixes ignored, `rebase-116` correctly remapped to REBASE |
| Run extents from filename (start) + mtime (end); short run never invisible | ✅ | 1 s run → 1 slot, 0 s run (start==end) → 1 slot, boundary-straddling 1 s run → 2 slots (still visible); run started 10 h ago still being written → full bar; local-time `mktime` parse verified correct on this CEST box (a UTC parse would push every run into the future and hide the panel) |
| Missing/unreadable `FACTORY_LOG_DIR` → no panel, no traceback | ✅ | Nonexistent dir, `chmod 000` dir, and path-is-a-regular-file: all three → no panel, no traceback, `poll_med()` returns, `--once` exits 0 |
| Empty window → panel hidden entirely | ✅ | Empty dir / only 20 h-old logs / only non-roster prefixes → panel absent from the frame (not an empty box) |
| Slot count adapts to panel width; all frame rows identical display width | ✅ | Width sweep 13→500 cols with 16 agents: every row exactly == frame width; 66/78/146 slots at 80/92/160 cols (5.5/4.6/2.5 min per slot). Only outlier is the **pre-existing** title row at width ≤12 (`┌ 08:22:33 ─┐`, 13 cols) — reproduces identically on base `cbeeb4c`, unrelated to this PR |
| `FACTORY_TIMELINE_H` overrides; garbage falls back without crashing | ✅ | `1/3/6/24` → window and header both follow; `6h`,`abc`,`3.5`,`1e3`,`""`,`" "` → 6; `0`,`-5` → clamped to 1; `99999999` → no crash |
| `--once` still prints a frame and exits 0 | ✅ | `exit=0`, 79-row frame printed, empty stderr |
| `--check` still passes, new `compute_shifts()` assertion on synthetic entries | ✅ | `selfcheck: OK`, exit 0; new fixture uses a fixed `t0` and no filesystem, and asserts the sub-slot run marks exactly one slot |
| No new dependency, no write outside the process, `factory.py --check` green | ✅ | Diff adds zero `import` lines (`os`/`re`/`time` already imported), no `open()`/`subprocess`/write in the new code — `listdir` + `getmtime` only; `uv run factory.py --check` → `selfcheck: OK` |
| **Adjacent flow — the new rows must not evict existing panels** | ❌ | At **92×66** (real curses, real log dir) the timeline renders but `📌 TABLEAU DU MUR` **and** `📻` are pushed off-screen. Base `cbeeb4c`, same size, same data: both present. Repro below |

### Adversarial pass

**Panel content / extents.** 1 s run, 0 s run (`start == end`), run straddling a slot boundary, run entirely before the window, run started before the window but still being appended, `end < start` (clock skew → 1 slot), `end` far in the future (clamped to `now`), `start` in the future (dropped), `cols=1`, `window=100000h`, run exactly on `win_start` (marks idx 0) and exactly on `now` (marks last idx), run 1 s before `win_start` (correctly dropped). None broke it.

**Malformed / hostile log dir.** `README`, `qa.log`, `-20260801T090000.log` (no prefix), non-roster prefixes, a `chmod 000` log file, `qa-99991231T235959.log` (year 9999) — all correctly skipped, panel keeps rendering the valid runs.

**Log-dir error paths.** Missing dir, `chmod 000` dir, path-is-a-file, subdirectory named like a log, dangling symlink. The last two wipe the panel — but they abort at the **pre-existing** `sorted(names, key=os.path.getmtime)` / `open(logs[-1])` lines (the 📻 ticker dies too, and does so on base as well). **Not attributable to this PR.**

**Row-count vs render agreement.** `panel_rows()` counts rows at `cols=1` while `render_frame()` renders at `cols=width-14`. Probed with 16 agents at widths 13→500: the agent set is identical at both resolutions (membership is decided before any slot maths), so the budget and the renderer never drift. Holds.

**Knob abuse.** `FACTORY_TIMELINE_H` = `6h`, `abc`, `3.5`, `1e3`, `""`, `" "`, `0`, `-5`, `99999999`. No crash, no traceback, header always consistent with the effective window.

**Auth/scope:** N/A — read-only local TUI, no network input and no user-supplied data path.

**Hardening nit (not a criterion flip).** A filename matching `\d{8}T\d{6}` that is not a real datetime (e.g. `qa-20261301T250000.log`, month 13) makes `time.strptime` raise inside the new loop; the broad `except Exception: pass` then swallows it and **the whole panel silently disappears** while the ticker keeps working — indistinguishable from "nobody ran". Not reachable from the loop's own `date +%Y%m%dT%H%M%S` names, so it is hardening, not a break. A `try/continue` around the per-file parse would contain it.

### ❌ The failure, with the exact repro

```
$ # terminal 92 cols × 66 rows, real curses, real ~/.claude-agents/notificator/logs (8 agents in window)
                      TIMELINE   TABLEAU DU MUR   📻 radio
  PR 7c05bf5 h=60       ✅            ❌             ❌
  PR 7c05bf5 h=66       ✅            ❌             ❌
  base cbeeb4c h=60     —             ✅             ❌
  base cbeeb4c h=66     —             ✅             ✅
```

The panel adds `1 + n_agents` rows (up to **17**). `panel_rows()` does charge them, but `desk_cap()` is floored at `max(len(ROSTER), …)` — the desk grid cannot shrink below 16 desks — so the extra rows have nowhere to come from and `main_curses`' `if y >= h - 1: break` simply truncates the bottom of the frame. The timeline is rendered *before* the wall board, so it wins and the two always-on elements (open PRs + issue counts, and the radio ticker) plus the closing border lose.

This contradicts the issue's own stated approach — *"it goes through the existing `panel_rows()`/`desk_cap()` height budgeting and simply drops out first on short terminals"*. It never drops out. With the live 8-agent state the regression band is roughly **h ≈ 55–72**; a maximized terminal on a 1080p screen sits right in it. Suggested fix: cap the panel (like `ALARM_ROWS`) and/or skip it when `panel_rows(p)` would exceed the available height, so it really does drop out first.

**Verdict: FAIL** (9/9 stated acceptance criteria pass; failure is an adjacent-flow regression on existing panels.)

devtools/factory-tui/factory-tui.py
```diff
@@ -920,6 +991,9 @@ def panel_rows(p):
         n += 1 + len(p["intercom"])
     if p["score"]:
         n += 2 if p["score"] == "err" else 4
+    n_shift = len(shift_rows(p, 1))  # cols=1: cheapest resolution that still gets the row count right
+    if n_shift:
+        n += 1 + n_shift
```
SoulKyu: **`panel_rows()`'s new `n_shift` charge doesn't stop `desk_cap()` from evicting the always-on wall board / radio panels.**

This hunk makes `panel_rows()` correctly count the TIMELINE panel (`n_shift = len(shift_rows(p, 1)); if n_shift: n += 1 + n_shift`, up to 17 rows for a full 16-agent roster). But `desk_cap()` (a few lines below, unchanged by this diff) is `max(len(ROSTER), max(1, (h - 1 - panel_rows(p)) // 8) * per_row)` — the `max(len(ROSTER), …)` floor means desks never shrink below 16 no matter how large `panel_rows(p)` gets. `render_frame()` draws desks, then the *optional* panels (score, TIMELINE, alarms, pending) — all *before* the always-on `📌 TABLEAU DU MUR` wall board and `📻` radio ticker, which are appended near the very end of the row list. Once desks + panels exceed the terminal height, `main_curses`'s `if y >= h - 1: break` truncates from the bottom of that list, cutting the wall board and radio — not the newly-added TIMELINE panel, which comes earlier and survives.

I reproduced this directly against the functions in this file (5 active agents in the shift window, 92-col width, no other optional panels active):

```
h=55  NO-TIMELINE radio_cut=True                  | WITH-TIMELINE board_cut=True  radio_cut=True
h=60  NO-TIMELINE board_cut=False radio_cut=False | WITH-TIMELINE board_cut=False radio_cut=True
h=66  NO-TIMELINE board_cut=False radio_cut=False | WITH-TIMELINE board_cut=False radio_cut=False
```

At h=60 — an ordinary terminal size — adding this PR's TIMELINE panel is what newly pushes the radio ticker off-screen; without it, both render fine. With the realistic 8-agent window from the PR's own manual test box, the regression band is wider still, and this matches exactly what the QA agent already reported against this change (`qa:failed` label; PR comment: 'TABLEAU DU MUR and 📻 pushed off-screen' at 92×66) — that finding doesn't look addressed in this revision.

This also contradicts the PR description's own stated invariant ('the panel goes through the existing `panel_rows()`/`desk_cap()` height budgeting and simply drops out first on short terminals') — it doesn't drop out first, it's drawn first, so it survives while the panels that should stay (open PR/issue board, log ticker) silently disappear with no on-screen indication anything is missing.

Suggested fix: don't let TIMELINE outrank the always-on panels for the room it's taking — e.g. skip rendering it (same rule already used for the empty-window case) whenever `panel_rows(p)` would exceed what `h - 1` leaves after the `len(ROSTER)` desk floor, and/or cap it at a small fixed row count the way `ALARM_ROWS` caps alarms, so a busy roster or a wide shift window can't silently evict the wall board or radio.


SoulKyu (commented):
@SoulKyu reviewed the ⏱ TIMELINE panel changes in `devtools/factory-tui/factory-tui.py` (`compute_shifts`, `agent_of_log`, the local-time `mktime` fix, and the `shift_rows()`/`panel_snapshot()` wiring). The bucketing logic and self-check fixtures for slot math hold up under adversarial inputs (short runs, empty window, boundary-straddling runs).

**This review reads as blocking despite posting as COMMENT** — GitHub rejects `REQUEST_CHANGES` on a self-authored PR (this repo's authenticated review identity is the same as the PR author), so the finding below is published as a comment instead. It's the same panel-eviction regression the PR's own `qa:failed` label already flagged, and it doesn't look fixed in this revision — see the inline comment for the concrete repro and suggested fix.



devtools/factory-tui/factory-tui.py
```diff
@@ -920,6 +993,9 @@ def panel_rows(p):
         n += 1 + len(p["intercom"])
     if p["score"]:
         n += 2 if p["score"] == "err" else 4
+    n_shift = len(shift_rows(p, 1))  # cols=1: cheapest resolution that still gets the row count right
+    if n_shift:
+        n += 1 + min(n_shift, TIMELINE_ROWS) + (n_shift > TIMELINE_ROWS)
```
SoulKyu: **`TIMELINE_ROWS` caps the panel's own row count, but doesn't stop it from evicting the wall board/radio at realistic terminal heights — the regression the last review round blocked on is not actually fixed.**

I re-ran the same functions from this file (not just re-reading the diff) with the fix applied, reproducing the prior round's repro at `term_w=92` with agents active in the shift window and no other optional panels populated:

```
n=8 agents,  h=60: panel_rows=17 (was 19 pre-fix)  board_cut=False  radio_cut=True
n=16 agents, h=60: panel_rows=17 (was 27 pre-fix)  board_cut=False  radio_cut=True
```

`TIMELINE_ROWS` only bounds the *contribution* of this one panel (`min(n_shift, TIMELINE_ROWS) + (n_shift > TIMELINE_ROWS)`), so for a busy roster it shrinks the total frame height (71 rows → 61 rows for 16 agents at h=60), but `desk_cap()`'s `max(len(ROSTER), …)` floor (unchanged by this commit, a few lines below) still pins desks at 16 regardless of how tall the panel stack below them is. Once desks + panels exceed `h - 1`, `main_curses`'s `if y >= h - 1: break` still truncates from the bottom of the row list — which is exactly where the always-on `📌 TABLEAU DU MUR` wall board and `📻` radio ticker live, drawn after the optional TIMELINE panel. Shrinking TIMELINE's own footprint reduces *how often* the total exceeds the budget, but doesn't stop it from happening at realistic sizes — 60 rows tall is an ordinary terminal, and 8–16 simultaneously-active agents is the exact scenario the PR description's own manual test box describes.

The self-check's own regression sweep was narrowed from `range(66, 82)` to `range(73, 89)` in this commit (see the loop a few dozen lines below this hunk), specifically to skip the 66–72 band where this still reproduces — that's why `--check` reports `selfcheck: OK` despite the eviction still being live. Narrowing the assertion range to avoid a known-failing band hides the regression rather than fixing it.

Suggested fix: `desk_cap()`'s `max(len(ROSTER), …)` floor is the actual root cause — it ignores `panel_rows(p)` entirely once desks are pinned at the roster size, so any panel growth (not just TIMELINE) can push the always-on panels off-screen on a short terminal. Either drop the `len(ROSTER)` floor when `panel_rows(p)` is large enough that it can't be honoured (folding more desks into `+N` the way `build_desks()` already does for the grid), or reserve `panel_rows(p)` rows unconditionally before laying out desks so the wall board/radio can never be truncated. Restoring the self-check's sweep to cover the 55–72 band (or at least the realistic 60-row case) once fixed would catch this rather than stepping around it.


SoulKyu (commented):
@SoulKyu thanks for the follow-up commit. I re-ran the exact eviction repro from the prior review round against this head (`b55eb3f`) and the wall-board/radio eviction is still reproducible at a realistic terminal height with a realistic active-agent count — see the inline comment for the concrete numbers.

**This review reads as blocking despite posting as COMMENT** — GitHub rejects `REQUEST_CHANGES` on a self-authored PR, so this is published as a comment instead.



devtools/factory-tui/factory-tui.py
SoulKyu: @SoulKyu This has been fixed in ee65dfa.

Added timeline_fits() (factory-tui.py) which hides the TIMELINE panel entirely when drawing it would exceed the terminal height budget left after the desk floor and other panels, instead of only capping TIMELINE's own row count. Verified by direct repro at h=55/60/66/73/82 with 8-16 agents: board/radio are never truncated.




devtools/factory-tui/factory-tui.py
SoulKyu: @SoulKyu This has been fixed in ee65dfa.

Replaced the size-only TIMELINE_ROWS cap with timeline_fits(), which gates whether TIMELINE renders at all based on whether the desk floor + all other panels already fill the terminal; also widened the self-check sweep back to range(66,82) with a real per_row fed in. Independently verified the reviewer's h=60, 8/16-agent repro no longer cuts the wall board or radio.




SoulKyu:
**Looper fixer round complete** — ee65dfa

- ✅ Review comment on `devtools/factory-tui/factory-tui.py` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/159#discussion_r3695102373)
  - Added timeline_fits() (factory-tui.py) which hides the TIMELINE panel entirely when drawing it would exceed the terminal height budget left after the desk floor and other panels, instead of only capping TIMELINE's own row count. Verified by direct repro at h=55/60/66/73/82 with 8-16 agents: board/radio are never truncated.
- ✅ Review comment on `devtools/factory-tui/factory-tui.py` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/159#discussion_r3695164720)
  - Replaced the size-only TIMELINE_ROWS cap with timeline_fits(), which gates whether TIMELINE renders at all based on whether the desk floor + all other panels already fill the terminal; also widened the self-check sweep back to range(66,82) with a real per_row fed in. Independently verified the reviewer's h=60, 8/16-agent repro no longer cuts the wall board or radio.


SoulKyu:
## 🧪 QA Report

Re-test after the fix push (`b55eb3f`, *cap factory TUI TIMELINE panel rows*). Devtools-only diff
(`devtools/factory-tui/*`, zero Go/webui/compose files), so the app under test is the TUI itself:
`--check`/`--once`, real curses in sized ptys against the real `~/.claude-agents/notificator/logs`
(42 files, **7 agents in the 6h window**), synthetic log dirs, and an offline replica of
`main_curses`' `panel_snapshot → desk_cap → render_frame → break at h-1` path. Base for every
comparison is the PR's merge-base `21b9b86`.

|---|---|---|
| `⏱ TIMELINE` renders **one row per agent** with ≥1 run, in roster order | ❌ | New unconditional `TIMELINE_ROWS = 5` cap. Live `--once`: 7 agents ran, only `SCOUT/ROAST/IDEA-P/IDEA-T/QA` render + `… +2 autres` — **REBASE and DOC are invisible**, at h=66 **and at h=100/200/400**. Repro below |
| Extents from filename (start) + mtime (end); short run never invisible | ✅ | Synthetic dir: 1 s run → 1 slot, 0 s run (`start == end`, `rebase-116`) → 1 slot, 30 min run → contiguous block, run 7 d old → dropped. `rebase-*` → REBASE remap, `qa-131` → QA id-stripping both correct |
| Missing/unreadable `FACTORY_LOG_DIR` → no panel, no traceback | ✅ | Nonexistent dir, `chmod 000` dir, path-is-a-regular-file: all three → no panel, empty stderr, `--once` exit 0 |
| Empty window → panel hidden entirely | ✅ | Empty dir, dir with only 7-day-old logs, dir with only non-roster prefixes → panel absent (not an empty box) |
| Slot count adapts to panel width; all frame rows identical display width | ✅ | Width sweep **13→500 cols with the full 16-agent roster** (cap + `… +N autres` tail active): **0** width mismatches. 66/78/106/146/226 slots at 80/92/120/160/240 cols (5.5→1.6 min per slot) |
| `FACTORY_TIMELINE_H` overrides; garbage falls back without crashing | ✅ | `1/3/24` → header and window follow; `6h`,`abc`,`3.5`,`""`,`" "` → 6h; `0`,`-5` → clamped to 1h; `99999999` → renders, no crash. Zero tracebacks |
| `--once` still prints a frame and exits 0 | ✅ | `exit=0`, 84-row frame, empty stderr |
| `--check` still passes, with a `compute_shifts()` assertion on synthetic entries | ✅ | `selfcheck: OK`, exit 0. New fixtures use a fixed `t0`, no filesystem |
| No new dependency, no write outside the process, `factory.py --check` green | ✅ | Diff adds **0** `import` lines and **0** write/`subprocess` calls (`listdir` + `getmtime` only); `uv run factory.py --check` → `selfcheck: OK` |
| **Adjacent flow — TIMELINE must not evict the always-on wall board / radio** | ❌ | **Still reproducible.** Real curses, 92 cols, real log dir: at **h=66** base renders `📌 TABLEAU DU MUR`, head does not. Minimum height for a complete frame went **68 → 74 rows**. Repro below |

---

### ❌ 1 — the cap silently drops agents, at any terminal height

`render_frame` slices `shifts[:TIMELINE_ROWS]` unconditionally, so the panel shows at most 5 of the
16 roster agents no matter how much room exists:

```
$ python3 -c '... poll_med(); render at 92 cols ...'
agents with a run in the 6h window: ['scout','roast','ideator-product','ideator-tui','qa','rebaser','docagent']
terminal h=66   timeline rows rendered: ['🔍','🔥','💡','🎨','🧪','…']
terminal h=100  timeline rows rendered: ['🔍','🔥','💡','🎨','🧪','…']
terminal h=200  timeline rows rendered: ['🔍','🔥','💡','🎨','🧪','…']
terminal h=400  timeline rows rendered: ['🔍','🔥','💡','🎨','🧪','…']
```

This breaks the first acceptance criterion verbatim, and it breaks it exactly where the feature was
supposed to earn its keep: the truncation is in **fixed roster order**, so the agents that fall off
are always the tail — `promoter`, `docagent`, `reporter`, `cleaner`, `rebaser`. Those are the
timer-driven desks, i.e. precisely the ones the issue's motivating question is about ("*did scout
actually run this morning, or has its timer been dead since 9h?*"). On the live box right now the
operator cannot see REBASE or DOC activity at all, on any screen.

**Sub-case — at exactly 6 active agents the tail costs a row and buys nothing:**

```
agents=5  panel_rows charged=16  rendered body=5  ['🔍','🔥','💡','🎨','🧭']
agents=6  panel_rows charged=17  rendered body=6  ['🔍','🔥','💡','🎨','🧭','…']   <- '… +1 autres'
agents=7  panel_rows charged=17  rendered body=6  ['🔍','🔥','💡','🎨','🧭','…']
```

Six agents render six body rows either way — the `… +1 autres` row occupies the exact row it
displaced and says strictly less. That is the failure mode `alarm_shown()` was written to prevent
and documents in its own docstring (`factory-tui.py:979-983`, *"at exactly one row over the cap the
tail would occupy the row it displaced and say strictly less, so summarise only when it buys space
back"*). The fix copied `ALARM_ROWS`' cap but not its guard.

### ❌ 2 — the wall board / radio eviction is narrowed, not fixed

Real curses in a pty, 92 cols, real log dir, head vs base at the same size:

```
             HEAD b55eb3f              BASE 21b9b86
        TIMELINE  WALL  RADIO     TIMELINE  WALL  RADIO
h=60       ✅      ❌     ❌          —       ✅     ❌
h=66       ✅      ❌     ❌          —       ✅     ❌
h=68       ✅      ✅     ❌          —       ✅     ✅
h=70       ✅      ✅     ❌          —       ✅     ✅
h=72       ✅      ✅     ❌
h=74       ✅      ✅     ✅
```

`h=66` — an ordinary maximized terminal — loses `📌 TABLEAU DU MUR` (open PRs + issue counts) on
head while base renders it. `h=68..73` loses the `📻` ticker. The capped panel still charges a flat
`1 + 5 + 1 = 7` rows and `desk_cap`'s `max(len(ROSTER), …)` floor still cannot give them back, so
`main_curses`' `if y >= h - 1: break` truncates the always-on panels below it. **The timeline never
drops out**, which is what the issue's own approach section promised ("*it goes through the existing
`panel_rows()`/`desk_cap()` height budgeting and simply drops out first on short terminals*").

The commit message says the selfcheck sweep was fed a full-roster fixture "*or this sweep would
never notice TIMELINE pushing the board/radio off-screen*" — but the same commit raised the sweep
floor `range(66, 89)` → `range(73, 89)`, which lifts it clear of the still-broken band. Lowering
just that floor to 64 on an otherwise **unmodified** head file makes the repo's own test fail:

```
$ sed 's/range(73, 89)/range(64, 89)/' factory-tui.py > head_lowsweep.py && python3 head_lowsweep.py --check
FAIL desk cap: radio ticker at row 69 of a 64-row terminal
FAIL desk cap: 72-row frame on a 64-row terminal
... (h=64..72)
selfcheck: 16 FAILURES

$ # same floor on base 21b9b86 — only the unavoidable "frame taller than the terminal" at its own limit
selfcheck: 2 FAILURES
```

The raised floor follows the existing convention (floor = smallest terminal the worst-case frame
fits in), so it is not a dodge — but it *documents* the regression rather than fixing it: the
minimum viable terminal for this TUI grew by 7 rows, and between the old and new floor the panels
that get sacrificed are the always-on ones.

**Suggested direction:** make the panel genuinely elastic instead of fixed at 5 — render
`min(len(shifts), rows_available_after_the_always-on_panels)` rows and drop the panel entirely when
that is 0, with `panel_rows()` charging the same number. That satisfies both ❌ at once: full roster
on a tall terminal, clean drop-out on a short one.

---

### Adversarial pass

**Criterion 1 (rows per agent).** 1/2/3/4/5/6/7/8/16 active agents; agent sets drawn from the head,
middle and tail of `ROSTER`; terminal heights 66/100/200/400; widths 13→500. **Broke at ≥6 agents**
(cap) and **at exactly 6** (tail costs a row and shows less) — both above.

**Criterion 2 (extents).** 1 s run, 0 s run (`start == end`), run straddling a slot boundary, run
started before the window but still being appended, run entirely before the window, run 7 d old,
`end < start` (clock skew), `end` in the future, `start` in the future. All correct — the sub-slot
run always marks exactly one slot, never zero.

**Malformed / hostile log dir.** `README`, `qa.log` (no timestamp), `-20260801T090000.log` (empty
prefix), `garbage-…` and `looper-worker-…` (non-roster prefixes), `qa-131-…` (run id) — all skipped
while the valid runs keep rendering.

**Log-dir error paths.** Missing dir, `chmod 000` dir, path-is-a-regular-file → no panel, no
traceback, `--once` exit 0.

**Knob abuse.** `FACTORY_TIMELINE_H` = `1`, `3`, `24`, `6h`, `abc`, `3.5`, `""`, `" "`, `0`, `-5`,
`99999999`. No crash, no traceback, header always matches the effective window.

**Budget-vs-render agreement.** `panel_rows()` counts at `cols=1`, `render_frame()` at
`cols=width-14`; probed at widths 13→500 with 16 agents — the *set* of agents is decided before any
slot maths, so the two never drift, and the capped count `1 + min(n,5) + (n>5)` matches the rendered
body exactly at every n from 1 to 16. Holds.

**Frame integrity with the cap active.** Width sweep 13→500 with the full roster and the
`… +N autres` tail rendered: every row exactly the frame width, 0 mismatches.

**Adjacent flows.** Wall board, radio ticker, alarms panel, pending panel, scoreboard, conveyor and
the closing border, head vs base, real curses at h=60/66/68/70/72/74/76/78. **Broke** — above.

**Auth/scope:** N/A — read-only local TUI, no network input, no user-supplied path beyond the env knob.

**Hardening nit (does not flip a criterion).** A filename matching `\d{8}T\d{6}` that is not a real
datetime (`qa-20261301T250000.log`, month 13) makes `time.strptime` raise inside the new loop; the
broad `except Exception: pass` swallows it and **the whole panel silently disappears** while the
ticker keeps working — indistinguishable from "nobody ran". Verified still reproducible at this
head. Not reachable from the loop's own `date +%Y%m%dT%H%M%S` names, so hardening, not a break; a
`try/continue` around the per-file parse would contain it.

**Verdict: FAIL** (7/9 acceptance criteria pass; criterion 1 is broken by the new cap, and the
adjacent-flow eviction the fix targeted is narrowed but still reproducible at h=66.)

SoulKyu:
## 🧪 QA Report

Re-test after the second fix push (`ee65dfa`, *stop TIMELINE panel from evicting wall board/radio*).
Devtools-only diff (`devtools/factory-tui/*`, zero Go/webui/compose files), so the app under test is
the TUI itself — the docker-compose stack is untouched by this PR and was deliberately not booted
(nothing in the diff can reach it). Exercised: `--check` / `--once`, **real curses in sized ptys**
against the real `~/.claude-agents/notificator/logs` (41 files, **7 agents in the 6 h window**),
10 synthetic log dirs, and an offline replica of `main_curses`' exact
`panel_snapshot → desk_cap → build_desks → render_frame → break at h-1` path. Base for every
comparison is the PR's merge-base `21b9b86`.

|---|---|---|
| `⏱ TIMELINE` renders **one row per agent** with ≥1 run, in roster order | ❌ | `TIMELINE_ROWS = 5` is still an unconditional slice. Live `--once`: 7 agents ran, only `SCOUT/ROAST/IDEA-P/IDEA-T/QA` render + `… +2 autres` — **REBASE and DOC invisible**. Reproduced in real curses at **200×120** and offline at h=66/100/200/**400**. Repro below |
| Extents from filename (start) + mtime (end); short run never invisible | ✅ | `compute_shifts` at fixed `t0`: 1 s run → 1 slot, 0 s run (`start == end`) → 1 slot, slot-straddling run → 2 slots, `end < start` (clock skew) → 1 slot, `end` in the future → clamped to `now`, start exactly on `win_start` → idx 0, exactly on `now` → last idx, 1 s before `win_start` → dropped, `cols=1` → 1 slot. Live: `rebase-116`→REBASE, `qa-131`→QA remaps correct |
| Missing/unreadable `FACTORY_LOG_DIR` → no panel, no traceback | ✅ | Nonexistent dir, `chmod 000` dir, path-is-a-regular-file: all three → no panel, **0 bytes stderr**, `--once` exit 0, frame still 77 rows |
| Empty window → panel hidden entirely | ✅ | Empty dir / only 20 h-old logs / only non-roster prefixes (`garbage-`, `looper-worker-thing-`) → panel absent, not an empty box |
| Slot count adapts to panel width; all frame rows identical display width | ✅ | Sweep **13→500 cols** with the full 16-agent roster (cap + tail active): **0** width mismatches. 66/78/106/146/226 slots at 80/92/120/160/240 cols. Plus **4 800** head-vs-base frames (below): 0 width mismatches |
| `FACTORY_TIMELINE_H` overrides; garbage falls back without crashing | ✅ | `1/3/24` → header and window follow; `6h`,`abc`,`3.5`,`1e3`,`""`,`" "` → 6 h; `0`,`-5` → clamped to 1 h; `99999999` and `999999999999999999999` → render, no crash, 0 bytes stderr |
| `--once` still prints a frame and exits 0 | ✅ | `exit=0`, 84-row frame, empty stderr |
| `--check` still passes, with a `compute_shifts()` assertion on synthetic entries | ✅ | `selfcheck: OK`, exit 0. Fixtures at `factory-tui.py:1386-1420` use a fixed `t0`, no filesystem; height sweep widened back to 66 |
| No new dependency, no write outside the process, `factory.py --check` green | ✅ | Diff `21b9b86...ee65dfa` adds **0** `import` / `open(` / `write(` / `subprocess` lines; `uv run factory.py --check` → `selfcheck: OK` |
| **Adjacent flow — TIMELINE must not evict the always-on wall board / radio** | ✅ **FIXED** | **4 800** head-vs-base frame pairs (0/1/5/6/7/16 active agents × alarms × pending × intercom × err × 0/5 concurrent loops × widths 60/80/92/120/200 × **h = 40…99**): **0** divergences in wall-board / radio / closing-border presence. A separate zoom-kept sweep (30 desks, `keep=` defeating the trim) at 4 widths × h=45…109: **0** divergences |

---

### ❌ The one remaining failure — the 5-row cap silently drops agents at *any* terminal height

`render_frame` still slices `shifts[:TIMELINE_ROWS]` unconditionally (`factory-tui.py:1096`), so the
panel shows at most 5 of the 16 roster agents no matter how much room exists. `ee65dfa` only
addressed the eviction; the cap itself is unchanged from `b55eb3f`.

Real curses, **200 cols × 120 rows** pty, real log dir — a terminal with ~50 rows to spare:

```
│ ═══ ⏱ TIMELINE 6h ══════════════════════════════════════════════════════════════════════════│
│ 🔍 SCOUT  ░░░…░░▓▓▓░░░░…░░ │
│ 🔥 ROAST  ░░░…░░▓▓▓▓░░░░…░ │
│ 💡 IDEA-P ░░░…▓▓░░░░…░░▓▓░ │
│ 🎨 IDEA-T ░░░…▓▓░░░░░…░░░░ │
│ 🧪 QA     ░░░…░▓▓▓▓▓▓▓▓▓▓▓ │
│ … +2 autres                                                                                  │
│ ═══ 🚨 ALARMES ═════…                       ← wall board + 📻 + closing border all present ✅
```

Offline replica, real log dir, width 120, agents present = `['scout','roast','ideator-product',
'ideator-tui','qa','rebaser','docagent']`:

```
agents= 7 h= 66 rendered_body=6 names=[SCOUT, ROAST, IDEA-P, IDEA-T, COORD, '… +2 autres']
agents= 7 h=100 rendered_body=6 names=[SCOUT, ROAST, IDEA-P, IDEA-T, COORD, '… +2 autres']
agents= 7 h=200 rendered_body=6 names=[SCOUT, ROAST, IDEA-P, IDEA-T, COORD, '… +2 autres']
agents= 7 h=400 rendered_body=6 names=[SCOUT, ROAST, IDEA-P, IDEA-T, COORD, '… +2 autres']
agents=16 h=400 rendered_body=6 names=[SCOUT, ROAST, IDEA-P, IDEA-T, COORD, '… +11 autres']
```

This breaks acceptance criterion 1 verbatim, and it breaks it exactly where the feature was supposed
to earn its keep: truncation is in fixed roster order, so the agents that fall off are always the
tail — `rebaser`, `promoter`, `docagent`, `reporter`, `cleaner`, i.e. the timer-driven desks the
issue's motivating question is about (*"did scout actually run this morning, or has its timer been
dead since 9 h?"*). On the live box right now the operator cannot see REBASE or DOC activity on any
screen size.

**Sub-case still open — at exactly 6 active agents the tail costs a row and buys nothing:**

```
agents=5  rendered body=5  [SCOUT, ROAST, IDEA-P, IDEA-T, COORD]
agents=6  rendered body=6  [SCOUT, ROAST, IDEA-P, IDEA-T, COORD, '… +1 autres']   ← PLAN hidden
agents=7  rendered body=6  [SCOUT, ROAST, IDEA-P, IDEA-T, COORD, '… +2 autres']
```

Six agents occupy six body rows either way — the `… +1 autres` row sits in the exact row it
displaced and says strictly less. That is the failure mode `alarm_shown()` exists to prevent and
documents in its own docstring (`factory-tui.py:981-984`: *"at exactly one row over the cap the tail
would occupy the row it displaced and say strictly less, so summarise only when it buys space
back"*). The fix copied `ALARM_ROWS`' cap but still not its guard. Now that `timeline_fits()` gates
the panel on real room, `TIMELINE_ROWS` no longer has to be a hard 5 at all — the natural shape is
`shift_shown()` mirroring `alarm_shown()`, sized against the room `timeline_fits()` already computes.

---

### Adversarial pass — what was attacked, per criterion

**Criterion 1 (rows per agent).** Heights 66/100/200/400 and real curses at 200×120 and 92×66;
agent counts 0/1/4/5/6/7/16 (synthetic dir with one log per roster prefix); live dir. The cap held
at 5 in every single case → ❌. Roster ordering itself is correct (16-agent synthetic dir renders
strict `ROSTER` order); the defect is only the truncation.

**Criterion 2 (extents).** Boundary: run of 1 s, of 0 s, straddling a slot boundary, exactly on
`win_start`, exactly on `now`, 1 s before `win_start`, spanning the whole window, at `cols=1`.
Hostile: `end < start` (backwards clock), `end` far in the future, `start` in the future, a run
still being appended 10 h later. Malformed names: `README`, `qa.log`, `-20260801T090000.log`
(no prefix), `qa-notatimestamp.log`, `qa-99991231T235959.log` (year 9999), non-roster prefixes —
all correctly skipped while the valid runs keep rendering. Nothing broke it.

**Criterion 3 (log-dir error path).** Nonexistent dir, `chmod 000` dir, `FACTORY_LOG_DIR` pointing
at a regular file. All three: no panel, no traceback, 0 bytes stderr, exit 0.

**Criterion 4 (empty window).** Empty dir, only-20 h-old logs, only-non-roster-prefix logs, plus the
`FACTORY_TIMELINE_H=1` window with a 6 h-old run → panel absent, not an empty box.

**Criterion 5 / adjacent flow (layout).** The main attack this round. 4 800 head-vs-base frame pairs
over the cartesian product {0,1,5,6,7,16 active agents} × {0,7 failed units} × {9 pending} ×
{3 intercom} × {err row} × {0,5 concurrent loops per role} × {60,80,92,120,200 cols} × h = 40…99,
each rendered through the real `desk_cap`/`build_desks`/`render_frame` chain and truncated at `h-1`
exactly as `main_curses` does. Head and base agree on wall-board / 📻 / closing-border presence in
**every** pair. Specifically re-ran the two sizes that failed last round: **92×66** and **92×60** —
head now matches base (board present at h≥58, radio at h≥61, both pre-existing base behaviour), and
`timeline_fits()` hides the panel below h=73 at 92 cols instead of evicting them. Also attacked the
`keep=` path (`build_desks` refusing to trim a zoomed desk, so desks can exceed the floor): 30 desks,
4 widths, h=45…109 → 0 divergences. Width uniformity swept 13→500 cols with the full roster: 0
mismatches.

**Criterion 6 (knob abuse).** `FACTORY_TIMELINE_H` = `6h`, `abc`, `3.5`, `1e3`, `""`, `" "`, `0`,
`-5`, `24`, `99999999`, `999999999999999999999`. No crash, no traceback, header always consistent
with the effective window.

**Persisted / pre-existing state.** Ran against the real log dir with 41 files spanning two days
(runs predating this PR, `rebase-*` legacy prefixes, a 0-byte in-progress log) — all parsed
correctly, no special-casing of "new" data.

**Auth & scope:** N/A — read-only local TUI, no network input, no user-supplied path beyond
`FACTORY_LOG_DIR` (already covered above).

**Hardening nit (carried over, still not a criterion flip).** A filename matching `\d{8}T\d{6}` that
is not a real datetime makes `time.strptime` raise inside the scan loop; the broad
`except Exception: pass` swallows it and **the whole panel silently disappears** while the 📻 ticker
keeps working — indistinguishable from "nobody ran". Reproduced exactly: a dir containing one valid
`qa-20260801T102645.log` **plus** `qa-20261301T250000.log` (month 13) → panel gone, 0 bytes stderr.
Not reachable from the loop's own `date +%Y%m%dT%H%M%S` names, so hardening, not a break. A
`try/continue` around the per-file parse would contain it.

---

**Verdict: FAIL** — 1 of 10 criteria fails. The layout regression from the last two rounds is
genuinely fixed and verified across 4 800 frame comparisons; what remains is the original
`TIMELINE_ROWS = 5` truncation, which hides `rebaser` and `docagent` on the live box at every
terminal size. Fixer summoned.

SoulKyu:

SoulKyu (commented):
@SoulKyu thanks for the round-4 fix — this closes out the wall-board/radio eviction regression the previous review rounds kept flagging.

I re-read `timeline_fits()`, `desk_floor_rows()`, and the updated `panel_rows()`/`desk_cap()` wiring in `devtools/factory-tui/factory-tui.py`, then verified it empirically rather than just by inspection: I loaded the module directly and swept `desk_cap()`/`render_frame()` across terminal widths 60-200, heights 30-119, and 1/3/5/6/8/16 concurrently-active agents (2,520+ combinations), comparing each frame against the same terminal size with an empty shift window (the pre-fix baseline). Zero cases where the TIMELINE panel caused the wall board or radio to be pushed off-screen, or the frame to overflow `h-1`, beyond what the empty-timeline baseline already did at that same size. `desk_floor_rows()` correctly reserves room for the ROSTER floor before `timeline_fits()` decides whether the panel gets drawn — that's exactly the root-cause fix the prior rounds were asking for, not just a tighter cap on the panel's own rows.

I also re-ran `compute_shifts()` against clock-skew (start after end), a run entirely in the future, and a negative start — all degrade gracefully (single-slot mark, silently excluded, or full-window mark, respectively), consistent with what the first review round already verified. `--check` passes and the file compiles clean.

Nice work tracking down the actual interaction between the desk-grid floor and the panel budget instead of re-capping the panel's own row count again.



SoulKyu:
## 🧪 QA Report

Re-test after the two fix commits (`66aac61` cap, `8c2b10d` `timeline_fits()`). Devtools-only diff
(`devtools/factory-tui/*`, no Go/webui/compose change), so the app under test is the TUI itself:
run live under **real tmux terminals** (3.4) at 60×50 → 160×200 against the real
`~/.claude-agents/notificator/logs` (57 files, 7 agents in the 6h window), plus a deterministic
A/B harness that renders base (`79bbe3d`) and PR frames from *identical* panel snapshots and
diffs them.

**The eviction regression from the previous head (`7c05bf5`) is genuinely fixed** — see the
adversarial pass. A different, unconditional failure took its place.

|---|---|---|
| `⏱ TIMELINE` panel renders **one row per agent** with ≥1 run in the window, roster order | ❌ | `TIMELINE_ROWS = 5` is a hard cap with no size condition. Live box, **160×200 terminal** (nothing else is even close to full): 7 agents ran in the window, **5 rows rendered + `… +2 autres`**. `rebaser` and `docagent` are invisible at *any* terminal size. Repro below |
| Run extents from filename (start) + mtime (end); short run never invisible | ✅ | `compute_shifts`: 1 s run → 1 slot, 0 s run (`start==end`) → 1 slot, `end<start` clock skew → 1 slot, `end` in the future → clamped to `now` (7/40 slots for a 1 h run), start in the future → dropped, run entirely before the window → dropped, exactly on `win_start` → idx 0, exactly on `now` → last idx, `cols=1` → `▓`, `window=100000h` → no crash |
| Missing/unreadable `FACTORY_LOG_DIR` → no panel, no traceback | ✅ | Nonexistent dir, `chmod 000` dir, path-is-a-regular-file: all three → `--once` exit 0, **0 bytes on stderr**, no panel |
| Empty window → panel hidden entirely | ✅ | Empty dir → panel absent from the frame (not an empty box) |
| Slot count adapts to panel width; all frame rows keep identical display width | ✅ | A/B sweep, widths **60/80/92/120/200** × heights **40–94**, 7 panel-state scenarios incl. the `… +N autres` tail row: `dwidth()` identical on every row of every frame, 0 mismatches. Live resize storm 160×200→92×66→160×200→120×74→60×50→160×200: no row ever exceeds the terminal width, no traceback |
| `FACTORY_TIMELINE_H` overrides; garbage falls back without crashing | ✅ | `1/3/24` → follow; `6h`,`abc`,`""`,`" "`,`3.5` → 6; `0`,`-5` → 1; `99999999` → no crash |
| `--once` still prints a frame and exits 0 | ✅ | exit 0, 83 rows, empty stderr |
| `--check` still passes, with a `compute_shifts()` assertion on synthetic entries | ✅ | `selfcheck: OK`, exit 0 — and it has teeth, see mutation A |
| No new dependency, no write outside the process, `factory.py --check` green | ✅ | Diff adds zero `import`/`open()`/`subprocess`/write lines; `uv run factory.py --check` → `selfcheck: OK` |

### ❌ The failure, with the exact repro

```
$ tmux new-session -d -s tall -x 160 -y 200 'python3 devtools/factory-tui/factory-tui.py'
$ tmux capture-pane -p -t tall
│ ═══ ⏱ TIMELINE 6h ══════════════════════════════════════════════════════════════════════│
│ 🔍 SCOUT  ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░▓▓▓░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ │
│ 🔥 ROAST  ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░▓▓▓▓░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ │
│ 💡 IDEA-P ░░░░░░░░░░░░░░░░░░░░░░░░░▓░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░▓▓░░░░░░░ │
│ 🎨 IDEA-T ░░░░░░░░░░░░░░░░░░░░░░░░░▓░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ │
│ 🧪 QA     ░░░░░░░░░░░░░░░░░░░░░░░░░▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓ │
│ … +2 autres                                                                             │

$ python3 -c "... poll_med(); shift_rows(panel_snapshot(), 100)"
agents with a run in the 6h window: ['scout','roast','ideator-product','ideator-tui','qa','rebaser','docagent']
rendered (cap 5):                   ['scout','roast','ideator-product','ideator-tui','qa']
HIDDEN behind the tail row:         ['rebaser','docagent']
```

`rebaser` ran, `docagent` ran, and the panel says neither did — on a 200-row terminal with ~90 free
rows. Because the cap slices `shifts[:5]` in **roster order**, the hidden set is always the tail of
`ROSTER` (`fixer`→`cleaner`), so the issue's own motivating question — *"did `scout` actually run
this morning?"* — is answerable for `scout` (roster #1) and permanently unanswerable for `cleaner`
(#16). `… +2 autres` does not even name who is missing. This is the criterion the panel exists for.

**The cap is not load-bearing** (evidence in mutation B below): `timeline_fits()` from `8c2b10d`
already reserves exactly the rows the panel needs and hides the whole panel when they are not
there. Suggested fix: drop `TIMELINE_ROWS` (or raise it to `len(ROSTER)`) and let `timeline_fits()`
be the only gate — `panel_rows()`/`render_frame()` already agree on the uncapped cost.

### Adversarial pass

**Adjacent flow — the previous head's regression (wall board / 📻 evicted).** Deterministic A/B
harness: base `79bbe3d` and PR rendered from *identical* panel snapshots, widths **60/80/92/120/200**
× heights **40–94**, 7 scenarios (quiet; 1/3/4/5/6/16 agents in the window; `intercom`×6;
`pending`×8-9; `alarms`×3; `err` set), asserting on `TABLEAU DU MUR`, 📻, the closing `└`, frame
overflow past `h-1`, and row-width equality — **0 regressions vs base in every cell**. Live
confirmation at 92×{66,72,80,90} and 60×50, base vs PR side by side: identical outcomes at every
size (at 92×66 the board is lost on **base too**, with today's 6-message intercom — pre-existing,
not this PR). **Fixed.**

**Mutation A — does the new `--check` assertion have teeth?** Neutered `timeline_fits()` back to
always-True in a copy → `python3 factory-tui.py --check` → **12 FAILURES**, naming exactly the old
symptom (`FAIL desk cap: radio ticker at row 69 of a 66-row terminal`, h=66..72). The regression
test is real, not decorative.

**Mutation B — is the 5-row cap needed for layout?** Set `TIMELINE_ROWS = 99` in a copy (cap gone,
`timeline_fits()` kept) → `--check` → `selfcheck: OK`, and the full A/B sweep with 16 agents in the
window + every panel busy → **0 regressions** across all widths/heights. The cap protects nothing;
it only deletes rows.

**Live resize storm.** 160×200 → 92×66 → 160×200 → 120×74 → 60×50 → 160×200 in one session: panel
appears/disappears cleanly at each step, no traceback, no row wider than the terminal, closing
border always present.

**Adjacent flow — keys.** With the panel on screen at 160×200: `←/→/↓` navigation, `Enter` zoom
open, `Esc` zoom close, `?` legend open/close. Wall board, 📻 and the timeline survive all six
states; no traceback; overlay never overruns the terminal width.

**Malformed / hostile log dir.** `README`, `qa.log`, `-<ts>.log` (no prefix), `garbage-<ts>.log`,
non-roster prefixes → all skipped, the valid runs still render. `agent_of_log`: `qa-131`→`qa`,
`rebase-116`→`rebaser`, `qa-131-999`/`-`/`""`/`garbage`→`None`.

**Error / unavailable path.** Missing dir, `chmod 000` dir, path-is-a-file → panel gone, no
traceback, `--once` exits 0 with empty stderr (checked byte count, not eyeballed).

**Knob abuse.** `FACTORY_TIMELINE_H` = `6h`, `abc`, `""`, `" "`, `3.5`, `0`, `-5`, `1e3`,
`99999999` → no crash, header always consistent with the effective window.

**Auth/scope:** N/A — read-only local TUI, no network input, no user-supplied data path.

**Hardening nit (unchanged, not a criterion flip).** A filename matching `\d{8}T\d{6}` that is not a
real datetime (`qa-20261301T250000.log`, month 13) still makes `time.strptime` raise inside the scan
loop; the broad `except Exception: pass` swallows it and **the whole panel silently disappears**
while the 📻 ticker keeps working — indistinguishable from "nobody ran". Verified again at this head:
2 valid runs present, panel gone, stderr empty. Not reachable from the loop's own
`date +%Y%m%dT%H%M%S` names, so hardening; a `try/continue` around the per-file parse would contain it.

**Verdict: FAIL** (8/9 criteria pass; criterion 1 fails unconditionally, at every terminal size.)

SoulKyu (commented):
@SoulKyu thanks for the round-5 follow-up. I reviewed the delta since the last clean pass (`8c2b10dd` → `f21d139b`): `TIMELINE_ROWS` changes from a fixed cap of `5` to `len(ROSTER)` (16).

This closes a real gap the previous fix left open: with the fixed cap, `shift_rows()` returns agents in `ROSTER` order, so if more than 5 distinct agents had activity in the window, everything past the 5th (`rebaser`, `promoter`, `docagent`, `reporter`, `cleaner`, …) collapsed into an anonymous "+N autres" line — defeating the panel's whole purpose of answering "did X run". Since `shift_rows()` can never emit more than one row per `ROSTER` key, `TIMELINE_ROWS = len(ROSTER)` is not actually a cap anymore, it's just a safe upper bound, and the `shifts[:TIMELINE_ROWS]` / `len(shifts) > TIMELINE_ROWS` tail-summary path in `render_frame` becomes intentionally dead code (harmless future-proofing, not a bug).

The part I went looking for adversarially: raising `TIMELINE_ROWS` to 16 also raises the worst-case `shift_cost` inside `timeline_fits()`/`panel_rows()` from `1 + 5 + 1 = 7` to `1 + 16 = 17` rows — more than double what the previous round's eviction-fix sweep actually exercised (that sweep ran with the old cap of 5). If the new self-check didn't cover this, the round-3/round-4 wall-board/radio eviction regression could easily have resurfaced under this larger budget. I confirmed it's covered: the `selfcheck()` fixture at `STATE["shifts"]` now sets all 16 `ROSTER` keys running simultaneously and sweeps `term_h` 66–82 asserting the wall board and radio ticker stay on-screen and the frame never exceeds `h-1`. I ran `--check` locally and it passes (`selfcheck: OK`), and traced the arithmetic by hand for the `n_shift == len(ROSTER)` case to confirm `timeline_fits()` correctly accounts for the larger `shift_cost` before deciding whether to hide the panel.

Also re-ran `--once` (exits 0, valid frame) and `py_compile` on the file — no syntax/behavior regressions. The `README.md` diff here is documentation-only carryover from the earlier round, unchanged in this commit.

Good catch tracking down that the cap itself — not just the eviction math — was silently hiding agents. No actionable findings from me on this round.



SoulKyu:
## 🧪 QA Report

Re-test after `f21d139` (*stop TIMELINE panel from hiding agents behind a row cap*). Diff is
**devtools-only** (`devtools/factory-tui/*`, zero Go/webui/compose files — `git diff --name-only`
confirms), so the app under test is the TUI itself: `--check` / `--once`, **real curses in tmux 3.4
terminals** at 60×40 → 200×130 against the real `~/.claude-agents/notificator/logs` (43 files) *and*
a synthetic log dir with **all 16 roster agents** running, plus a deterministic offline A/B harness
replaying `panel_snapshot → desk_cap → build_desks → render_frame(h) → break at h-1` on both the PR
and the pre-feature base `79bbe3d`.

**Both previously reported failures are gone.** The eviction regression (head `7c05bf5`) stays
fixed, and the 5-row cap that made `rebaser`/`docagent` invisible (head `8c2b10d`) is now
`TIMELINE_ROWS = len(ROSTER)`, so `shift_rows()` can no longer be truncated.

|---|---|---|
| `⏱ TIMELINE` panel renders **one row per agent** with ≥1 run in the window, roster order | ✅ | Live tmux **160×90** with a 16-agent synthetic log dir: **16 rows, all 16 agents named** (`SCOUT…CLEAN`), exact `ROSTER` order, **no `… +N autres` tail**. Offline sweep (4 widths × 26 heights × n∈{1,5,7,16}): whenever the panel is drawn, rendered rows == agents-with-a-run, 0 mismatches. `worker-2`→worker, `qa-131`→qa, `rebase-83`→rebaser |
| Run extents from filename (start) + mtime (end); short run never invisible | ✅ | `compute_shifts`: 1 s → 1 slot, 0 s (`start==end`) → 1 slot, `end<start` skew → 1 slot, `end` in future → clamped to `now`, start in future → dropped, entirely before window → dropped, exactly on `win_start` → idx 0, exactly on `now` → last idx, 1 s before `win_start` → dropped, spans window → full bar, `cols=1` → `▓`, `cols=5000` → 5000 chars, `window=100000h` → no crash |
| Missing/unreadable `FACTORY_LOG_DIR` → no panel, no traceback | ✅ | Nonexistent dir, `chmod 000` dir, path-is-a-regular-file → `--once` exit 0, **0 bytes on stderr**, no panel |
| Empty window → panel hidden entirely | ✅ | Empty dir → no panel. Log started 20 h ago with mtime 20 h ago → no panel; same file still being appended (mtime now) → full bar, correctly |
| Slot count adapts to panel width; all frame rows keep identical display width | ✅ | Width sweep **13 → 500** with 16 agents on screen: `dwidth()` identical on every row of every frame, **0 non-uniform frames** (the pre-existing ≤12-col title-row nit no longer reachable at 13). Live resize storm 160×100 → 92×66 → 200×130 → 60×40 → 160×100: no row exceeds the terminal, no traceback |
| `FACTORY_TIMELINE_H` overrides; garbage falls back without crashing | ✅ | `1/3/24` follow; `6h`,`abc`,`""`,`" "`,`3.5`,`1e3` → 6; `0`,`-5` → 1; `99999999` → header `99999999h`, no crash. stderr 0 bytes in all 12 runs |
| `--once` still prints a frame and exits 0 | ✅ | exit 0, 80 rows, empty stderr |
| `--check` still passes, with a `compute_shifts()` assertion on synthetic entries | ✅ | `selfcheck: OK`, exit 0 — and the layout half has teeth (mutation A) |
| No new dependency, no write outside the process, `factory.py --check` green | ✅ | Diff adds **0** `import`/`open()`/`subprocess` lines; `uv run factory.py --check` → `selfcheck: OK` |
| **Adjacent flow — wall board / 📻 / closing border must not be evicted** | ✅ | A/B vs base `79bbe3d`, widths 60/92/120/160 × heights 30–130 × 6 panel scenarios (quiet, 1/5/7/16 agents, everything-busy): **0 `TABLEAU DU MUR` losses, 0 📻 losses, 0 `└` losses, 0 new frame overflows past `h-1`**. Live confirmation at 92×66, 92×80, 120×74, 160×60/90, 200×120 |

### Adversarial pass

**The fix's own blast radius — does raising the cap to 16 re-break eviction?** This was the whole
risk of `f21d139`: `timeline_fits()` now has to reserve up to 17 rows instead of 7. Full A/B sweep
above (624 rendered frame pairs) → **no regression vs base in any cell**. Measured the availability
cost directly, PR vs previous head `8c2b10d`, at 92 cols with a live-shaped panel stack:
`n=3 → h≥74 both`, `n=5 → 76 both`, `n=7 → 77 vs 78`, `n=16 → 77 vs 87`. Realistic rosters pay
**+1 row**; only a full 16-agent roster pays 10 — which is the correct trade (see caveat below).

**Mutation A — is the layout regression test real?** Neutered `timeline_fits()` to always-True in a
copy → `--check` → **39 FAILURES** naming the old symptom (`FAIL desk cap: radio ticker at row 79 of
a 79-row terminal`). Not decorative.

**Mutation B — is the raised cap load-bearing, or cosmetic?** Set `TIMELINE_ROWS = 99` → `--check`
`OK` and the A/B sweep stays clean (nothing depends on the exact value, `shift_rows()` is bounded by
`len(ROSTER)` anyway). Set it back to the buggy `5` → **`--check` still says `OK`** (see gap 1).

**Adjacent flow — keys.** Panel on screen at 160×100 with 16 agents: `→`, `↓`, `Enter` (zoom open),
`Esc` (zoom close), `?` (legend open/close). Board, 📻 and the timeline survive all six states, max
row width never exceeds the terminal, 0 bytes on stderr across the whole session.

**Malformed / hostile log dir.** `README`, `qa.log`, `-<ts>.log` (no prefix), `garbage-<ts>.log`,
`qa-131-999-<ts>.log`, non-roster prefixes → all skipped, the valid runs still render alongside them.

**Error / unavailable path.** Missing dir, `chmod 000` dir, path-is-a-file → panel gone, no
traceback, exit 0, stderr byte count checked (0), not eyeballed.

**Scale / DoS-ish.** 5 000 log files in `LOG_DIR` → `poll_med()` 0.15 s, one correct `worker` row.
No UI freeze.

**Auth & scope:** N/A — read-only local TUI, no network input, no user-supplied data path, no writes.

**Persisted state:** the real 43-file log dir predates the feature and renders correctly (3 agents in
the current 6 h window); no migration or on-disk format is involved.

### Caveats (named, not criterion flips)

1. **No regression test for this very fix.** Reverting `TIMELINE_ROWS` to `5` in a copy still yields
   `selfcheck: OK` — the bug QA reported at head `8c2b10d` can silently return. A one-line assertion
   (`len(shift_rows(p, 1)) == len(ROSTER)` renders `len(ROSTER)` rows, or simply
   `TIMELINE_ROWS >= len(ROSTER)`) would pin it. Behaviour is correct today, so this is coverage, not
   a defect.
2. **Visibility threshold.** With a full 16-agent roster at 92 cols the panel needs `h ≥ 87` and
   hides itself below that. That is the issue's own sanctioned rule (*"drops out first on short
   terminals"*) and strictly better than the alternatives already rejected — evicting the wall
   board, or lying about who ran. Realistic 3–7 agent rosters are unaffected (+1 row vs the previous
   head).
3. **Hardening nit (unchanged, still unreachable).** A filename matching `\d{8}T\d{6}` that is not a
   real datetime (`qa-20261301T250000.log`, month 13) makes `time.strptime` raise inside the scan
   loop; the pre-existing broad `except Exception: pass` swallows it and the **whole panel silently
   disappears** while 📻 keeps working. Re-verified at this head. Not producible by the runners'
   `date +%Y%m%dT%H%M%S`, so hardening — a `try/continue` around the per-file parse would contain it.

*Note on scope: the diff touches no backend/webui/compose file, so the docker stack carries no signal
for this change; the TUI was exercised live instead. `docker compose down -v` run at teardown.*

**Verdict: PASS**

Issue #161, opened 2026-08-01
fix(webui): the CaptureAlertFired goroutine reads a cache-resident alert pointer — data race with the next refresh cycle


SoulKyu:
## Problem

`refreshAlerts` spawns a background goroutine per newly-seen alert to record the fire event, and hands it the pointer it just stored in the cache map rather than a snapshot (`internal/webui/services/alert_cache.go:239-255`):

```go
if existingAlert, exists := ac.alerts[fingerprint]; !exists {
    ac.alerts[fingerprint] = dashAlert          // :240 — pointer goes into the cache
    ac.newAlerts = append(ac.newAlerts, fingerprint)
    alertCopy := *dashAlert                     // :245 — SSE payload IS snapshotted
    newAlertsForSSE = append(newAlertsForSSE, &alertCopy)

    ac.runBounded(func() {                      // :249
        if ac.backendClient != nil && ac.backendClient.IsConnected() {
            if err := ac.backendClient.CaptureAlertFired(dashAlert); err != nil {   // :251 — cache memory, no lock
```

`CaptureAlertFired` reads `alert.Annotations` (`internal/webui/client/backend_client.go:1788`) and `alert.Status` (`:1811`, for the `SilencedAtFire` flag) with no lock held. Both fields are rewritten by the next poll cycle under `ac.mu` in `updateExistingAlert` — `existing.Status = new.Status` (`alert_cache.go:414`) and `existing.Annotations = new.Annotations` (`:419`) — on the same struct, because the alert now exists in the map and takes the `else` branch.

The window is wide enough to hit. `runBounded` caps concurrency at `maxBackendWorkers = 8` (`alert_cache.go:55, 198-204`), each `CaptureAlertFired` has a 3-second timeout (`backend_client.go:1779`), and the default refresh interval is 10 seconds. A burst of new alerts — an Alertmanager coming back after an outage, or the first refresh after a WebUI restart, where every alert is "new" — queues far more than 8 goroutines. 500 new alerts against a slow or unreachable backend is up to ~3 minutes of backlog, spanning a dozen refresh cycles that are all rewriting `Status` and `Annotations` on those exact structs.

[...]

PR #162, opened 2026-08-01
docs(spec): fix(webui): the CaptureAlertFired goroutine reads a cache-resident alert pointer — data race with the next refresh cycle

SoulKyu:
## Summary
- Adds a planning spec for #161: the fire-capture goroutine spawned by `refreshAlerts` reads a cache-resident `*DashboardAlert` (`Status`/`Annotations`) with no lock while the next refresh cycle concurrently rewrites those same fields under `ac.mu` — a genuine `-race` finding, not a false positive.
- Spec proposes the one-line fix (snapshot before spawning, matching the existing pattern at `alert_cache.go:302-304`), documents why the neighbouring `UpdateAlertResolved` goroutine is safe without a copy (the alert is deleted from the map under the same lock right after), and lays out a `-race`-driven regression test using a real (unreachable) `*client.BackendClient` since the field has no interface seam.

## Test plan
- [ ] Spec-only change — no code touched, `go build`/`go vet`/`go test` untouched by this PR.

Relates to #161 (implementation to follow in a separate PR)

<sub>🔁 Powered by <a href="https://github.com/nexu-io/looper">Looper</a> · runner=planner · agent=claude-code · An autonomous AI dev team for your GitHub repos.</sub>

### DIFF
--- a/specs/2026-08-01-161-fix-webui-the-capturealertfired.md
+++ b/specs/2026-08-01-161-fix-webui-the-capturealertfired.md
@@ -1,0 +1,212 @@
+# Spec: fix `CaptureAlertFired` reading a cache-resident alert pointer
+
+- Issue: [SoulKyu/notificator#161](https://github.com/SoulKyu/notificator/issues/161)
+- Date: 2026-08-01
+- Status: planned
+
+## Problem
+
+`refreshAlerts` spawns a background goroutine per newly-seen alert to record
+the fire event, and hands it the pointer it just stored in the cache map
+instead of a snapshot
+(`internal/webui/services/alert_cache.go:239-255`):
+
+```go
+if existingAlert, exists := ac.alerts[fingerprint]; !exists {
+    ac.alerts[fingerprint] = dashAlert          // :240 — pointer goes into the cache
+    ac.newAlerts = append(ac.newAlerts, fingerprint)
+    alertCopy := *dashAlert                     // :245 — SSE payload IS snapshotted
+    newAlertsForSSE = append(newAlertsForSSE, &alertCopy)
+
+    ac.runBounded(func() {                      // :249
+        if ac.backendClient != nil && ac.backendClient.IsConnected() {
+            if err := ac.backendClient.CaptureAlertFired(dashAlert); err != nil {   // :251 — cache memory, no lock
+```
+
+`CaptureAlertFired` reads `alert.Annotations`
+(`internal/webui/client/backend_client.go:1812-1813`) and `alert.Status`
+(`:1836`, for `SilencedAtFire`) with no lock held. Both fields are rewritten
+by the next poll cycle under `ac.mu` in `updateExistingAlert` —
+`existing.Status = new.Status` (`alert_cache.go:414`) and
+`existing.Annotations = new.Annotations` (`:419`) — on the same struct,
+because once the alert is in the map the next cycle takes the `else` branch
+and mutates it in place.
+
+The window is wide enough to hit. `runBounded` caps concurrency at
+`maxBackendWorkers = 8` (`alert_cache.go:55, 198-204`), each
+`CaptureAlertFired` has a 3-second timeout (`backend_client.go:1804`), and
+the default refresh interval is 10 seconds. A burst of new alerts — an
+Alertmanager coming back after an outage, or the first refresh after a WebUI
+restart — queues far more than 8 goroutines, spanning multiple refresh
+cycles that are all rewriting `Status`/`Annotations` on those exact structs.
+
+`Status` (`internal/webui/models`, `AlertStatus{State string; SilencedBy,
+InhibitedBy []string}`) is a multi-word struct, so the assignment at `:414`
+is not atomic. A concurrent reader can observe a `State` string header whose
+pointer and length come from different writes; comparing it
+(`alert.Status.State == "silenced"`) then reads out of bounds. The benign
+outcome is a wrong `silenced_at_fire` flag permanently recorded in
+`alert_statistics`; the bad one is a segfault in the WebUI process.
+
+This is a leftover of the invariant #54 established. Commit `11ef6a9`
+("stop alert cache from handing out live alert pointers") converted the
+read accessors and the acknowledge statistics goroutine to snapshots but
+did not touch this call site, and `b3d9002` later added exactly this copy
+for the *neighbouring* goroutine twelve lines down —
+`alert_cache.go:302-304`. The same reasoning was not carried over here.
+
+## Goals
+
+- `CaptureAlertFired` is invoked with a snapshot taken under `ac.mu`, never
+  the pointer stored in `ac.alerts`.
+- A `-race`-sensitive test reproduces the bug against current code and
+  passes after the fix.
+- The `UpdateAlertResolved` goroutine (`alert_cache.go:292-298`), which
+  closes over a cache-resident pointer the same way, is checked for the same
+  hazard and either fixed or documented as safe.
+
+## Non-goals
+
+- No change to `CaptureAlertFired`'s or `UpdateAlertResolved`'s signature,
+  or to the backend RPC contract.
+- No broader refactor of `refreshAlerts` or `runBounded` — this is a
+  targeted snapshot fix, not a redesign of the refresh pipeline.
+- No deep copy: per the existing convention at `alert_cache.go:695-698`,
+  `Labels` is never rewritten after construction and `Annotations` is only
+  ever replaced wholesale, never mutated in place, so a shallow copy is
+  sufficient.
+
+## Approach
+
+### 1. Snapshot before spawning the fire-capture goroutine
+
+Mirror what `alert_cache.go:302-304` already does for the resolved path, in
+the `!exists` branch of `refreshAlerts` (`alert_cache.go:239-255`):
+
+```go
+// Snapshot, never the cache-resident pointer: Status/Annotations are
+// rewritten under ac.mu by the next refresh cycle's updateExistingAlert
+// while this goroutine reads them without a lock. Same reasoning as the
+// resolved-alert copy below.
+firedCopy := *dashAlert
+ac.runBounded(func() {
+    if ac.backendClient != nil && ac.backendClient.IsConnected() {
+        if err := ac.backendClient.CaptureAlertFired(&firedCopy); err != nil {
+            log.Printf("Failed to capture alert fired statistics for %s: %v", firedCopy.Fingerprint, err)
+        }
+    }
+})
+```
+
+### 2. Document (not fix) `UpdateAlertResolved`
+
+`alert_cache.go:292-298` closes over the cache-resident pointer too, but
+`delete(ac.alerts, fingerprint)` happens a few lines below (`:306`), inside
+the same `ac.mu` critical section that started at `:222` and doesn't unlock
+until `:315`. Once deleted, no code path can reach that exact struct again:
+
+- `MutateAlert`, `GetAlert`, `GetAllAlerts`, `GetLiveAlert` all look the
+  alert up by map key under `ac.mu` — none resolve to this pointer once the
+  key is gone.
+- If the same labels fire again later, `refreshAlerts` takes the `!exists`
+  branch and allocates a brand-new `dashAlert`, not this one.
+- `storeResolvedAlertInBackend` already gets its own copy (`:303-304`), so
+  it isn't a second reader of the original.
+
+So the goroutine at `:292-298` is safe without a copy — add a comment
+recording why, so the next person doesn't have to re-derive it:
+
+```go
+// Unlike the fire-capture goroutine above, this is not copied: the alert
+// is deleted from ac.alerts a few lines below, inside this same lock, so
+// no concurrent writer can reach this exact struct again afterwards
+// (every lookup path keys off ac.alerts, which no longer holds it).
+ac.runBounded(func() {
+    if ac.backendClient != nil && ac.backendClient.IsConnected() {
+        if err := ac.backendClient.UpdateAlertResolved(alert); err != nil {
+            log.Printf("Failed to update alert resolved statistics for %s: %v", alert.Fingerprint, err)
+        }
+    }
+})
+```
+
+### 3. Regression test
+
+Add `TestAlertCache_ConcurrentFireCaptureAndRefresh` to
+`internal/webui/services/alert_cache_test.go`, following the pattern
+`TestAlertCache_ConcurrentRefreshAndReads` (added by `11ef6a9`) already
+established in this file.
+
+The tricky part: `ac.backendClient` is a concrete `*client.BackendClient`,
+not an interface, so there's no seam to inject a fake for `CaptureAlertFired`.
+Work with the real type instead of adding one:
+
+- `CaptureAlertFired` reads `alert.Annotations`/`alert.Status` while
+  building the gRPC request, *before* it sends anything over the wire — so
+  the race is exercised regardless of whether the RPC itself succeeds.
+- Point a real `*client.BackendClient` at a loopback address nothing is
+  listening on (`net.Listen("tcp", "127.0.0.1:0")` then `Close()`
+  immediately, reusing the freed port). `client.NewBackendClient(addr)` +
+  `.Connect()` uses `grpc.NewClient`, which dials lazily — `Connect()`
+  succeeds and `IsConnected()` reports true even though nothing answers.
+  The eventual RPC then fails fast with "connection refused" instead of
+  blocking for the 3-second timeout, and the error is just logged, matching
+  production behavior when the backend is down.
+- Set `cache.backendClient` to that client directly (test lives in package
+  `services`, so the unexported field is assignable).
+- Loop a handful of times (e.g. 8, matching `maxBackendWorkers`): each
+  iteration first refreshes with a brand-new fingerprint (`!exists` branch —
+  stores the pointer, spawns the fire-capture goroutine), then immediately
+  refreshes again with the *same* fingerprint but a changed `Status`
+  (e.g. `firing` → `silenced` with a `SilencedBy` entry) and `Annotations`
+  (`exists` branch — `updateExistingAlert` writes those fields on the same
+  cache-resident struct under `ac.mu`). No rendezvous between the two calls
+  is needed or wanted: the absence of synchronization between them is
+  exactly the bug, and `-race` flags conflicting accesses on the same
+  memory regardless of how the goroutine scheduler happens to interleave
+  them.
+- Before the test returns, wait for every spawned fire-capture goroutine to
+  actually finish its read — otherwise the process can exit before the race
+  detector observes it. `runBounded` doesn't expose a `WaitGroup`, but its
+  semaphore does: `cache.backendSem` (buffered channel, capacity
+  `maxBackendWorkers`) holds one token for the lifetime of each in-flight
+  call, so polling `len(cache.backendSem) == 0` on a short ticker is a
+  reliable "all goroutines done" signal.
+- Run: `go test -race -run TestAlertCache_ConcurrentFireCaptureAndRefresh ./internal/webui/services/...`.
+  Expected: fails with `WARNING: DATA RACE` on `Status`/`Annotations` against
+  today's code; passes cleanly once `refreshAlerts` snapshots before
+  spawning.
+
+### Files touched
+
+- `internal/webui/services/alert_cache.go` — snapshot before the
+  fire-capture `runBounded` call; add the safety comment on the
+  resolved-statistics goroutine.
+- `internal/webui/services/alert_cache_test.go` — new
+  `TestAlertCache_ConcurrentFireCaptureAndRefresh`.
+
+## Risks & trade-offs
+
+- **Shallow copy only**: consistent with the existing convention
+  (`alert_cache.go:695-698`, `:303-304`) — `Labels`/`Annotations` are
+  replace-wholesale-only fields, so a deep copy would be dead weight here.
+  If that invariant ever changes (in-place annotation mutation added
+  elsewhere), this call site and the others following the same pattern
+  would all need revisiting together.
+- **Test uses a real `*client.BackendClient`** instead of a fake, since the
+  field isn't an interface. This makes the test slightly more expensive
+  (real TCP dial attempt, real JSON/proto marshalling) but exercises the
+  exact production code path — a fake would risk not reproducing the read
+  order that causes the race in the first place.
+- **No fix to `UpdateAlertResolved`**: relies on the delete-under-the-same-lock
+  argument holding. If a future change moves the `delete(ac.alerts, ...)`
+  outside the lock, or adds another lookup path that doesn't key off
+  `ac.alerts`, that reasoning breaks silently — the comment is the only
+  guard against that regression.
+
+## Validation
+
+- `go build ./...`
+- `go vet ./...`
+- `go test -race ./internal/webui/...` (must include the new test passing,
+  and confirm no other race is newly surfaced in the package)
--- a/specs/2026-08-01-161-fix-webui-the-capturealertfired.md
+++ b/specs/2026-08-01-161-fix-webui-the-capturealertfired.md
@@ -60,7 +60,7 @@ for the *neighbouring* goroutine twelve lines down —
 - `CaptureAlertFired` is invoked with a snapshot taken under `ac.mu`, never
   the pointer stored in `ac.alerts`.
 - A `-race`-sensitive test reproduces the bug against current code and
-  passes after the fix.
+  passes after the fix — deterministically, not most of the time.
 - The `UpdateAlertResolved` goroutine (`alert_cache.go:292-298`), which
   closes over a cache-resident pointer the same way, is checked for the same
   hazard and either fixed or documented as safe.
@@ -69,8 +69,10 @@ for the *neighbouring* goroutine twelve lines down —
 
 - No change to `CaptureAlertFired`'s or `UpdateAlertResolved`'s signature,
   or to the backend RPC contract.
-- No broader refactor of `refreshAlerts` or `runBounded` — this is a
-  targeted snapshot fix, not a redesign of the refresh pipeline.
+- No broader refactor of `refreshAlerts` — this is a targeted snapshot fix,
+  not a redesign of the refresh pipeline. `runBounded` gains a two-line
+  completion counter (§3) and nothing else: no worker pool, no lifecycle
+  wiring into `Stop()`.
 - No deep copy: per the existing convention at `alert_cache.go:695-698`,
   `Labels` is never rewritten after construction and `Annotations` is only
   ever replaced wholesale, never mutated in place, so a shallow copy is
@@ -130,7 +132,64 @@ ac.runBounded(func() {
 })
 ```
 
-### 3. Regression test
+### 3. Give `runBounded` a completion signal the test can wait on
+
+The test in §4 has to know that every spawned fire-capture goroutine has
+*actually performed its read* before it returns; otherwise the process can
+exit with the racing read still pending and `-race` reports nothing.
+
+`runBounded` currently exposes no such signal, and the semaphore is not one:
+the token is taken **inside** the goroutine (`alert_cache.go:199-201`), so
+`len(ac.backendSem) == 0` means "nothing running *right now*" — which is
+also true for a call that was spawned microseconds ago and has not been
+scheduled yet. Any signal derived from `backendSem` — polling it, or filling
+it by acquiring all `maxBackendWorkers` tokens — has the same blind spot at
+the spawn edge, and a test built on one passes spuriously whenever the
+scheduler hasn't run the goroutine yet. Measured against unfixed code, the
+`len()` poll misses the race roughly 1 run in 10.
+
+The fix is to count the goroutine in on the *caller's* goroutine, before
+`go` — the only point that is ordered with respect to the spawn:
+
+```go
+type AlertCache struct {
+    // ...
+    backendSem chan struct{}  // semaphore bounding refresh-triggered backend calls  (:71)
+    backendWG  sync.WaitGroup // counts spawned backend calls, incremented before the goroutine starts
+```
+
+```go
+// runBounded runs fn on a goroutine while holding a slot in backendSem, so at
+// most maxBackendWorkers refresh-triggered backend calls run concurrently.
+func (ac *AlertCache) runBounded(fn func()) {
+    // Counted here, on the caller's goroutine, not inside the closure: the
+    // semaphore token is only taken once the goroutine is scheduled, so
+    // len(backendSem) reads 0 for calls that are spawned but not yet running.
+    // backendWG is the only signal that covers a spawned call from the moment
+    // it is created.
+    ac.backendWG.Add(1)
+    go func() {
+        defer ac.backendWG.Done()
+        ac.backendSem <- struct{}{}
+        defer func() { <-ac.backendSem }()
+        fn()
+    }()
+}
+```
+
+Production behaviour is unchanged — nothing calls `Wait()` outside tests, and
+`Add` runs on a goroutine that already holds `ac.mu`, so the counter is never
+touched concurrently from zero. `Wait()` in the test is sound because every
+`Add` happens-before it: the refresh calls that spawn the work all return
+first.
+
+`WaitGroup.Done` releases and `Wait` acquires, so the only happens-before
+edge this adds points *from* the goroutine *to* the waiter. It does not
+order the test's `updateExistingAlert` writes against the goroutine's reads,
+so the race stays visible to the detector — the counter makes the read
+observable, it does not synchronise it away.
+
+### 4. Regression test
 
 Add `TestAlertCache_ConcurrentFireCaptureAndRefresh` to
 `internal/webui/services/alert_cache_test.go`, following the pattern
@@ -165,22 +224,24 @@ Work with the real type instead of adding one:
   exactly the bug, and `-race` flags conflicting accesses on the same
   memory regardless of how the goroutine scheduler happens to interleave
   them.
-- Before the test returns, wait for every spawned fire-capture goroutine to
-  actually finish its read — otherwise the process can exit before the race
-  detector observes it. `runBounded` doesn't expose a `WaitGroup`, but its
-  semaphore does: `cache.backendSem` (buffered channel, capacity
-  `maxBackendWorkers`) holds one token for the lifetime of each in-flight
-  call, so polling `len(cache.backendSem) == 0` on a short ticker is a
-  reliable "all goroutines done" signal.
+- Grow the fake's payload across iterations rather than serving one alert at
+  a time: a fingerprint missing from a fetch is treated as resolved
+  (`alert_cache.go:276-312`), which deletes it and spawns the resolved-path
+  goroutines instead of the one under test.
+- Close with `cache.backendWG.Wait()` (§3) as the last statement. That is the
+  whole wait — no ticker, no sleep, no `len(cache.backendSem)` poll.
 - Run: `go test -race -run TestAlertCache_ConcurrentFireCaptureAndRefresh ./internal/webui/services/...`.
   Expected: fails with `WARNING: DATA RACE` on `Status`/`Annotations` against
   today's code; passes cleanly once `refreshAlerts` snapshots before
-  spawning.
+  spawning. Verify the *failing* direction repeats — apply §3 alone, keep
+  §1 unapplied and run `-count=10`: all ten runs must report the race. A
+  recipe that detects it 9 times out of 10 is not a regression test.
 
 ### Files touched
 
 - `internal/webui/services/alert_cache.go` — snapshot before the
-  fire-capture `runBounded` call; add the safety comment on the
+  fire-capture `runBounded` call; `backendWG` field plus the two `runBounded`
+  lines that make completion observable; add the safety comment on the
   resolved-statistics goroutine.
 - `internal/webui/services/alert_cache_test.go` — new
   `TestAlertCache_ConcurrentFireCaptureAndRefresh`.
@@ -198,6 +259,13 @@ Work with the real type instead of adding one:
   (real TCP dial attempt, real JSON/proto marshalling) but exercises the
   exact production code path — a fake would risk not reproducing the read
   order that causes the race in the first place.
+- **A `WaitGroup` in production code for a test's benefit**: `backendWG` is
+  two lines and is never waited on outside tests. The alternative — leaving
+  `runBounded` alone and inferring completion from `backendSem` — was tried
+  and is unsound at the spawn edge (§3), and there is no way to close that
+  gap from outside `runBounded`. The counter also leaves the door open for
+  `Stop()` to drain in-flight backend calls later, which it currently
+  doesn't.
 - **No fix to `UpdateAlertResolved`**: relies on the delete-under-the-same-lock
   argument holding. If a future change moves the `delete(ac.alerts, ...)`
   outside the lock, or adds another lookup path that doesn't key off


SoulKyu (commented):
Spec-only PR (no code touched) proposing the fix plan for #161. Verified the core analysis against current `main`: every cited line number, function name, and commit hash (`11ef6a9`, `b3d9002`) checks out exactly, and the proposed snapshot-before-spawn fix genuinely closes the race (the snapshot copy is never inserted into the cache map, so no concurrent writer can reach it). The `UpdateAlertResolved` safety argument (delete happens under the same lock, before the section's unlock) also holds up against every lookup path I checked.

One actionable, non-blocking gap in the spec's regression-test design (Approach, section 3, the "Before the test returns, wait for every spawned fire-capture goroutine to actually finish its read" bullet describing polling `cache.backendSem`):

**Problem:** `runBounded` spawns each goroutine with a bare `go func(){ ac.backendSem <- struct{}{}; ... }()` — the semaphore token is only acquired *inside* the goroutine, once the scheduler actually runs it. Right after the test's loop of refresh calls returns, none of the freshly spawned capture goroutines may have run yet, so the semaphore's length can legitimately read as empty — indistinguishable from "all goroutines already finished". If the poll loop starts checking immediately, it can observe this false-empty state on its very first tick and declare completion before any goroutine has performed its unsynchronized read. Since `-race` can only flag conflicting accesses it actually observes, the test could then pass cleanly on today's buggy code too — not because the race is fixed, but because the read never ran in time to be observed. That undermines the spec's own stated expectation that the test fails against today's code and passes once the fix lands.

**Suggested fix:** don't treat a single "semaphore length is zero" read as sufficient completion evidence. Either (a) require the poll to first observe the semaphore count rise above zero (proving goroutines have actually started) before treating a subsequent zero as "drained", or (b) debounce the drained state — only accept zero as final once it has held across several consecutive poll ticks spaced a few milliseconds apart. Please add this detail to the test-plan bullet so the implementer doesn't ship a test that silently never exercises the race.



SoulKyu:
## 🧪 QA Report

Spec-only PR (`looper/planner/*`, diff = `specs/2026-08-01-161-fix-webui-the-capturealertfired.md` only). Per QA policy the linked issue's acceptance criteria are **not** graded — the fix is intentionally not implemented here. The **spec itself** is graded: is every codebase reference accurate at `14c6cd2`, and would following it as written misdirect the implementer?

App booted via `make test` (docker-compose: webui `:8081`, backend `:8080`/`:50051`, postgres, alertmanager) — all four containers running, `GET /login` → 200, alertmanager `/api/v2/status` → 200.

|---|---|---|
| Diagnosis is a real race, at the claimed locations | ✅ | Implemented the spec's test recipe verbatim, ran `go test -race` on unfixed `14c6cd2`: `DATA RACE` — read at `backend_client.go:1812` (`alert.Annotations`) and `:1836` (`alert.Status`) from `refreshAlerts.func1` (`alert_cache.go:251`) vs. previous write in `updateExistingAlert` at `:419` / `:414`. Exactly the two field pairs the spec names. |
| Every `alert_cache.go` line reference accurate | ✅ | `239-255` `!exists` branch ✓, `:240` map store ✓, `:245` SSE `alertCopy` ✓, `:249` `runBounded` ✓, `:251` `CaptureAlertFired(dashAlert)` ✓, `:55` `maxBackendWorkers = 8` ✓, `198-204` `runBounded` ✓, `:222` `mu.Lock` / `:315` `Unlock` ✓, `292-298` resolved goroutine ✓, `300-304` copy + comment ✓, `:306` `delete` ✓, `:414`/`:419` `Status`/`Annotations` writes ✓, `695-698` shallow-copy convention comment ✓. |
| Every `backend_client.go` line reference accurate | ✅ | `:1804` `3*time.Second` timeout ✓, `1812-1813` `alert.Annotations` ✓, `:1836` `alert.Status.State == "silenced" \|\| len(SilencedBy) > 0` ✓ — and all three reads happen **before** the RPC send at `:1839`, as the spec asserts. |
| Supporting facts (types, commits, defaults) accurate | ✅ | `AlertStatus{State string; SilencedBy, InhibitedBy []string}` at `dashboard.go:58-62` ✓. `11ef6a9` = "stop alert cache from handing out live alert pointers" ✓, `b3d9002` = "…copy alert for bg goroutine…" ✓. "default refresh interval 10s" ✓ — `config.go:206` and the **live container log**: `Alert cache initialized with sync interval: 10s`. `TestAlertCache_ConcurrentRefreshAndReads` exists (`alert_cache_test.go:1209`) ✓. |
| Proposed fix is correct and complete | ✅ | Applied the spec's snippet verbatim (`firedCopy := *dashAlert`, inside the `ac.mu` section). `go build ./...` ✓, `go vet ./...` ✓, spec-recipe test `-race -count=5` → **0 races, 5/5 clean**. Reverted afterwards; branch untouched. |
| Test recipe is followable and reproduces the bug | ✅ | Every seam the recipe depends on is real: `client.NewBackendClient(addr)` + `Connect()` uses `grpc.NewClient` (lazy dial) and **`IsConnected()` returns true against a closed loopback port** — verified live, RPC then fails fast with `connection refused`. `backendClient` is a concrete `*client.BackendClient` (`:68`) and the test package is `services`, so direct field assignment works ✓. `fakeAlertFetcher` seam exists to drive `refreshAlerts` ✓. Recipe detects the race **30/30 runs** on unfixed code. |
| `UpdateAlertResolved` "safe without a copy" reasoning holds | ✅ | Verified independently: `delete(ac.alerts, fp)` at `:306` is inside the same `222→315` critical section; all four accessors the spec names (`MutateAlert:505`, `GetAllAlerts:691`, `GetAlert:807`, `GetLiveAlert:899`) key off `ac.alerts` and return snapshots; `storeResolvedAlertInBackend` gets its own copy at `:303`. Also checked what the spec didn't: no other `AlertCache` field retains a live `*DashboardAlert`, and `go.mod` is `go 1.23.0` so the `for … range ac.alerts` loop variable captured by the goroutine is per-iteration (would have been a real bug pre-1.22). |
| Validation gate in the spec is achievable | ✅ | At `14c6cd2` unmodified: `go build ./...` = 0, `go vet ./...` = 0, `go test -race ./internal/webui/...` = 0 races, all 3 packages `ok`. So "confirm no other race is newly surfaced" is a meaningful, reachable gate. |
| Spec describes a live path, not dead code | ✅ | On the running stack, new alerts keep arriving (`31 → 32 → 33 active alerts` across refresh cycles) and each takes the `!exists` branch. Posted `QAFireCapture` to alertmanager and confirmed the row landed in postgres: `select fingerprint, alert_name, silenced_at_fire, metadata->>'annotations'` → `QAFireCapture \| f \| {"summary": "qa fire capture probe"}`. Both racing fields are persisted by this exact path, and `silenced_at_fire` is only ever inserted (`statistics_query.go` only reads it in `WHERE` clauses) — so "permanently wrong" is accurate. |
| "Shallow copy is sufficient" non-goal justified | ✅ | Searched for in-place mutation of `Annotations`/`Labels` maps on cached alerts — the only two hits (`statistics_query.go:1390,1403`) are on a freshly-built backend response item, never on a webui cache-resident `DashboardAlert`. Invariant holds. |

**Adversarial pass:**

- **Diagnosis / fix** — Did not take the spec's word for the race. Wrote its test, ran it on unfixed code (`DATA RACE` at exactly the claimed lines), applied its fix, re-ran `-count=5` (clean), reverted. Tried to break the fix's sufficiency by hunting for a deep-copy requirement: grepped every in-place `Annotations[k]=`/`Labels[k]=`/`delete(…)` in `internal/` — none touch cached alerts, so the shallow copy is not hiding a second race.
- **Test-recipe seams (the part most likely to be fiction)** — Attacked each claimed seam rather than reading it: dialed a *closed* loopback port and asserted `Connect()` succeeds **and** `IsConnected()` is true (both hold — this is the load-bearing claim, and it's correct); confirmed the reads at `:1812`/`:1836` precede the wire send so a dead backend still exercises them; confirmed the unexported-field injection compiles from package `services`. Ran the recipe **30×** to hunt for flakiness — 30/30 detected.
- **Boundary / flakiness** — ⚠️ **One spec statement is factually wrong**: *"polling `len(cache.backendSem) == 0` on a short ticker is a reliable 'all goroutines done' signal"*. It is not a join — `runBounded` (`:199-201`) acquires the token **inside** the spawned goroutine, so the semaphore reads empty before the goroutine has started. Probed this directly (one alert, check immediately after `refreshAlerts`): **22/30 runs read `len(backendSem)==0` while the fire-capture goroutine was still pending**. This did **not** flip the criterion to ❌ because the recipe's 8-iteration loop overlaps earlier goroutines with later refreshes and still detected the race 30/30 — the wrong rationale doesn't produce a broken test. But it should be corrected before it becomes load-bearing (a `sync.WaitGroup`, or a comment that the loop, not the wait, is what guarantees the reproduction); a future reduction to 1 iteration would silently turn this test into a false pass.
- **Error / unavailable path** — Ran the whole recipe against an unreachable backend (the spec's own premise): errors are logged and swallowed, no panic, no hang, RPCs fail fast instead of eating the 3s timeout — matching what the spec predicts.
- **Persisted state** — Checked the real database rather than trusting the impact claim: `alert_statistics` rows are written by this path with both racing fields, and `silenced_at_fire` has no update path, so corruption is permanent as stated.
- **Adjacent flows** — Verified the spec's change wouldn't disturb neighbours: `go test -race ./internal/webui/...` is fully green both before and after the proposed fix, and the `UpdateAlertResolved` / `storeResolvedAlertInBackend` goroutines twelve lines away were re-derived from scratch rather than accepted (including the pre-1.22 loop-variable trap the spec never mentions — not an issue at `go 1.23.0`).

**Verdict: PASS**

SoulKyu:

SoulKyu (commented):
@SoulKyu Reviewed the spec (this is a rebase-only update vs. the prior review — the spec file content and the code it references are byte-identical, confirmed via `git diff` against the previously-reviewed SHA and `git log` on `alert_cache.go`/`backend_client.go` since then). Independently re-verified the diagnosis, the proposed snapshot fix, and the `UpdateAlertResolved` safety argument (delete happens inside the same `ac.mu` critical section, before unlock, so no later refresh cycle can reach that struct again) against current `main` — all hold up.

One non-blocking gap remains in the test-recipe design (spec section "3. Regression test", the bullet beginning "Before the test returns, wait for every spawned fire-capture goroutine..."):

**Problem:** the bullet proposes polling `len(cache.backendSem) == 0` as the "all fire-capture goroutines finished" signal. Reading `runBounded` directly in `internal/webui/services/alert_cache.go` (around line 197-203) shows the semaphore token is acquired inside the spawned goroutine body — the first statement of the `go func(){...}()` is the channel send — not by the caller before `go func()` returns. So immediately after the test's loop of refresh calls finishes, the semaphore length can legitimately read zero simply because none of the freshly spawned goroutines have been scheduled yet — a state indistinguishable from "all of them already finished."

**Why it matters:** if the poll loop's first tick fires before any goroutine has actually run, it observes this false-empty state and the test can return before the unsynchronized read (`CaptureAlertFired`'s access to `alert.Annotations`/`alert.Status`) ever executes. Since `-race` only flags conflicts it actually observes, such a run would pass cleanly against today's buggy code too — not because the race is fixed, but because the read never happened in time to be observed. That contradicts the spec's own stated expectation a few lines below ("Expected: fails with a data race warning against today's code").

**Suggested change:** don't treat a single "semaphore length is zero" read as sufficient completion evidence in that bullet. Either (a) require the poll to first observe the semaphore count rise above zero, proving goroutines have actually started, before accepting a later zero reading as "drained," or (b) debounce — only accept zero as final once it has held across several consecutive poll ticks a few milliseconds apart. Please fold this detail into the bullet so the implementer doesn't ship a regression test that can silently never exercise the race it's meant to catch.

Solid spec otherwise — the commit archaeology and the `UpdateAlertResolved` safety reasoning in particular are thorough. Nice work chasing this down to the exact commit that introduced the gap.



SoulKyu:
## 🧪 QA Report

Spec/docs-only PR (`looper/planner/*`, diff = 1 file under `specs/`). Per the spec-PR rule I did **not** grade #161's acceptance criteria — the feature is intentionally unimplemented here. I graded the **spec itself**: booted the stack and checked every codebase reference at `7666ebf`, then implemented the spec verbatim in a throwaway worktree to see whether following it produces working code.

|---|---|---|
| Bug is real, and is the one described | ✅ | `-race` reproduces 7 race blocks, all on the fire-capture path: read `backend_client.go:1812` (Annotations) & `:1836` (Status) from `alert_cache.go:251`, vs writes `alert_cache.go:419` / `:414` / `:288`. Exactly the pairs the spec names. |
| All `alert_cache.go` line refs accurate | ✅ | Verified each: `:55` (`maxBackendWorkers = 8`), `:198-204` (`runBounded`), `:239-255` incl. `:240`/`:245`/`:249`/`:251`, `:292-298`, `:302-304`, `:306` (delete), `:222`→`:315` (lock span), `:414`, `:419`, `:695-698`. All exact. |
| All `backend_client.go` line refs accurate | ✅ | `:1804` (3s timeout), `:1812-1813` (Annotations), `:1836` (`SilencedAtFire`). Note the spec **corrected** the issue's stale numbers (1788/1811/1779) — it is right, the issue is not. |
| Type/commit/config claims accurate | ✅ | `AlertStatus{State string; SilencedBy, InhibitedBy []string}` ✔; `11ef6a9` title + `Closes #54` + it did add `TestAlertCache_ConcurrentRefreshAndReads` ✔; `b3d9002` = "copy alert for bg goroutine" ✔; default refresh 10s (`alert_cache.go:112`, live logs show 10s cycles) ✔. |
| Proposed fix actually fixes it | ✅ | Applied `firedCopy := *dashAlert` verbatim: all 7 races gone, 5/5 clean `-race` runs. `go build ./...` ✔, `go vet ./...` ✔, `go test -race ./internal/webui/...` → 98 pass. |
| Proposed regression test works as promised | ✅ | Built it verbatim from the spec's recipe: **20/20** runs report `WARNING: DATA RACE` against unfixed code, **5/5** pass after the fix. The lazy-dial trick is real — `Connect()` to a dead port succeeds and `IsConnected()` returns true (`grpc.NewClient`, `backend_client.go:59-75`). |
| `UpdateAlertResolved` "safe without a copy" reasoning | ✅ | Holds. `delete` at `:306` is inside the same critical section (`:222`→`:315`); `GetAllAlerts:701`, `GetAlert:813`, `GetLiveAlert:910` all return copies, so no path resolves to that pointer post-delete. No race block in any run had `UpdateAlertResolved` as the reader. |
| Shallow copy is sufficient (spec's non-goal) | ✅ | No in-place `Annotations[k]=` / `Labels[k]=` anywhere in `internal/webui`; only wholesale replacement (`:419`, `:488`). The only in-place map writes live in `internal/backend/services/statistics_query.go:1390,1403` — different package, own structs. |
| Spec's stated goroutine-wait justification | ⚠️ | **Inaccurate** — see below. Does not break the prescribed test, but the text should be corrected. |

### ⚠️ One inaccurate claim to fix in the spec text

Approach §3 states that polling `len(cache.backendSem) == 0` is *"a reliable 'all goroutines done' signal"*. It is not. `runBounded` (`:198-204`) spawns the goroutine **and then** acquires the token *inside* it, so the semaphore is still empty until the new goroutine is scheduled. Measured directly (200 trials, one fire-capture goroutine each): **162/200 (81%)** false all-clears when the condition is evaluated right after `refreshAlerts` returns.

This did **not** break the test the spec prescribes — the 8-iteration loop triggers the race well before the wait is reached (20/20 detections) — so it is a wrong rationale inside a procedure that works, not a defect that misdirects the implementation. But an implementer who trims the loop to a single alert pair, or who trusts this as the sole synchronization, gets a test that exits early and **silently passes against buggy code**. Suggest a `sync.WaitGroup` seam, or at minimum wait for the semaphore to go non-empty first.

**Adversarial pass:**
- **Does the fix actually work / is the diagnosis right?** Didn't take the spec's word — checked out `7666ebf` in a scratch worktree, wrote the spec's test verbatim, ran `-race` against unfixed code (races on the exact fields/lines claimed), applied the one-line fix, re-ran (clean, 5/5). Both directions confirmed.
- **Test-recipe claims (highest risk in any spec):** attacked the two load-bearing ones. Lazy-dial/`IsConnected()`-true → **held** (asserted in the test, passed). `len(backendSem)==0` as done-signal → **broke it**, 162/200 false all-clears; reported above.
- **Shallow-copy sufficiency:** grepped every in-place map mutation of `Labels`/`Annotations` repo-wide to try to invalidate the "replace wholesale only" invariant the no-deep-copy non-goal rests on. Only hits are in a different package on unrelated structs → invariant holds. Also found a writer the spec doesn't cite (`UpdateAlert`, `:488`) — same fix covers it, so this strengthens rather than breaks the spec.
- **`UpdateAlertResolved` safety argument:** tried to find a surviving reference to the deleted struct — checked every accessor (`GetAllAlerts`, `GetAlert`, `GetAlertByFingerprint`, `GetLiveAlert`, `MutateAlert`) for a handed-out live pointer; all copy under the lock. Also confirmed empirically across 20+ `-race` runs that no race ever names it as the reader.
- **Stale line numbers (the classic spec rot):** re-derived every cited line independently rather than trusting the issue — which is how I caught that the *issue's* `backend_client.go` numbers are stale and the spec's are correct. The race detector's own stack traces (`:1812`, `:1836`, `:251`, `:419`, `:414`) independently corroborate the spec's numbering.
- **Boot/adjacent:** full `make test` stack (postgres/alertmanager/backend/webui all healthy), webui serves `200`, and live logs show `refreshAlerts` cycling every 10s over 26–27 alerts — the exact code path under discussion is live at this SHA. Full package suite `go test -race ./internal/webui/...` = 98 pass, no unrelated breakage.
- Env note: `templ` was absent and had to be installed pinned to go.mod's `v0.3.906` (a newer version emits `templ.ResolveAttributeValue`, which won't compile against the pinned lib). Not a PR defect.

**Verdict: PASS**

SoulKyu:

SoulKyu (commented):
@SoulKyu Reviewed this rebase of the spec for #161 at `d5b32000`. The diagnosis, the snapshot fix, and the `UpdateAlertResolved` safety argument all still hold (re-verified independently against `main`, matching the two prior QA passes' PASS verdicts on earlier SHAs of this same content).

One non-blocking gap carries over unchanged from the prior reviews on this PR — inline anchoring isn't usable on the spec document added by this PR (its filename trips the review tool's content-safety gate), so noting it here instead, located in the "Approach" section's "3. Regression test" subsection, the bullet beginning "Before the test returns, wait for every spawned fire-capture goroutine...":

**`len(cache.backendSem) == 0` is not a reliable "goroutines done" signal for the new test.** That bullet tells the implementer to wait for spawned fire-capture goroutines by polling the semaphore length on a short ticker, calling it "a reliable 'all goroutines done' signal."

**Problem:** `runBounded` (`internal/webui/services/alert_cache.go:198-204`) spawns `go func(){ ac.backendSem <- struct{}{}; defer func(){ <-ac.backendSem }(); fn() }()`. The semaphore token is acquired *inside* the goroutine body, not by the caller before `go func()` returns. Immediately after the test's refresh-call loop returns, none of the freshly spawned fire-capture goroutines may have been scheduled yet, so the semaphore length can legitimately read zero — indistinguishable from "all goroutines already finished."

**Why it matters:** if the poll's first tick fires before any goroutine has run, it observes this false-empty state and the test can return before the fire-capture RPC call's unsynchronized read of the alert's Annotations/Status fields ever executes. Since the race detector only flags conflicts it actually observes, such a run would pass cleanly on today's buggy code too — not because the race is fixed, but because the read never happened in time to be observed. That contradicts the spec's own stated expectation a few lines below that the test fails against today's code.

This exact gap was already flagged on this PR's two prior head SHAs (independently measured false-all-clear rates of 73-81% when checked immediately after refresh in those reviews), and the spec text is unchanged across all three SHAs (rebase only) — it hasn't been corrected yet.

**Suggested fix:** don't treat a single semaphore-zero read as sufficient. Either (a) require the poll to first observe the semaphore count rise above zero before accepting a later zero as "drained", (b) add a WaitGroup seam the test can join on instead of polling the semaphore, or (c) debounce — only accept zero as final once it holds across several consecutive poll ticks a few ms apart. Fold whichever into that bullet so an implementer who trims the loop or relies solely on this signal doesn't ship a test that silently never exercises the race.



SoulKyu:
## 🧪 QA Report

Spec/docs-only PR (`looper/planner/*`, diff = 1 file under `specs/`, +212). Per the spec-PR rule I did **not** grade #161's acceptance criteria — the feature is intentionally unimplemented here. I graded the **spec itself** at the post-rebase SHA `d5b3200`: the rebase moved this spec onto four new `main` commits, so the primary risk this round is **line-reference drift** — every cited line was re-derived from scratch, not carried over from the previous QA runs.

Stack booted (`make test` → 4 containers healthy): webui `:10081` `/login` → 200, backend `:10080` `/health` → 200, alertmanager `:11093` `/api/v2/status` → 200.

|---|---|---|
| Diagnosis is a real race, at the claimed locations, **at this SHA** | ✅ | Implemented the spec's §3 recipe verbatim, ran `-race` on unmodified `d5b3200`: **6 `DATA RACE` blocks**, read `backend_client.go:1812` (`Annotations`) and `:1836` (`Status`) from `refreshAlerts.func1` (`alert_cache.go:251`) vs. write `updateExistingAlert` `:419` / `:414`. Exactly the two field pairs the spec names. |
| All `alert_cache.go` line refs survive the rebase | ✅ | Re-dumped every cited line at `d5b3200`: `:55` `maxBackendWorkers = 8` ✓, `198-204` `runBounded` ✓, `:222` `mu.Lock` → `:315` `Unlock` ✓, `239-255` incl. `:240` map store / `:245` SSE `alertCopy` / `:249` `runBounded` / `:251` `CaptureAlertFired(dashAlert)` ✓, `292-298` resolved goroutine ✓, `302-304` copy+comment ✓, `:306` `delete` ✓, `:414`/`:419` ✓, `695-698` shallow-copy convention ✓. **Zero drift** — the four new `main` commits don't touch these files. |
| All `backend_client.go` line refs accurate | ✅ | `:1804` `3*time.Second` ✓, `1812-1813` `alert.Annotations` ✓, `:1836` `alert.Status.State == "silenced" \|\| len(SilencedBy) > 0` ✓ — and all reads precede the wire send at `:1839`, as the spec asserts (that's what makes the dead-backend test work). |
| Type / commit / config claims accurate | ✅ | `AlertStatus{State string; SilencedBy, InhibitedBy []string}` at `models/dashboard.go:58-62` ✓. `11ef6a9` = "stop alert cache from handing out live alert pointers" ✓, `b3d9002` = "…copy alert for bg goroutine…" ✓. 10s default confirmed twice — `alert_cache.go:112` and `config/config.go:206` — and in the **live container log** (refresh cycles at `10:03:25 → 10:03:35`). `TestAlertCache_ConcurrentRefreshAndReads` exists (`alert_cache_test.go:1209`) ✓. |
| Proposed fix is correct and complete | ✅ | Applied `firedCopy := *dashAlert` verbatim in a scratch worktree: all 6 races gone, `-count=5` → 5/5 clean. `go build ./...` ✓, `go vet ./...` ✓, `go test -race ./internal/webui/...` → all 4 packages `ok`, 0 races. Reverted; **PR branch untouched**. |
| Test recipe is followable and reproduces the bug | ✅ | Every seam is real: `client.NewBackendClient(addr)` + `Connect()` to a **closed** loopback port succeeds *and* `IsConnected()` returns true (asserted in the test, passed) — the load-bearing lazy-dial claim holds. `backendClient` is a concrete `*client.BackendClient` (`:68`), test package is `services`, so direct field assignment compiles. `fakeAlertFetcher` seam exists. Recipe detects the race **30/30** independent runs on unfixed code. |
| `UpdateAlertResolved` "safe without a copy" reasoning holds | ✅ | Re-derived independently: `delete(ac.alerts, fp)` at `:306` sits inside the same `222→315` critical section; read `GetAllAlerts:700-704`, `GetAlert:811-816`, `GetLiveAlert:910-912` bodies — all `alertCopy := *alert` under the lock, so no path hands out the deleted struct; `storeResolvedAlertInBackend` takes its own copy at `:303`. Across 30+ `-race` runs, no race block ever named `UpdateAlertResolved` as reader. |
| "Shallow copy is sufficient" non-goal justified | ✅ | Hunted for a deep-copy requirement: only in-place `Labels[k]=`/`Annotations[k]=` writes in the repo are `alert_cache.go:344`+`handlers.go:290` (fresh `transformedLabels` map, not a cached alert) and `backend/services/statistics_query.go:1390,1403` (different package, backend response items). `updateExistingAlert` replaces `Annotations` wholesale (`:419`). Invariant holds. |
| Impact claim ("permanently wrong `silenced_at_fire`") is accurate | ✅ | Verified against the live DB, not the prose: posted `QA162FireCapture` to alertmanager, row landed in `alert_statistics` with **both** racing fields — `metadata->>'annotations'` = `{"summary": "qa162 fire capture probe"}`, `silenced_at_fire` = `f`. `SilencedAtFire` is written only at insert (`statistics_grpc_service.go:395`) and only ever read in `WHERE` clauses (`statistics_query.go:122,1473,1561`) — no update path, so "permanent" is correct, and a corrupted value silently skews stats filters. |
| Validation gate is achievable | ✅ | On pristine `d5b3200`: `go build ./...` = 0, `go vet ./...` = 0, `go test -race ./internal/webui/...` = 0 races. So "confirm no other race is newly surfaced" is a meaningful, reachable gate. |
| Spec's semaphore-join rationale | ⚠️ | **Factually wrong** — see below. Does not break the prescribed procedure, so no criterion flips. |

### ⚠️ One inaccurate statement, unchanged across three QA passes

Approach §3: *"polling `len(cache.backendSem) == 0` on a short ticker is a reliable 'all goroutines done' signal"*. It is not a join. `runBounded` (`:199-201`) spawns the goroutine and acquires the token **inside** it, so the semaphore reads empty before the goroutine has started.

Measured directly at this SHA (one fire-capture goroutine per trial, sampled right after `refreshAlerts` returns): **185/200 (92.5%) false all-clears**.

It does **not** flip the criterion, because the recipe's 8-iteration loop triggers and detects the race long before the wait is reached — 30/30 detections. But I quantified the downside: trimming the loop to **1 iteration** drops detection to **29/30**, i.e. the broken join does let a buggy build pass silently. Wrong rationale inside a procedure that works — the implementer should replace it with a `sync.WaitGroup` seam (or state plainly that the loop, not the wait, guarantees reproduction). Flagged in the two previous QA rounds and still present in the text.

**Adversarial pass:**

- **Rebase drift (the new risk this round, and the reason not to trust prior QA):** re-dumped all 17 `alert_cache.go` line citations and all 4 `backend_client.go` ones at `d5b3200` rather than diffing against the old report; then checked `git log -- <both files>` to confirm none of the four new `main` commits (`79bbe3d`, `225749a`, `21b9b86`, `05bf64f`) touch them. No drift.
- **Did new `main` code add a reader the one-line fix would miss?** Re-read `refreshAlerts` `206-320` looking for other goroutines closing over a cache-resident pointer — the only three are the fire-capture one (the bug), `UpdateAlertResolved` (argued safe, verified), and `storeResolvedAlertInBackend` (already copies at `:303`). No new hazard.
- **Is the spec's writer list complete?** Tried to break it: enumerated every `MutateAlert` caller. Found one the spec never names — `dashboard_handlers.go:1019-1024` writes `a.Status.State = "resolved"` on the cached struct from a user-triggered resolve, a **second** concurrent writer of a field `CaptureAlertFired` reads at `:1836`. This strengthens rather than breaks the spec: the snapshot fix is writer-agnostic and covers it. Also cross-checked every field `CaptureAlertFired` reads against every field `updateExistingAlert` writes — `Status` and `Annotations` are the complete intersection, so no racing field is missing from the diagnosis.
- **Diagnosis & fix, both directions:** did not take the spec's word — reproduced on unfixed code (races at the exact claimed lines), applied the fix, re-ran `-count=5` clean, ran the whole `internal/webui/...` race suite clean, reverted.
- **Test-recipe seams (highest-risk part of any spec):** attacked the lazy-dial claim directly (`Connect()` + `IsConnected()` asserted against a *closed* port → both hold); confirmed the reads precede the RPC send so a dead backend still exercises them; confirmed the unexported-field injection compiles from package `services`; ran the recipe 30× hunting flakiness (30/30). Broke the semaphore-join claim (185/200) and quantified its blast radius (29/30 at 1 iteration).
- **Error / unavailable path:** ran the whole recipe against an unreachable backend — RPCs fail fast with `connection refused`, errors logged and swallowed, no panic, no hang, no 3s-timeout stall. Matches what the spec predicts.
- **Persisted state:** checked the real Postgres instead of trusting the impact prose — the fire-capture path does write both racing fields, and `silenced_at_fire` has no update path, so corruption is permanent as claimed.
- **Live path, not dead code:** live webui logs show `refreshAlerts` cycling every 10s over 21–22 alerts and taking the `!exists` branch for new ones; my probe alert traversed the exact path end-to-end into `alert_statistics`.
- Env notes (not PR defects): `templ` was absent and had to be installed pinned to `go.mod`'s `v0.3.906`; and `docker-compose.yml` hard-codes `container_name:` values that collide with a parallel QA slot (`qa158`) — suffixed locally, working tree only, never committed.

**Verdict: PASS**

SoulKyu:

SoulKyu (commented):
@SoulKyu Spec-only PR (single new file under `specs/`), content byte-identical to the three prior rebases on this same PR — re-verified independently against current `main` (tip `9b8db47`). The diagnosis is a genuine race (confirmed `runBounded`/`refreshAlerts` in `internal/webui/services/alert_cache.go` still match every cited line), the proposed snapshot fix closes it without a deep copy, and the `UpdateAlertResolved` safety argument (delete happens inside the same `ac.mu` critical section, before unlock) holds up against every lookup path.

Inline anchoring isn't usable on the spec document added by this PR (its filename trips the review tool's content-safety gate), so noting the finding here instead — same issue prior QA/review rounds hit on earlier SHAs of this identical text.

**Location:** the new spec file, section "Approach" -> "3. Regression test", the bullet beginning "Before the test returns, wait for every spawned fire-capture goroutine..." (ends at "...reliable 'all goroutines done' signal.").

**Problem:** that bullet tells the implementer to wait for spawned fire-capture goroutines to finish by polling `len(cache.backendSem) == 0` on a short ticker, calling it "a reliable 'all goroutines done' signal." But `runBounded` (`internal/webui/services/alert_cache.go:198-204`, unchanged on current `main`) spawns the goroutine first and only acquires the semaphore token inside it: `go func() { ac.backendSem <- struct{}{}; defer func() { <-ac.backendSem }(); fn() }()`. The token is acquired by the spawned goroutine, not by the caller before `go func()` returns. Immediately after the test's loop of refresh calls finishes, none of the freshly spawned fire-capture goroutines may have been scheduled yet, so `len(cache.backendSem)` can legitimately read `0` — a state indistinguishable from "all goroutines already finished."

**Why it matters:** if the poll loop's first tick fires before any goroutine has actually run, it observes this false-empty state and the test can return before `CaptureAlertFired`'s unsynchronized read of `alert.Annotations`/`alert.Status` ever executes. Since the race detector only flags conflicts it actually observes, such a run would pass cleanly against today's buggy (unfixed) code too — not because the race is fixed, but because the read never happened in time to be observed. That contradicts the spec's own stated expectation two lines below ("Expected: fails with a data race warning ... against today's code"). An implementer who trims the prescribed 8-iteration loop, or who relies solely on this bullet's join logic, can ship a regression test that silently never exercises the race it's meant to catch.

**Suggested change:** don't treat a single semaphore-zero read as sufficient completion evidence. Either (a) require the poll to first observe the semaphore count rise above zero, proving goroutines have actually started, before accepting a later zero as "drained"; (b) add a WaitGroup seam the test can join on instead of polling the semaphore at all; or (c) debounce, only accepting zero as final once it has held across several consecutive poll ticks a few milliseconds apart. Please fold one of these into that bullet so the wait logic is actually load-bearing rather than incidentally masked by the loop's iteration count.

Solid spec otherwise, and this gap does not block merge since it's spec text, not shipped code — nice job chasing this race down to the exact commit that introduced the gap.



SoulKyu:
## 🧪 QA Report

Spec-only planner PR (`looper/planner/*`, diff touches only `specs/`). Per QA policy the linked issue's acceptance criteria are **not** graded — the fix is intentionally not implemented here. What is graded is the **spec itself**: is every codebase reference accurate at this SHA, and would following it as written misdirect the implementer?

App booted from this SHA (webui `:10081`, backend `:10080`, alertmanager `:11093`, all `200`); `Alert cache initialized with sync interval: 10s`, refresh cycling every 10s over 20 active alerts.

|---|---|---|
| Problem is real, not a false positive | ✅ | Reproduced under `-race`: stacks land exactly on the cited lines — read `backend_client.go:1812` (Annotations) from `alert_cache.go:251` via `runBounded:202`, vs. previous write `updateExistingAlert` `alert_cache.go:419`. Also `:1836`/`:1825` reads vs `:414` write. |
| Impact claim (corrupted `silenced_at_fire`) is real | ✅ | Live app: first refresh after boot took the `!exists` branch for all 20 alerts → `select count(*) from alert_statistics` = 21 rows, 2 with `silenced_at_fire=true`. The racing read feeds a permanently-recorded column, as the spec states. |
| All codebase references accurate at this SHA | ✅ | Checked every one: `alert_cache.go` :239-255, :240, :245, :249, :251, :292-298, :302-304, :306, lock :222→:315, :414, :419, :55 (`maxBackendWorkers = 8`), :198-204, :695-698; `backend_client.go` :1804 (3s timeout), :1812-1813, :1836; `AlertStatus{State; SilencedBy; InhibitedBy}`; `backendSem` cap = `maxBackendWorkers`; default 10s (`config/config.go:206`, confirmed in live boot log); commits `11ef6a9` / `b3d9002` exist with the claimed subjects. No drift. |
| Proposed fix is correct and sufficient | ✅ | Applied §1 verbatim (`firedCopy := *dashAlert`): race gone. `go test -race -count=3` on the new test clean, `go test -race ./internal/webui/...` clean (4 pkgs), `go build ./...` and `go vet ./...` exit 0. |
| "Shallow copy is sufficient" (no deep copy) | ✅ | Tried to break it: swept every `.Annotations[` / `.Labels[` write in `internal/`. All webui-path uses are reads; the only map writes (`statistics_query.go:1390,1403`) build a fresh result map in the backend package, never a cache-resident `DashboardAlert`. Invariant holds, so the shared map in the shallow copy is safe. |
| `UpdateAlertResolved` "safe without a copy" argument | ✅ | Attacked each leg: `GetLiveAlert` (:900) does exist and **copies under `RLock`** (:911), so it can't leak the pointer; every other accessor keys off `ac.alerts`, which no longer holds the struct after `delete` at :306 (inside the same :222→:315 critical section); `go 1.23.0` (go.mod) ⇒ per-iteration loop vars, so closing over `alert` in the range loop is not itself a bug. Argument is sound. |
| Regression-test recipe works as specified (§3) | ❌ | **The drain-wait is unsound.** `runBounded` acquires the token *inside* the spawned goroutine (`alert_cache.go:199-201`), so `len(backendSem) == 0` also means "spawned but not yet scheduled" — not "done". The test can return before the racing read happens. Measured against the **buggy** code, 10 runs: spec's wait → **9/10** detect the race, run 10 exited `0` with 0 races. Replacing only the wait with a sound one → **10/10**. The spec calls this "a reliable 'all goroutines done' signal"; it is not. |
| Validation commands valid | ✅ | All three (`go build ./...`, `go vet ./...`, `go test -race ./internal/webui/...`) run clean with the spec's fix applied. |

**Adversarial pass:**
- *Problem/fix*: not taken on faith — implemented §3's test verbatim and ran it against pristine PR code (races, on the cited lines), then applied §1 and re-ran (`-count=3` clean, then whole-package clean). Both directions of the spec's own claim exercised.
- *Boundary/flakiness*: ran the spec's recipe **10×** against unfixed code specifically to find a spurious pass — found one (run 10). Then isolated the cause by changing *only* the wait, 10× more → 10/10 fail. This is what flipped the last criterion to ❌; the exact trigger is the scheduler not having run the `runBounded` goroutine before the poll's first `len()` read.
- *Shallow-copy invariant*: hunted for any in-place `Annotations`/`Labels` mutation that would defeat the copy repo-wide; none on the cache path.
- *Resolved-path reasoning*: tried to find a writer that survives the `delete` — checked `GetLiveAlert`, `GetAlert`, `GetAllAlerts`, `MutateAlert`, `UpdateAlert` (:492), and the loop-variable capture semantics. Could not break it.
- *Reference drift*: re-derived all ~20 line numbers against this SHA rather than trusting the issue text (the issue's own numbers — `:1788`/`:1811`/`:1779` — are stale; the spec's are correct).
- *Live path*: confirmed via postgres that the racing read actually reaches a persisted column, so this isn't a theoretical hazard.

**One fix needed:** §3's "wait for goroutines" step. Any of — a brief settle before polling, draining by acquiring all `maxBackendWorkers` tokens, or a `WaitGroup` — restores 10/10 detection. Everything else in the spec is accurate and verified working; this is a small correction to an otherwise unusually precise spec.

*(Env note: unrelated to the PR — `make test` tags images `:latest` while the harness's compose pins `:qa162`, and the hardcoded `container_name`s collide across parallel QA slots; worked around locally, no commit made.)*

**Verdict: FAIL**

SoulKyu (commented):
@SoulKyu This revision fixes the non-blocking gap flagged on all four prior review rounds for this PR: §3 ("Give `runBounded` a completion signal the test can wait on") replaces the unsound `len(cache.backendSem) == 0` poll with a `backendWG sync.WaitGroup` counted via `Add(1)` on the caller's goroutine, before `go func()` spawns the worker. That closes the exact spawn-edge blind spot every prior round called out — the semaphore token was only ever taken *inside* the goroutine, so an immediate post-loop poll could read a false-empty state before any goroutine ran.

I re-verified the rest of the spec independently against current `main`:
- The core race diagnosis holds: `refreshAlerts` (`internal/webui/services/alert_cache.go:239-255`) hands `CaptureAlertFired` the cache-resident `dashAlert` pointer with no snapshot, while `updateExistingAlert` (`:414`, `:419`) rewrites `Status`/`Annotations` on that same struct under `ac.mu` on the next cycle. `CaptureAlertFired` (`internal/webui/client/backend_client.go:1799-1836`) reads both fields unlocked.
- The `UpdateAlertResolved` safety argument still checks out: `delete(ac.alerts, fingerprint)` happens inside the same critical section, before unlock, so no later lookup path can reach that struct again.
- The new `backendWG` design is sound: all three `runBounded` call sites (`:249`, `:292`, `:304`) execute while `refreshAlerts` holds `ac.mu` on a single ticker-driven goroutine, so `Add`/`Wait` are never called concurrently from zero, and `Start()` isn't invoked by the test recipe — matching the spec's claim that production behaviour is unchanged and the WaitGroup only adds a happens-before edge from goroutine to waiter, not synchronizing away the race under test.
- The `grpc.NewClient` lazy-dial claim behind the test recipe (§4) matches `BackendClient.Connect()` (`backend_client.go:59-60`).

Nice work chasing this through to a genuinely deterministic regression-test design — solid spec, ready to build from.



SoulKyu:
## 🧪 QA Report

Spec-only planner PR (`looper/planner/*`, diff touches only `specs/`). Per QA policy the linked issue's acceptance criteria are **not** graded — the fix is intentionally not implemented here. What is graded is the **spec itself**: is every codebase reference accurate at this SHA, and would following it as written misdirect the implementer?

This head is the re-spin of the one thing the previous round failed (`head=4ec1972`: the `len(backendSem) == 0` drain-wait was unsound at the spawn edge, 9/10 detection). The new §3 replaces it with a `backendWG sync.WaitGroup` incremented on the caller's goroutine before `go`.

App booted from this SHA (webui `:9081`, backend `:9080`, alertmanager `:10093` — all `200`); refresh cycling every 10s, active alert count moving `21 → 22` (i.e. the `!exists` branch under test is live).

|---|---|---|
| Problem is real, not a false positive | ✅ | Implemented the spec verbatim in a scratch clone; `-race` lands on exactly the cited lines — read `backend_client.go:1836` (`Status`) from `alert_cache.go:254` via `runBounded`, vs. previous write `updateExistingAlert` at the line the spec cites as `:414` (`:417` after §3's +3 lines). Annotations pair (`:1812` vs `:419`) reported too. |
| Impact claim (corrupted `silenced_at_fire`) is real | ✅ | Live postgres: `select count(*), count(*) filter (where silenced_at_fire) from alert_statistics` → `21 / 1`. The unsynchronised read feeds a permanently persisted column, as the spec states. |
| **§3 gives a sound completion signal (the previous FAIL)** | ✅ | Applied §3 alone, kept §1 **unapplied**, 10 separate `-race` runs: **10/10** report the race (2–7 reports each). Then re-ran 10× at `GOMAXPROCS=1`, the configuration that most easily leaves a spawned goroutine unscheduled: **10/10** again. 20/20 total — deterministic, as §3 claims. The WaitGroup edge does not synchronise the race away. |
| Proposed fix (§1) is correct and sufficient | ✅ | Added `firedCopy := *dashAlert`: 3 × `go test -race -count=3` → clean, 0 races. `go test -race ./internal/webui/...` → **162 tests / 11 packages pass, no race**. `go build ./...` and `go vet ./...` exit 0. |
| §3 is safe in production as claimed | ✅ | All three `runBounded` call sites (`:249`, `:292`, `:304`) are inside the single `ac.mu` section `:222→:315`, so `Add` is serialised; nothing calls `Wait()` outside the test; `go vet ./...` (copylocks) clean; the whole suite, which drives `refreshAlerts` from many tests, passes with no `WaitGroup` misuse panic. |
| §4 recipe is buildable as written | ✅ | Followed it literally, no improvisation: freed-ephemeral-port `*client.BackendClient` — `Connect()` succeeds and `IsConnected()` is true (lazy `grpc.NewClient`, `backend_client.go:59-60` ✓), RPC fails fast with `connection refused` and is only logged; test wall time 0.05s, no 3s stall. |
| §4 "grow the fake's payload" warning is accurate | ✅ | Attacked it by doing the opposite — one alert per fetch: 7 `marking alert … as resolved` lines per run, i.e. exactly the resolved-path detour the spec warns about (`alert_cache.go:276-312`). Warning is real, not decoration. |
| "Shallow copy is sufficient" (no deep copy) | ✅ | Swept the repo for in-place map writes (`.Annotations[…]=` / `.Labels[…]=`): only 2 hits, both `internal/backend/services/statistics_query.go:1390,1403`, both writing freshly `make`-ed maps in the backend package, never a cache-resident `DashboardAlert`. Invariant holds. |
| §2 `UpdateAlertResolved` "safe without a copy" argument | ✅ | Attacked each leg: `GetLiveAlert` (`:899`) copies under `RLock` (`:908-911`) so it cannot leak the pointer; `delete(ac.alerts, …)` at `:306` is inside the same `:222→:315` critical section; every other accessor keys off `ac.alerts`. Could not construct a surviving writer. |
| All other codebase references accurate at this SHA | ✅ | Re-derived independently: `alert_cache.go` `:239-255`, `:240`, `:245`, `:249`, `:251`, `:276-312`, `:292-298`, `:303-304`, `:306`, `:414`/`:419`, `:55` (`maxBackendWorkers = 8`), `:71` (`backendSem`), `:198-204`, `:695-698`; `backend_client.go:1804` (3s timeout), `:1812-1813`, `:1836`, `:59-60`; `webuimodels.AlertStatus{State string; SilencedBy, InhibitedBy []string}` (`dashboard.go:57-61`) — the spec's package attribution is right; default 10s (`config/config.go:206`, confirmed by the live 10s cadence); commits `11ef6a9` / `b3d9002` exist with the claimed subjects. |
| Validation section runs clean | ✅ | `go build ./...` = 0, `go vet ./...` = 0, `go test -race ./internal/webui/...` = 0 with §1+§3+§4 applied. |

**Adversarial pass:**
- *Determinism (the previously-failing criterion)*: not taken on faith. 10 runs with §3-only at default `GOMAXPROCS`, then 10 more at `GOMAXPROCS=1` — the scheduler configuration designed to expose a spawned-but-not-scheduled goroutine. 20/20 detected. The spec's "all ten runs must report the race" bar is met with margin.
- *Both directions of the claim*: race reproduced against unfixed code, then §1 applied and re-run (`-count=3` ×3, then the full package) — clean. A spec proving only one direction would be graded on that.
- *Recipe traps*: checked the ways this test could hang or self-neuter — 3s gRPC timeout (fast-fails instead, 0.05s), `IsConnected()` returning false (it doesn't), the resolved-path detour (verified by doing the wrong thing on purpose), fingerprint drift when mutating `Status`/`Annotations` (labels unchanged → same fingerprint, `exists` branch taken as intended).
- *§3 as a production change, not just a test hook*: hunted for `WaitGroup` misuse (Add-concurrent-with-Wait from zero), a `Wait()` caller outside tests, and `copylocks` from the new field. None — all three `runBounded` sites `Add` under `ac.mu`, vet clean, full suite clean.
- *Shallow-copy invariant / resolved-path reasoning*: swept repo-wide for the mutation that would defeat the copy, and for a lookup path surviving the `delete`. Neither exists.
- *Reference drift*: re-derived every line number against this SHA rather than trusting the issue text (the issue's own `:1788`/`:1811`/`:1779` are stale; the spec's are correct).
- *Live path*: confirmed via postgres that the racing read reaches a persisted column, so this is not a theoretical hazard.
- *One claim I could not confirm (not a failure)*: §3's rationale states the old `len()` poll "misses the race roughly 1 run in 10". On this host I could not reproduce that — both a 1ms-ticker poll and an immediate busy-poll detected 10/10. The **structural** argument behind it is verifiably true (`alert_cache.go:199-201`: the token is taken inside the goroutine, so `len()` reads 0 for a spawned-not-scheduled call), and the spec no longer recommends that approach, so this misdirects nobody — a quantitative footnote, not guidance.
- *Nit, no impact*: §1 cites the existing resolved-path copy as `:302-304`; the copy itself is at `:303` (`:300-302` is its comment). §2 cites it correctly as `:303-304`.

Everything the spec asserts about the codebase is accurate at `a815d17`, and following it end-to-end produced a working, deterministic regression test plus a fix that clears `-race` across the whole `internal/webui/...` tree. The one defect from the previous round is genuinely closed.

*(Env notes, unrelated to the PR: `make test` tags images `:latest` while the harness's compose pins `:qa162`; the hardcoded `container_name`s collide across parallel QA slots; and `make test`'s `clean` step deletes the committed `_templ.go`/`output.css` and regenerates them with whatever `templ` is on `PATH` — v0.3.1020 here vs `go.mod` v0.3.906 — which breaks `go build ./...` until restored. All worked around locally; no commit made, working tree restored.)*

**Verdict: PASS**

SoulKyu:
Closing: this spec served its purpose (issue #161 is closed). The spec text remains on `looper/planner/161-fix-webui-the-capturealertfired`.

PR #168, opened 2026-08-02
fix(webui): snapshot alert before spawning CaptureAlertFired goroutine

### LINKED ISSUE #161: fix(webui): the CaptureAlertFired goroutine reads a cache-resident alert pointer — data race with the next refresh cycle

SoulKyu:
## Summary

`refreshAlerts` spawned the `CaptureAlertFired` statistics goroutine with the pointer it had just stored in `ac.alerts`, not a snapshot. The very next refresh cycle rewrites that same struct's `Status` and `Annotations` under `ac.mu` in `updateExistingAlert`, while the goroutine reads them with no lock held. `Status` is a multi-word struct, so the unsynchronized write/read pair can produce a torn read — best case a wrong `silenced_at_fire` permanently recorded in `alert_statistics`, worst case a crash.

This mirrors the fix already applied to the neighbouring resolved-alert goroutine (`b3d9002`), which just wasn't carried over to this call site.

## Changes

- `internal/webui/services/alert_cache.go`: snapshot `dashAlert` into `firedCopy` under `ac.mu` before spawning the `CaptureAlertFired` goroutine, and pass the copy instead of the cache-resident pointer.
- Checked the `UpdateAlertResolved` goroutine twelve lines down, as the issue asked: it doesn't need the same treatment because its fingerprint is deleted from `ac.alerts` within the same locked section, so no later refresh can reach that struct again. Added a comment recording that reasoning instead of a redundant copy.
- `internal/webui/services/alert_cache_test.go`: new `TestAlertCache_CaptureAlertFiredSnapshot`, which drives a refresh that inserts new alerts immediately followed by a refresh that mutates their `Status`/`Annotations`. Verified it fails under `go test -race` against the pre-fix code (caught the exact write/read pair at `alert_cache.go:419` / `backend_client.go:1836`) and passes after the fix.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test -race ./internal/webui/...` (162 tests pass)
- [x] Confirmed the new test fails with `-race` on the pre-fix code and passes after

Closes #161

### DIFF
--- a/internal/webui/services/alert_cache.go
+++ b/internal/webui/services/alert_cache.go
@@ -245,10 +245,14 @@ func (ac *AlertCache) refreshAlerts() {
 			alertCopy := *dashAlert
 			newAlertsForSSE = append(newAlertsForSSE, &alertCopy)
 
-			// Capture alert fired event for statistics
+			// Capture alert fired event for statistics. Snapshot the same way as
+			// the resolved path below: dashAlert now lives in ac.alerts and the
+			// next refresh's updateExistingAlert will rewrite its Status and
+			// Annotations under ac.mu while this goroutine reads them unlocked.
+			firedCopy := *dashAlert
 			ac.runBounded(func() {
 				if ac.backendClient != nil && ac.backendClient.IsConnected() {
-					if err := ac.backendClient.CaptureAlertFired(dashAlert); err != nil {
+					if err := ac.backendClient.CaptureAlertFired(&firedCopy); err != nil {
 						log.Printf("Failed to capture alert fired statistics for %s: %v", dashAlert.Fingerprint, err)
 					}
 				}
@@ -288,7 +292,10 @@ func (ac *AlertCache) refreshAlerts() {
 			alert.Status.State = "resolved"
 			alert.EndsAt = alert.ResolvedAt
 
-			// Update alert resolved event for statistics
+			// Update alert resolved event for statistics. Uses the cache-resident
+			// pointer, not a snapshot: fingerprint is deleted from ac.alerts below
+			// within this same locked section, so no later refresh can take the
+			// existing-alert branch and mutate this struct again.
 			ac.runBounded(func() {
 				if ac.backendClient != nil && ac.backendClient.IsConnected() {
 					if err := ac.backendClient.UpdateAlertResolved(alert); err != nil {
--- a/internal/webui/services/alert_cache_test.go
+++ b/internal/webui/services/alert_cache_test.go
@@ -2,6 +2,7 @@ package services
 
 import (
 	"errors"
+	"fmt"
 	"sync"
 	"testing"
 	"time"
@@ -11,6 +12,7 @@ import (
 	"notificator/internal/alertmanager"
 	alertpb "notificator/internal/backend/proto/alert"
 	"notificator/internal/models"
+	"notificator/internal/webui/client"
 	webuimodels "notificator/internal/webui/models"
 )
 
@@ -1443,4 +1445,53 @@ func TestAlertCache_AcknowledgmentDoesNotSurviveResolution(t *testing.T) {
 	if refired.AcknowledgedBy != "" || !refired.AcknowledgedAt.IsZero() || refired.AcknowledgeReason != "" {
 		t.Errorf("acknowledgment fields not cleared: %+v", refired)
 	}
+}
+
+// TestAlertCache_CaptureAlertFiredSnapshot reproduces #161: a refresh that
+// sees a brand-new alert used to hand the CaptureAlertFired goroutine the
+// pointer it had just stored in ac.alerts, rather than a snapshot. The very
+// next refresh cycle mutates that same struct's Status and Annotations under
+// ac.mu, in updateExistingAlert, while the goroutine reads them with no lock
+// held. Run with -race: this fails on the pre-fix code (dashAlert handed
+// straight to CaptureAlertFired) and passes once the goroutine gets its own
+// copy.
+func TestAlertCache_CaptureAlertFiredSnapshot(t *testing.T) {
+	// grpc.NewClient connects lazily, so this never dials: IsConnected() is
+	// true immediately, and CaptureAlertFired reads its alert argument before
+	// it ever touches the network.
+	backendClient := client.NewBackendClient("127.0.0.1:1")
+	if err := backendClient.Connect(); err != nil {
+		t.Fatalf("Connect: %v", err)
+	}
+
+	cache := NewAlertCache(nil, backendClient, 90, 10*time.Second)
+
+	var firing []alertmanager.AlertWithSource
+	for i := 0; i < 5; i++ {
+		firing = append(firing, alertmanager.AlertWithSource{
+			Alert: models.Alert{
+				Labels:      map[string]string{"alertname": fmt.Sprintf("Race%d", i)},
+				Annotations: map[string]string{"summary": "before"},
+				Status:      models.AlertStatus{State: "firing"},
+				StartsAt:    time.Now(),
+			},
+			Source: "prod",
+		})
+	}
+
+	cache.alertmanagerClient = &fakeAlertFetcher{alerts: firing}
+	cache.refreshAlerts() // inserts the alerts, spawns a CaptureAlertFired goroutine per alert
+
+	silenced := make([]alertmanager.AlertWithSource, len(firing))
+	for i, a := range firing {
+		silenced[i] = a
+		silenced[i].Alert.Annotations = map[string]string{"summary": "after"}
+		silenced[i].Alert.Status = models.AlertStatus{State: "suppressed", SilencedBy: []string{"sil-1"}}
+	}
+	cache.alertmanagerClient = &fakeAlertFetcher{alerts: silenced}
+	cache.refreshAlerts() // same fingerprints: rewrites Status/Annotations on the cached structs
+
+	// Give the CaptureAlertFired goroutines time to run their (unsynchronized,
+	// pre-fix) field reads before the test process exits.
+	time.Sleep(300 * time.Millisecond)
 }


SoulKyu (commented):
@SoulKyu thanks for the quick, well-scoped fix on #161.

Verified the core claim directly: reverted `alert_cache.go` to its pre-fix state (keeping the new test), ran `go test -race ./internal/webui/services/... -run TestAlertCache_CaptureAlertFiredSnapshot`, and reproduced the exact race — write at `alert_cache.go:414` (`updateExistingAlert`, `existing.Status = new.Status`) racing against the read at `backend_client.go:1836` (`alert.Status.State`/`SilencedBy` inside `CaptureAlertFired`), exactly as described. Restored the fix and the same test passes cleanly, as does the full `go test -race ./internal/webui/services/...` suite and `go build ./...`.

Checked the shallow-copy choice for `firedCopy := *dashAlert`: `Annotations` is a map field, but `updateExistingAlert` only ever replaces it wholesale (`existing.Annotations = new.Annotations`), never mutates the map in place, so the shallow copy is sufficient — no lingering aliasing race there.

Also checked the reasoning left in the comment for the neighbouring `UpdateAlertResolved` call site: the fingerprint is deleted from `ac.alerts` under the same lock before unlock, so a later refresh can only re-add that fingerprint through the "new alert" branch (a freshly converted, distinct pointer), never through `updateExistingAlert` against the same struct — so skipping the snapshot there is correct, not an oversight.

Minimal, correctly scoped, and backed by a test that actually proves the fix rather than just asserting no panic. Nice work.



SoulKyu:
## 🧪 QA Report

|---|---|---|
| `CaptureAlertFired` invoked with a snapshot taken under `ac.mu`, never the `ac.alerts` pointer | ✅ | `alert_cache.go:249` `firedCopy := *dashAlert` sits inside the `ac.mu.Lock()` section spanning `:222`–`:322`; `:251` passes `&firedCopy`. Live differential below proves the behavioural difference. |
| Test in `alert_cache_test.go` fails under `-race` pre-fix, passes post-fix | ✅ | Restored pre-fix `alert_cache.go` into a scratch copy: **5/5 independent runs FAIL** with `WARNING: DATA RACE` at exactly `alert_cache.go:414` (Status) / `:419` (Annotations) vs `backend_client.go:1812` / `:1836` — the precise pair the issue predicted. On this PR: **15/15 runs pass**. |
| `UpdateAlertResolved` snapshotted or carries a comment justifying safety | ✅ | Comment at `:292`–`:295`. Reasoning verified: `delete(ac.alerts, fingerprint)` at `:306` is in the same locked section, and a re-fire builds a *fresh* struct via `convertToDashboardAlert`, so no later writer can reach it. Exercised live (below) — 645 rows resolved correctly, 0 anomalies. |
| `go build ./...`, `go vet ./...`, `go test -race ./internal/webui/...` pass | ✅ | build clean, vet clean, **162 tests pass** under `-race`. |

### Live differential — the fix's actual effect on recorded statistics

Reproduced the issue's exact scenario against the running stack (300 new alerts ≫ the 8-worker
`runBounded` semaphore, backend paused so capture goroutines backlog, then the alerts silenced +
annotations rewritten while those goroutines were still queued, then backend resumed). I built a
**pre-fix webui image as a control** and ran the identical sequence:

| Build | rows | wrong `silenced_at_fire` | wrong (post-fire) annotations |
|---|---|---|---|
| **This PR** | 204 | **0** | **0** |
| Pre-fix control | 242 | **199** | **206** |

Every alert fired *unsilenced* with `summary=FIRE_TIME`. The pre-fix build permanently recorded 199
rows as silenced-at-fire and 206 rows with the post-fire annotation — the exact
uncorrectable `alert_statistics` corruption the issue describes. This PR records zero.

**Adversarial pass:**

- **Criterion 1 (snapshot).** Attacked the *shallow*-copy assumption: confirmed `Annotations`/`Labels` maps are replaced wholesale in `updateExistingAlert:414-419` and built fresh (`transformedLabels`) in `convertToDashboardAlert`, never mutated in place, and `Status.SilencedBy` slices likewise — so sharing the map/slice headers is safe. Probed the one pointer still dereferenced from the cache inside the goroutine, `dashAlert.Fingerprint` at `:252`: `Fingerprint` is write-once at construction (`:371`, no assignment anywhere in non-test webui code), so it is not a race — the clean 162-test `-race` suite agrees. Boundary: burst of 300 (≫ 8 workers), plus 0/1-alert refreshes via the existing suite.
- **Criterion 2 (test).** Checked for flakiness in both directions rather than a single run: 5/5 independent pre-fix runs caught it, 15/15 post-fix runs clean. Noted the test is assertion-free by design — it is purely a race-detector probe, so it is vacuous without `-race`; that matches the `11ef6a9` test the issue asked it to mirror.
- **Criterion 3 (resolved path).** Attacked it live instead of trusting the comment: restarted the fake Alertmanager so every cached fingerprint vanished at once, forcing a mass-resolve through `UpdateAlertResolved`. Result: 645 rows gained `resolved_at` + `mttr_seconds`, values sane (13s–658s), **0 rows with `resolved_at < fired_at`** — no torn/misattributed reads on the un-snapshotted pointer.
- **Error & unavailable path.** `docker pause` on the backend mid-flow: 120 `Failed to capture alert fired statistics … DeadlineExceeded` logged (failure is *visible*, not silently swallowed), webui stayed `healthy`, **0 restarts, 0 panics, 0 runtime race reports**, dashboard and SSE kept serving throughout. No wiped-state-as-success.
- **Malformed & hostile** (alert annotations treated as untrusted). 500 KB `summary` → stored intact; emoji/ANSI-escape/`<script>`/`' OR 1=1 --` → handled; empty `annotations{}` and empty `severity` → rows written correctly. A `\x00` NUL byte is rejected by PostgreSQL jsonb (`SQLSTATE 22P05`) — **pre-existing**, untouched by this diff, and it fails gracefully (logged, no crash, other alerts unaffected).
- **Auth & scope.** `/api/v1/dashboard/data` and `/api/v1/statistics/summary` → **401** unauthenticated. Session correctly invalidated after a webui restart (compose leaves `NOTIFICATOR_SESSION_SECRET` empty ⇒ random per boot).
- **Adjacent flows.** Dashboard HTML 200; authenticated `/dashboard/data` returned 50 alerts including the injected ones; SSE `/dashboard/stream` returned 200 `text/event-stream` with live `event:update` payloads.
- **Persisted state.** Rows survived a webui image swap and an Alertmanager restart; the resolved path correctly updated rows created *before* those restarts, not just fresh ones.

Two unrelated observations, out of scope for this PR and **not** counted against it: `NOTIFICATOR_POLLING_SYNC_INTERVAL` is bound via `viper.BindEnv` but ignored at load (cache still logs `10s`); and `docker-compose.yml` uses fixed `container_name:` values, which collide between parallel QA slots.

**Verdict: PASS**

Issue #45, opened 2026-07-24
fix(webui): a partial Alertmanager fetch mass-resolves every alert of the unreachable source and stampedes the backend


SoulKyu:
## Problem

`MultiClient.FetchAllAlerts` swallows per-source failures: it collects errors per client and only returns one when **every** source failed *and* zero alerts came back.

```go
// internal/alertmanager/client.go:678-697
for name, client := range mc.clients {
    alerts, err := client.FetchAlerts()
    if err != nil { errors = append(errors, ...); continue }   // :681-682
    ...
}
if len(errors) > 0 && len(allAlerts) == 0 { return nil, errors[0] }  // :693
return allAlerts, nil                                                // :697 <- partial result, err == nil
```

`AlertCache.refreshAlerts` then treats "fingerprint absent from this fetch" as "the alert resolved", with no guard on how much of the cache vanished at once:

```go
// internal/webui/services/alert_cache.go:200-229
for fingerprint, alert := range ac.alerts {
    if !currentFingerprints[fingerprint] {
        alert.IsResolved = true; alert.ResolvedAt = time.Now(); alert.Status.State = "resolved"
        go func(...) { ac.backendClient.UpdateAlertResolved(resolvedAlert) }(alert)   // :209-215
        go ac.storeResolvedAlertInBackend(&alertCopy)                                  // :221
        delete(ac.alerts, fingerprint)                                                 // :223
    }
}
```

**Concrete failure path** — two configured Alertmanagers, `prod` (3000 firing alerts) and `staging` (50):

1. `prod` is unreachable for a single 10 s poll (`Polling.SyncInterval` default, `config/config.go:191`) — restart, network blip, 503.
2. `FetchAllAlerts` returns the 50 `staging` alerts and `err == nil` (only one of two clients failed, and `len(allAlerts) != 0`).
3. `refreshAlerts` marks all 3000 `prod` alerts resolved: 3000 `UpdateAlertResolved` goroutines + 3000 `storeResolvedAlertInBackend` goroutines, each of which makes **three** more gRPC calls (`GetComments`, `GetAcknowledgments`, `CreateResolvedAlert`, `alert_cache.go:811-866`) — roughly 12 000 concurrent gRPC calls over a single connection with a 10 s deadline each (`internal/webui/client/backend_client.go`).
[...]

PR #131, opened 2026-07-27
feat(dashboard): keyboard-first triage — navigate, select and act on alerts without the mouse

SoulKyu:
## Summary
- Add a `DashboardKeyboardMixin` that layers keyboard triage over the existing alert table: `j`/`k` move a visible row cursor (scrolled into view), `Enter`/`o` open the detail modal, `x`/`Shift+X` toggle row/all selection through the same `selectedAlerts` state the checkboxes use, `a`/`s` acknowledge/silence the selection (or the cursor row if nothing is selected), `c` opens the detail modal on the Comments tab and focuses the input, `h` hides the cursor row, `Esc` clears the selection or defers to an open modal, `?` toggles a cheat-sheet overlay (`components/keyboard_shortcuts_help.templ`).
- All actions delegate to the existing mixin methods (`dashboard_actions.templ`, `showAlertDetails`) — no new action logic, so bulk-action error handling stays in one place, and nothing is removed: every action stays clickable.
- The modal-open / typing-target guard already used by `focusSearch()` is factored into a shared `shortcutsActive()` helper so the two checks can't drift.
- The row cursor is renormalized after every full reload and incremental SSE merge: it stays on its row if still present, resets to the first row otherwise.
- Grouped view (`viewMode === "group"`) has no flat row list to walk, so row shortcuts there surface an explicit action-error banner instead of silently no-op'ing.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./internal/webui/...`
- [x] `make webui-templates` regenerates the `_templ.go` files with no diff (already up to date)
- [ ] Manual keyboard-only pass in a browser: search, navigate with `j`/`k`, select two alerts with `x`, acknowledge with `a`, confirm `?` opens/closes the cheat sheet and `Esc` clears selection

Closes #111

<sub>🔁 Powered by <a href="https://github.com/nexu-io/looper">Looper</a> · runner=worker · agent=claude-code · An autonomous AI dev team for your GitHub repos.</sub>

PR #170, opened 2026-08-02
feat(dashboard): Alertmanager source freshness — live/stale/unreachable

### LINKED ISSUE #89: feat(dashboard): Alertmanager source freshness — show which sources are live, stale or unreachable
## User & need

The on-call engineer looking at a quiet dashboard at 3am, and the one debugging "why didn't we see that page?".

Notificator merges alerts from **every** configured Alertmanager into one cache
(`FetchAllAlertsDetailed`, `internal/alertmanager/client.go:674`). When one source fails, the
cache deliberately **keeps its last known alerts and only logs it**:

```
internal/webui/services/alert_cache.go:180
"Alert cache refresh: failed to fetch alerts from %s, keeping its cached alerts untouched: %v"
```

That behaviour is correct (it was the fix for the mass-resolve stampede, #45) — but it is
completely invisible from the browser:

- A source that has been unreachable for 20 minutes still shows its 20-minute-old alerts as if
  they were live. Nothing marks them stale.
- A source that was unreachable *since startup* contributes **zero** alerts. An empty dashboard
  looks exactly like "all quiet". At 3am that is the most expensive ambiguity in the product.
- The header's "Last updated" (`dashboard_core.templ:404`) is the **cache-wide** refresh time. It
  keeps ticking as long as *one* Alertmanager answers, so it actively reassures you while a source
  is down.
- Health data exists but only for probes: `GET /health/alertmanager` (`handlers.go:356`) calls
  `GetHealthStatus()`, which does a **fresh synchronous `TestConnection()` per instance**
  (`client.go:829`). No UI consumes it, and it says nothing about *poll* freshness.

Cost today: silent blindness. The team has no way, from the console they live in, to answer
"am I seeing everything right now?" — the answer is in the WebUI's stdout.

## Proposed feature

**A per-source freshness indicator in the dashboard header, plus a stale-source banner.**

From the user's side of the screen:

- A small **Sources** pill in the dashboard header: `3/3 live`, green. Degraded → amber
  `2/3 live`; all down → red. Clicking it opens a popover with one row per Alertmanager:
  name, state (`live` / `stale` / `unreachable`), **last successful poll** as a relative time
  ("8s ago", "14m ago"), the alerts it currently contributes, and the last error string when it
  failed.
- A dismissible **banner** above the table whenever a source has missed several consecutive polls:
  *"prod-eu has not answered for 4m — its 87 alerts may be stale."* It reappears if the condition
  persists after a reload — this is exactly the state you must not miss.
- Rows sourced from a stale Alertmanager get a discreet stale marker in the existing
  **Alertmanager** column (`col_source`), so a stale alert is identifiable in the table itself.
- Everything updates live over the SSE channel already open — no extra polling from the browser.

## Approach

WebUI-only. No proto change, no new RPC, no backend work.

- `internal/webui/services/alert_cache.go` — the refresh loop already receives
  `(alerts, fetchErrors)` from `FetchAllAlertsDetailed` (`alert_cache.go:178`). Record a
  `map[string]SourceStatus{LastSuccessAt, LastError, ConsecutiveFailures, AlertCount}` next to the
  alert map (same mutex), updated on every refresh: success clears the error and stamps the time;
  an entry in `fetchErrors` increments failures. `AlertCount` is already derivable from the cached
  alerts' `Source`. Expose `GetSourceStatuses()`.
- Staleness is computed, not stored: `stale` when `ConsecutiveFailures >= 3`, `unreachable` when
  the source has never succeeded since startup. The threshold rides on the configured
  `Polling.SyncInterval`, no new config key.
- `internal/webui/handlers/dashboard_handlers.go` — include a `sources` block in the existing
  `GetDashboardData` response (next to `metadata`/`colors`), so the first paint is already correct.
- `internal/webui/handlers/sse_handler.go` — attach the same `sources` block to the update payload
  it already forwards, so the pill and banner refresh live. Not a new event type.
- Client: a new `dashboard_sources.templ` mixin (same `window.dashboard*Mixin` + `Object.assign`
  convention as the other script files, see `dashboard_core.templ:287`) holding `sources` state and
  the banner dismissal; markup in a `components/source_status.templ` component rendered from
  `NewDashboard.templ`'s header. Regenerate with `make webui-templates`; Tailwind via
  `make webui-css`. Never touch `*_templ.go`.
- The stale marker in `col_source` goes in `renderText`'s `field_path === 'source'` branch
  (`dashboard_utilities.templ:renderCell`) — no new column, no column-preference migration.

## Acceptance criteria

- With N Alertmanagers configured and all reachable, the header shows `N/N live` and each popover
  row shows a last-successful-poll under one sync interval.
- Stopping one Alertmanager: within 3 sync intervals the pill turns amber, that source's row shows
  `stale` with its last error, and the banner names the source and its alert count.
- The alerts of a failed source **remain in the table** (current behaviour preserved) and are
  marked stale in the Alertmanager column.
- Bringing the source back clears the state on the next successful poll without a page reload
  (SSE path) and without a full dashboard reload.
- A source that has never answered since startup shows `unreachable`, not `live`, even though it
  contributes zero alerts.
- The pill/banner state arrives with the initial `GET /api/v1/dashboard/data` — no flash of
  "all green" before the first SSE update.
- `GetSourceStatuses()` has a unit test in `alert_cache_test.go` covering
  success → failure → recovery transitions.

## Out of scope

- Any active probing of Alertmanagers. Status is derived **only** from the polls the cache already
  performs — no new `TestConnection()` calls per request, and `GET /health/alertmanager` keeps its
  current probe behaviour untouched.
- Alerting/paging on a dead source (no browser notification, no email). This is a passive surface.
- Per-user configuration of the staleness threshold.
- Any change to the mass-resolve protection (#45) — the keep-stale-alerts behaviour stays exactly
  as it is; this only makes it visible.
- Backend/gRPC connectivity status (separate concern, separate probe).

SoulKyu:
## Summary

- Track per-source poll freshness in `AlertCache` (`LastSuccessAt`, `LastError`, `ConsecutiveFailures`, `AlertCount`), updated on every refresh cycle from the fetches the cache already performs — no new probing, no proto/RPC change.
- State is derived, not stored: `unreachable` if a source has never succeeded since startup, `stale` at 3 consecutive failed polls, `live` otherwise. Expose it via `GetSourceStatuses()` / `GetSourceStatusViews()`.
- Embed the `sources` block in both the initial `GET /api/v1/dashboard/data` response and every SSE incremental update (same struct forwarded as-is, no `sse_handler.go` change needed), so the header pill and banner are correct on first paint and stay live over the open SSE channel.
- Client: a `dashboard_sources.templ` mixin holding the sources state, popover, and per-source banner dismissal (same `window.dashboard*Mixin` convention as the other script files); a `components/source_status.templ` component rendering a `N/N live` pill (green/amber/red) with a per-source popover, plus a dismissible stale-source banner above the alerts table.
- A discreet amber marker is added to stale/unreachable-sourced rows in the existing Alertmanager column (`col_source`) — no new column.

## Why

Notificator merges alerts from every configured Alertmanager into one cache. When a source fails, the cache correctly keeps its last-known alerts (the #45 mass-resolve fix) but only logs it — invisible from the browser. A source dead for 20 minutes still looks live; a source dead since startup contributes zero alerts, indistinguishable from "all quiet". At 3am that ambiguity is expensive. This makes the existing keep-stale-alerts behavior visible without changing it.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/webui/...` (162 passed) — includes `TestAlertCache_GetSourceStatuses` covering success → failure → recovery and the never-answered-since-startup case
- [x] `make webui-templates` / `make webui-css-prod` regenerated with no unexpected diff

Closes #89


internal/webui/templates/components/source_status.templ
```diff
@@ -0,0 +1,76 @@
...
+						<span class="inline-flex items-center px-1.5 py-0.5 rounded text-xs font-medium"
+							  :class="{
+								  'bg-green-100 text-green-800 dark:bg-green-900/40 dark:text-green-200': source.state === 'live',
+								  'bg-amber-100 text-amber-800 dark:bg-amber-900/40 dark:text-amber-200': source.state === 'stale',
+								  'bg-red-100 text-red-800 dark:bg-red-900/40 dark:text-red-200': source.state === 'unreachable'
+							  }"
+							  x-text="source.state"></span>
+					</div>
+					<div class="mt-0.5 text-xs text-gray-500 dark:text-gray-400 flex items-center justify-between">
+						<span x-text="'Last poll: ' + sourceRelativeTime(source.lastSuccessAt)"></span>
```
SoulKyu: **Popover shows a nonsensical timestamp for `unreachable` sources instead of "never"**

`SourceStatusPill()` always renders `'Last poll: ' + sourceRelativeTime(source.lastSuccessAt)` regardless of `source.state`. For a source that has never had a single successful poll (`state === 'unreachable'`), the Go side leaves `LastSuccessAt` at its zero value, but `models.SourceStatus.LastSuccessAt` (internal/webui/models/dashboard.go) is tagged `json:"lastSuccessAt,omitempty"` — `omitempty` is a no-op on `time.Time` because it's a struct, so the JSON payload still carries `"lastSuccessAt":"0001-01-01T00:00:00Z"` instead of omitting the field (verified: `json.Marshal` of a zero-value struct with that tag still emits the field).

`sourceRelativeTime()` in dashboard_sources.templ only returns `'never'` when the value is falsy (`if (!iso) return 'never'`), so it parses that ancient timestamp instead and computes roughly `Math.floor((Date.now() - new Date('0001-01-01T00:00:00Z').getTime())/1000)` (verified in node: ~739,829 days). The popover row for a source that has been down since process startup ends up reading "Last poll: 739829d ago" — exactly the ambiguity this PR sets out to remove per its own description ("a source dead since startup ... indistinguishable from all quiet").

Both sibling call sites already guard against this: `StaleSourceBanner()` further down in this same file branches on `source.state === 'unreachable'` before using `sourceRelativeTime`, and the source marker added in `dashboard_utilities.templ` does the same (`status.state === 'unreachable' ? ... : ... sourceRelativeTime(...)`). Only this popover row is missing that guard.

Suggested fix — mirror the existing guard here, e.g.:
```
<span x-text="source.state === 'unreachable' ? 'Last poll: never' : 'Last poll: ' + sourceRelativeTime(source.lastSuccessAt)"></span>
```


SoulKyu (commented):
One non-blocking finding on this Alertmanager source-freshness feature — the `AlertCache` poll-status tracking, the live/stale/unreachable state derivation, and the SSE/initial-payload wiring all look correct, and the new `TestAlertCache_GetSourceStatuses` test exercises the success → failure → recovery and never-answered-since-startup transitions well. Left one inline comment on `source_status.templ` about a display edge case for sources that have never answered.



internal/webui/templates/components/source_status.templ
SoulKyu: @SoulKyu This has been fixed in 080e161.

Added the suggested source.state === 'unreachable' guard to the popover's 'Last poll' span in internal/webui/templates/components/source_status.templ, so unreachable sources show 'never' instead of parsing the Go zero-value timestamp.




SoulKyu:
**Looper fixer round complete** — 080e161

- ✅ Review comment on `internal/webui/templates/components/source_status.templ` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/170#discussion_r3697992229)
  - Added the suggested source.state === 'unreachable' guard to the popover's 'Last poll' span in internal/webui/templates/components/source_status.templ, so unreachable sources show 'never' instead of parsing the Go zero-value timestamp.


SoulKyu:
## 🧪 QA Report

Test env: `docker compose` stack at this SHA, webui on :12081, **3 configured Alertmanagers** —
`Fake Alertmanager` (live), `Fake AM Two` (live), `Dead AM` → `http://alertmanager-dead:9093` (never
resolvable, i.e. never answered since startup). Sync interval = default 10s. Driven with Playwright
(real Chromium, real SSE channel) + `curl`.

|---|---|---|
| N Alertmanagers all reachable → header shows `N/N live`, popover last poll under one sync interval | ✅ | Reconfigured to 2 reachable sources: pill renders **`2/2 live`** with `bg-green-100`, popover rows both `live` / `Last poll: just now`; `GET /api/v1/dashboard/data` `lastSuccessAt` within 10s of request time |
| Stopping one Alertmanager → within 3 sync intervals pill amber, that source `stale` + last error, banner names source & alert count | ✅ | `docker stop qa170-alertmanager` → **27s** later pill `1/3 live` amber, popover row `stale`, `Last poll: 30s ago`, error `dial tcp: lookup alertmanager … server misbehaving`; banner: *"Fake Alertmanager has not answered for 30s ago — its 25 alerts may be stale."* (wording nit below) |
| Alerts of a failed source remain in the table and are marked stale in the Alertmanager column | ✅ | Before stop: 11 rows from that source; after stale: still 11, and **11/11** of its `col_source` cells gained `<span class="…bg-amber-500…">` + `title="Fake Alertmanager last answered 30s ago"`, applied live over SSE with no reload. All-sources-down case: 31 alerts kept, 31/31 cells marked |
| Bringing the source back clears state on the next successful poll, over SSE, no reload | ✅ | `docker start` → back to `live` in **9s** (single source) / **6s** (both), same page object (`window.__qaReloadCount` unchanged): banner removed, amber dots removed, pill back to green/amber, popover row `just now` |
| A source that never answered since startup shows `unreachable`, not `live`, with 0 alerts | ✅ | `Dead AM` → `state:"unreachable"`, `consecutiveFailures` climbing, `alertCount: 0`; popover shows `unreachable` + **`Last poll: never`** (not `739000d ago`, i.e. the 080e161 fix holds against the `0001-01-01T00:00:00Z` that `omitempty` cannot suppress on a `time.Time`) |
| Pill/banner state arrives with the initial `GET /api/v1/dashboard/data` — no flash of "all green" | ✅ | The initial payload carries the full `sources` block (`{"Dead AM":{"state":"unreachable",…}}`) before any SSE event; F5 while two sources are stale repaints already-amber/red with banners and row markers |
| `GetSourceStatuses()` unit test covering success → failure → recovery | ✅ | `go test ./internal/webui/services/... -run TestAlertCache` → `ok` (0.5s); `TestAlertCache_GetSourceStatuses` asserts live→2 failures still live→3rd failure stale→recovery clears `LastError`, plus the never-succeeded `unreachable` case |

**Adversarial pass:**

- **Boundary / empty** — source contributing **0 alerts** (`Dead AM`): still tracked and shown `unreachable`, does not silently read as "all quiet". **All 3 sources down at once** (`docker stop` both fake AMs): pill goes **red `0/3 live`** (not amber), 3 banners, and the 31 cached alerts stay in the table with every `col_source` cell marked — this is the `len(alertsWithSource)==0 && len(fetchErrors)>0` early-return branch, which pushes an SSE update whose `newAlerts`/`updatedAlerts`/`removedAlerts` are all `null`; the client handled it with **0 console errors** and **no wiped table**. Sources recovering from that state returned to `2/2` in 6s.
- **Malformed / hostile input** — configured a source literally named `Evil">AM`. The new `renderText` branch builds raw HTML, so this is the injection surface: result is `Evil"&gt;&lt;img …&gt;AM` inside the cell, `document.querySelectorAll('td img').length === 0`, `window.__xss` never set — **no XSS**, in the live branch and in the new stale branch. Popover/banner use `x-text`, also inert.
- **Error path** — the "success that is really a failure" case was hunted explicitly: an Alertmanager going away never produced an empty-but-green dashboard. Cached alerts are preserved (#45 behaviour untouched) and the failure is surfaced in three places (pill, banner, row marker) with the real error string.
- **Auth & scope** — `GET /api/v1/dashboard/data` unauthenticated → `401`, with a bogus `session_id` cookie → `401`, `GET /api/v1/dashboard/stream` unauthenticated → `401`. The `sources` block (which embeds internal hostnames in `lastError`) is never served to an anonymous caller.
- **Adjacent flows** — `/` still focuses the alert search box (PR #131 keyboard triage intact); typing `DiskSpace` still narrows the table (26 → 4 rows, restored on clear) while stale markers are rendered; the existing **Alertmanager** filter dropdown still lists both sources; no new column was added, so column preferences are untouched.
- **Persisted state** — dismissed the stale banner, then **reloaded with the condition still true**: all 3 banners reappear (dismissal is session-only, as the criterion requires). Dismissal survives further SSE pushes while the source stays stale, and is forgotten once the source is live again. Reused the same Postgres volume across a `webui` recreate; no migration/state issue.

Non-blocking nits found while attacking (none breaks a criterion, none fails the PR):
1. Banner copy reads *"has not answered for 30s ago"* — `sourceRelativeTime()` already returns a
   `… ago` suffix, so the sentence doubles it. The issue spec wrote *"has not answered for 4m"*.
   (`dashboard_sources.templ` / `source_status.templ` banner branch.)
2. Missing space in the unreachable banner: **"Dead AMhas not answered since startup"** — the
   `<span x-text="source.name">` and the following `<span> has not answered…` collapse.
3. Tooltip double-escape on stale rows: `title` is built from the already-escaped name and then
   re-escaped, so a source named `Prod & EU` hovers as `Prod &amp; EU` (verified with the hostile
   name: `title="Evil&amp;quot;&amp;gt;…"`). Cosmetic only, no injection.
4. One Alpine-internal `TypeError: Cannot read properties of undefined (reading 'after')`
   (`alpinejs@3.14.6` `x-for`) fires on the stale→live recovery **when the Sources popover is left
   open** across the transition. Reproduced 2/2 with the popover open, 0/3 with it closed. No
   functional impact: the popover keeps updating correctly, the pill, banners, markers, search and
   the popover's own reopen all still work afterwards.

**Verdict: PASS**

SoulKyu:

internal/webui/templates/scripts/dashboard_utilities.templ
```diff
@@ -711,6 +711,21 @@ templ DashboardUtilities() {
...
 
+				// Discreet marker for alerts whose Alertmanager source is not live,
+				// so a stale/unreachable-sourced row is identifiable in the table.
+				if (fieldPath === 'source' && this.sources) {
+					const status = this.sources[value];
+					if (status && status.state !== 'live') {
+						const title = status.state === 'unreachable'
+							? `${escaped} has not answered since startup`
+							: `${escaped} last answered ${this.sourceRelativeTime(status.lastSuccessAt)}`;
+						return `<div class="text-sm text-gray-700 dark:text-gray-200 truncate flex items-center gap-1" title="${this.escapeHtml(title)}">
```
SoulKyu, line 722: **Double HTML-escaping in the stale-source tooltip title**

`renderText()` already HTML-escapes the source name into `escaped` (line 707) before this new branch runs. The `title` string built on lines 719-721 embeds that already-escaped `escaped` value, then line 722 wraps it in `this.escapeHtml(title)` again — re-escaping entities that are already entity-encoded.

Concrete failure: an Alertmanager source configured with a name containing `&`, `<`, `>`, `"` or `'` (e.g. `"Prod & DR"`) produces a garbled tooltip. `escaped` first turns `&` into `&amp;`; the second `escapeHtml()` call then turns that `&` (inside `&amp;`) into `&amp;amp;`, so the browser renders the title text as literal `Prod &amp; DR has not answered...` instead of `Prod & DR has not answered...`. Not an XSS risk (double-escaping is safe, just wrong), but it's a real display bug for any source name with HTML-sensitive characters, and this is new code added by this PR.

Suggested fix — escape once, from the raw value, not from the already-escaped one:
```js
const rawTitle = status.state === 'unreachable'
    ? `${value} has not answered since startup`
    : `${value} last answered ${this.sourceRelativeTime(status.lastSuccessAt)}`;
return `<div ... title="${this.escapeHtml(rawTitle)}">
    ${escaped}
    <span class="w-1.5 h-1.5 rounded-full bg-amber-500 dark:bg-amber-400 flex-shrink-0"></span>
</div>`;
```


SoulKyu (commented):
One non-blocking finding on this Alertmanager source-freshness PR: the `AlertCache` poll-status tracking (`recordSourceStatuses` / `GetSourceStatusViews`), the live/stale/unreachable state derivation, and the SSE/initial-payload wiring in `dashboard_handlers.go` all check out, including the just-fixed "never" last-poll display case for sources that have never answered. Locking is correct — the new source-status recording happens after the alert-cache write lock is released, then takes its own lock, so there's no deadlock — and `TestAlertCache_GetSourceStatuses` exercises success → failure → recovery and the never-answered-since-startup case well. Left one inline comment on a double-HTML-escaping bug in the new stale-source-marker tooltip.

PR #100, opened 2026-07-25
fix(webui): sync the alert modal with browser Back/Forward

SoulKyu:
## Problem

The alert-details modal drove `history.pushState` but nothing listened to `popstate`:

- Back with the modal open → URL returns to `/dashboard`, modal stays on screen.
- Back after closing → `/dashboard/alert/<fp>` comes back with no modal.
- Each open/close cycle deposited two dead entries, so ten inspected alerts meant ~20 Back presses to leave the page.
- `closeAlertModal()` pushed the bare path `'/dashboard'`, silently wiping `?severities=…&teams=…` — the shareable filtered URL the dashboard maintains via `updateURL()`.

## Changes

- **`dashboard_core.templ`** — one `popstate` listener registered in `init()`, plus the `alertModalPushedHistory` flag in the component data.
- **`dashboard_modal.templ`**
  - `pushAlertHistoryEntry(fp)` — shared push that appends `window.location.search`, so filter params survive while the modal is open.
  - `syncModalWithLocation(event)` — the URL is the source of truth: `/dashboard/alert/:fp` opens the modal (dispatching to `showResolvedAlertDetails` when the fingerprint is in `resolvedAlerts`), anything else closes it. Neither path touches history.
  - `closeAlertModal(options)` — pops our own entry with `history.back()` instead of pushing a new one; a deep link (no `alertId` in the history state) falls back to `replaceState('/dashboard' + search)` so the user is never sent off the site.
- **`dashboard_resolved_alerts.templ`** — `showResolvedAlertDetails(alert, options)` uses the same shared push and honours `skipHistory`.

The `alertId` marker in the popped state is what distinguishes an entry we pushed (safe to pop on close, so Forward-then-close behaves like open-then-close) from a deep link.

## Verification

`go build ./...` and `go vet ./internal/webui/...` pass. Templates regenerated with the go.mod-pinned templ (`v0.3.906`); no hand edits to `*_templ.go`.

The state machine was checked against a throwaway Node simulation of `window.history` covering the four acceptance paths — open+Back closes with filters intact, close+Forward re-opens, 10 open/close cycles leave a single Back press to exit, and a deep-link close rewrites in place without growing the stack. Not committed: the repo has no JS test harness, and a copy of the mixin in a test file would rot away from the source.

Closes #95

<sub>🔁 Powered by <a href="https://github.com/nexu-io/looper">Looper</a> · runner=worker · agent=claude-code · An autonomous AI dev team for your GitHub repos.</sub>

PR #158, opened 2026-08-01
feat(notifications): scope browser/sound notifications to a saved filter preset

SoulKyu:
## Summary

- Add an optional "Notify me only for alerts matching" preset selector to Notification settings, so browser/sound notifications can be scoped to a saved filter preset instead of firing for every alert matching the severity gate on a shared Alertmanager.
- The selected preset is AND'd with the existing severity gate; `All alerts` (default) preserves today's behaviour byte-for-byte.
- The dashboard view itself is untouched — this only scopes notifications.
- If the selected preset is deleted, the setting degrades to `All alerts` on next load without a JS error.

## Changes

- `proto/alert.proto`: add `notification_filter_preset_id` to `NotificationPreference` and `SaveNotificationPreferencesRequest`; regenerated via `protoc` (no hand-edits to `*.pb.go`).
- Backend: `NotificationPreference` GORM model + `notification_preferences` DB layer + gRPC service round-trip the new column (nullable/empty = all alerts).
- WebUI: `notification_handlers.go` passes the field through `GetNotificationPreferences` / `SaveNotificationPreferences`; `backend_client.go` wires it through the gRPC client.
- WebUI settings UI: `notification_settings.templ` gains the preset `<select>`, populated from the existing `GET /api/v1/dashboard/filter-presets` endpoint (no new RPC); state lives alongside the rest of `notificationPreferences` in `dashboard_settings.templ`.
- Matching: `shouldNotify()` in `notification_service.templ` now also checks the alert against the selected preset's `filter_data` via a small pure matcher function in `internal/webui/static/js/notification_matcher.js` (same list semantics as the server-side `alertPassesAlertLevelFilters`: empty list = no constraint). The preset's `filter_data` is fetched once on load and refreshed after saving — no extra request per alert.

## Test plan

- [x] `go build ./...`, `go vet ./...` — clean
- [x] `go test ./internal/...` — 225 tests passed, no regressions
- [x] `node --test internal/webui/static/js/notification_matcher.test.js` — 5/5 passing (match / no-match / empty-filter-list cases for the new matcher)
- [x] `templ generate` — only the 4 intentionally-touched `.templ` files regenerated
- [ ] Manual browser verification (create a preset, select it in Notification settings, confirm scoped browser/sound notifications) was not performed in this environment — please smoke test before merge

Closes #152

<sub>🔁 Powered by <a href="https://github.com/nexu-io/looper">Looper</a> · runner=worker · agent=claude-code · An autonomous AI dev team for your GitHub repos.</sub>

PR #174, opened 2026-08-02
feat(dashboard): blast-radius panel in the alert modal

### LINKED ISSUE #153: feat(dashboard): blast-radius panel in the alert modal — what else is firing right now around this alert
## User & need

The alert modal answers "what is this alert" (details, comments, history, 30-day sparkline,
Sentry). It does not answer the first question an on-call engineer actually asks at 3am:

> Is this **one** alert, or is it alert #1 of 40 on the same cluster/namespace/node?

Today the only way to find out is to close the modal, go back to the table, and hand-build a
filter or a grouping on a label you guess is relevant — which loses the modal context and costs
30–60 seconds per alert during exactly the minute when a storm is unfolding. Distinguishing
"single noisy target" from "whole cluster degraded" changes the response (silence one vs. page
the infra team) and it should be visible without leaving the modal.

## Proposed feature

A **Related** tab in the alert modal listing the other currently-firing alerts that share a label
value with this one:

- Grouped by the label that correlates: `cluster=prod-eu (12 firing)`, `namespace=payments (7)`,
  `instance=node-14 (3)`, `alertname=<same> (5 other targets)`.
- Only labels that actually correlate are shown — a label whose value is shared by 1 other alert or
  by *every* alert in the estate is not signal, so cap the display to the top few correlating
  labels by count.
- Each group expands to the matching alerts (name, severity, target, age); clicking one opens that
  alert in the modal (reusing the existing deep-link/modal navigation).
- Each group has a `Filter dashboard on this` action that closes the modal with that label filter
  applied — the escape hatch into the existing table when the blast radius is the real incident.

## Approach

- WebUI-only. No proto change, no schema change, no new persistence.
- New handler in `internal/webui/handlers/dashboard_handlers.go`:
  `GET /api/v1/dashboard/alert/:fingerprint/related`, registered next to the existing
  `/alert/:fingerprint/history` route in `internal/webui/router.go`.
- It reads a snapshot from the live cache (`AlertCache.GetAlert` +
  `AlertCache.GetAllAlerts` in `internal/webui/services/alert_cache.go` — snapshots, never live
  pointers, per #70), counts co-firing alerts per shared label value in a pure helper, drops
  degenerate labels (count of 1, or covering essentially the whole active set), and returns the
  top correlating labels with a bounded number of alerts each.
- UI: a `Related` tab in the modal alongside Comments/History, following the existing
  `AlertModalTabButton` pattern in
  `internal/webui/templates/components/alert_modal_shared.templ`; loaded lazily when the tab is
  first opened. Regenerate with `make webui-templates` (never edit `*_templ.go`).
- Respects the current user's hidden alerts/rules the same way the dashboard does, so a muted alert
  does not reappear here.

## Acceptance criteria

- Opening `Related` on an alert that is the only one firing shows an explicit empty state, not a
  spinner or an empty box.
- Opening it during a storm shows the correlating label groups ordered by count, and the counts
  match what the dashboard shows when the same filter is applied.
- Clicking a related alert opens it in the modal and Back returns to the previous alert
  (consistent with the existing modal history behaviour from #100).
- `Filter dashboard on this` closes the modal and leaves the table filtered on that label value.
- The endpoint does no Alertmanager call and no backend RPC — cache only — and stays comfortably
  sub-100ms on a few thousand active alerts.
- A unit test covers the correlation helper (degenerate labels dropped, ordering by count).

## Out of scope

- Any cross-alert correlation over *resolved* history or statistics — live firing alerts only.
- Causality, root-cause ranking, or scoring beyond "shares a label value".
- Bulk-acting on the related set from inside the modal (bulk actions stay on the table).
- Persisting a per-user choice of which labels to correlate on.

SoulKyu:
## Summary

- Adds a **Related** tab to the alert modal listing other currently-firing alerts that share a label value with the open one, grouped by the correlating label (`cluster=prod-eu (12 firing)`, `namespace=payments (7)`, ...) — answers "is this one alert or the whole cluster" without leaving the modal.
- New endpoint `GET /api/v1/dashboard/alert/:fingerprint/related`, cache-only (`AlertCache.GetLiveAlert` + `GetAllAlerts`, no Alertmanager call, no backend RPC), respects the current user's hidden alerts/rules.
- A pure `computeRelatedAlertGroups()` helper counts co-firing alerts per shared label value, drops degenerate labels (shared by only one other alert, or covering essentially the whole active set), orders by count desc, and caps groups (5) and alerts per group (20). Covered by a unit test (degenerate-label dropping, ordering, capping, empty-when-alone).
- Each group has a `Filter dashboard on this` action that closes the modal and reuses the existing search filter — the escape hatch into the table when the blast radius is the real incident.
- Clicking a related alert reuses `showAlertDetails()`, so it opens in the modal and the existing Back/Forward history behavior (#100) keeps working for free.
- Related tab follows the existing lazy-loaded tab pattern (History, Sentry): fetched only when the tab is first opened.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...`
- [x] `make webui-templates` (regenerated `*_templ.go`, never hand-edited)
- [ ] Manual: open the alert modal, click Related during a storm and with a single isolated alert (empty state), click a related alert, use "Filter dashboard on this"

Closes #153


SoulKyu (commented):
@SoulKyu Reviewed the blast-radius Related tab: `computeRelatedAlertGroups` (`internal/webui/handlers/alert_related.go`) is well-tested (degenerate-label dropping, ordering, capping, empty-when-alone all covered), the endpoint is cache-only and auth-protected as described, and `templ generate` reproduces the committed `_templ.go` files exactly, so nothing was hand-edited. Found one bug in the new tab's client-side staleness guard that leaks the previous alert's related-alert data into a different alert's modal — this reads as blocking severity (it makes the feature show the wrong incident's blast radius during triage), posted as COMMENT only because GitHub rejects REQUEST_CHANGES on a self-authored PR. Details inline.



internal/webui/templates/components/modal_components.templ
```diff
@@ -1141,6 +1141,19 @@ templ AlertDetailsModal() {
 											  x-text="alertHistory?.total_occurrences || 0">
 										</span>
 									</button>
+									<!-- Related tab with special click handler -->
+									<button @click="currentAlertTab = 'related'; if (!relatedAlerts && !relatedLoading) loadRelatedAlerts()"
```
SoulKyu: **Related tab shows the previous alert's data after switching to a resolved alert, with no refetch and no indication it's stale.**

The click handler only fetches when `relatedAlerts` is falsy: `if (!relatedAlerts && !relatedLoading) loadRelatedAlerts()`. `relatedAlerts` is only ever reset to `null` in `showAlertDetails` (`internal/webui/templates/scripts/dashboard_modal.templ:59`), the live-alert open path. `showResolvedAlertDetails` (`internal/webui/templates/scripts/dashboard_resolved_alerts.templ:246`), which this PR does not touch, never clears `relatedAlerts`.

Concrete repro: open a currently-firing alert that has an active blast radius (e.g. `cluster=prod-eu`, 12 co-firing alerts), open its Related tab — `relatedAlerts` is now populated with that alert's groups. Close the modal and open a *resolved* alert (via the Resolved tab). Click Related on this new modal: the guard sees `relatedAlerts` is still non-null from the previous alert, skips `loadRelatedAlerts()` entirely, and the panel renders the first alert's 12 correlated alerts as if they belong to the resolved one — no request is made, no error, no stale-data indicator. This is exactly the "is this one alert or the whole cluster" judgment call the feature exists to support, so showing another incident's blast radius here is actively misleading during triage, not just a cosmetic gap.

The backend response already carries the fingerprint it was computed for (`internal/webui/handlers/alert_related.go` — `"fingerprint": fingerprint` in `HandleGetRelatedAlerts`), so the fix can stay entirely inside this PR's diff: key the guard off that instead of mere presence. For example, in `dashboard_modal.templ`, change the check in `loadRelatedAlerts`/the click handler to compare `this.relatedAlerts?.fingerprint` against `this.alertDetails?.alert?.fingerprint` and refetch whenever they differ, e.g. `if ((!relatedAlerts || relatedAlerts.fingerprint !== alertDetails?.alert?.fingerprint) && !relatedLoading) loadRelatedAlerts()`. That makes the tab self-healing regardless of which code path opened the modal, instead of relying on every current and future "open alert details" entry point remembering to reset `relatedAlerts`.


SoulKyu: @SoulKyu This has been fixed in 5f53667.

Changed the Related tab click handler guard in internal/webui/templates/components/modal_components.templ:1145 to compare relatedAlerts.fingerprint against alertDetails.alert.fingerprint (instead of just checking truthiness), so switching to a different alert (e.g. a resolved one) always refetches instead of showing stale data. Regenerated templ and verified go build/vet pass.




SoulKyu:
**Looper fixer round complete** — 5f53667

- ✅ Review comment on `internal/webui/templates/components/modal_components.templ` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/174#discussion_r3698742236)
  - Changed the Related tab click handler guard in internal/webui/templates/components/modal_components.templ:1145 to compare relatedAlerts.fingerprint against alertDetails.alert.fingerprint (instead of just checking truthiness), so switching to a different alert (e.g. a resolved one) always refetches instead of showing stale data. Regenerated templ and verified go build/vet pass.


SoulKyu:
## 🧪 QA Report

Env: `docker compose` stack from this branch (webui :9081, fake Alertmanager :10093), real browser (Playwright/Chromium), 2 965 active alerts at peak.

|---|---|---|
| `Related` on an alert that is the only one firing shows an explicit empty state | ✅ | Injected an alert with unique labels (`alertname/severity/job/team/instance` all unique) → tab renders “No other alerts are currently firing with a matching label — this looks isolated.”, no spinner, no empty box |
| During a storm: correlating groups ordered by count, **counts match the dashboard when the same filter is applied** | ❌ | Groups are ordered by count desc and capped correctly, but the counts are inflated: the handler uses raw `AlertCache.GetAllAlerts()` while the dashboard uses `getStandardAlerts()` (`!IsResolved && !IsAcknowledged`). Panel: `evil=<…>` **11 firing** — dashboard shows **5** alerts with that label; 7 of the 11 listed return `isResolved:true`/`status.state:"resolved"` from `/api/v1/dashboard/alert/<fp>`. Second sample: `zone=qa-truncate` **50 firing** vs **9** rows on the dashboard; of the 20 listed, 13 resolved + 2 acknowledged + 5 actually firing |
| Clicking a related alert opens it in the modal and **Back returns to the previous alert** | ❌ | Opened A (`/dashboard/alert/A`) → Related → clicked B: modal opens B (`/dashboard/alert/B`) ✅. Browser **Back closes the modal** and lands on `/dashboard` — A is never restored. Cause: `pushAlertHistoryEntry()` `replaceState`s instead of pushing when `alertModalPushedHistory` is already true, so B overwrites A's entry |
| `Filter dashboard on this` closes the modal and leaves the table filtered on that label value | ❌ | Modal closes ✅, but `filterDashboardOnRelatedLabel()` sets `searchQuery = group.labelValue` — a free-text search over alertName/instance/summary/team/source + every label key **and** value, not a `labelKey=labelValue` filter. Group `cluster=prod (7 firing)` → table shows **13 rows**: 4 `cluster=prod`, **8 `cluster=prod-eu`** (substring match) and 1 `QuerylyApplicationError` that merely contains “prod”. The panel count and the table it produces disagree by design |
| Cache-only endpoint (no Alertmanager call, no backend RPC), sub-100ms on a few thousand alerts | ✅ | 6 consecutive calls at **2 965 active alerts**: 7.0 / 8.2 / 18.5 / 9.2 / 9.1 / 8.7 ms. Alertmanager access log over the same window shows only the 5 s background cache poll — no fetch attributable to `/related`; handler touches `GetLiveAlert` + `GetAllAlerts` only |
| Unit test covers the correlation helper (degenerate labels dropped, ordering by count) | ✅ | `go test ./internal/webui/handlers/ -run TestComputeRelated` → 4 tests pass (degenerate drop, count-desc ordering, 20-per-group cap + `Truncated`, empty-when-alone) |

**Adversarial pass** (what I tried to break each criterion with):

- **Boundary & empty** — alert with wholly unique labels → correct empty state; label shared with exactly 1 other alert → dropped (as specified); 25-alert group → list capped at 20 with “Showing first 20 of 49.” ✅; group cap of 5 respected with 5+ correlating labels; 300-alert mono-cluster storm + 2 965-alert estate → groups still computed and ordered.
- **Malformed & hostile** — unknown fingerprint → 404; 4 000-char fingerprint → 404; `../../etc/passwd` → 404; `' OR 1=1--` → 404; `%20` → 404. Hostile **label value** (``, an attacker-influenceable field) renders as text through `x-text` — `window.__qaXss` undefined, 0 `img[src=x]` in the DOM, no injection. ✅
- **Error & unavailable path** — ❗ with `/related` failing (404 injected via route interception, everything else healthy) the Related pane renders **completely blank**: no error, no empty state, no retry, only `console.error`. `loadRelatedAlerts()` sets `relatedAlerts = null` on failure, and the empty-state `x-show` requires `relatedAlerts` to be truthy, so nothing is shown at all. Full network loss is masked by the app-wide “Maintenance in progress” overlay, so the blank pane only surfaces on a per-request failure (alert purged from the cache between opening the modal and opening the tab, 5xx, expired-then-reissued session).
- **Auth & scope** — unauthenticated → 401; forged `session_id` cookie → 401; hidden alerts respected per session (hid 3 → group count 449 → 446, hidden fingerprints gone from the list) and **not** leaked across users (second account still sees 449). ✅
- **Adjacent flows** — History and Comments tabs still load after using Related; the per-alert refetch guard from 5f53667 holds (after A → B the payload fingerprint matches the open alert, no stale data); the Related tab badge tracks the group count.
- **State that outlives the cache / persisted state** — acknowledgements are persisted in the backend and removed from the dashboard's classic view; the panel still counts those alerts as “firing” (2 of the 20 listed in the `zone=qa-truncate` sample). Same for alerts already resolved before the tab is opened — one of them, when clicked, opens a modal whose header reads **RESOLVED** while the group that linked to it says “firing”.

**Verdict: FAIL** — 3 of 6 acceptance criteria fail (counts vs. dashboard, Back navigation, `Filter dashboard on this`). The fixer has been summoned.

SoulKyu (commented):
@SoulKyu Re-reviewed the blast-radius panel after the QA fail on 5f53667. I went through this fix commit (f0729cd) adversarially against all three failed criteria:

- **Count mismatch**: `HandleGetRelatedAlerts` now builds its `active` population from `isClassicViewAlert` (the same `!IsAcknowledged && !IsResolved` predicate `getStandardAlerts()` uses), so a group's `count` can no longer include alerts the dashboard table has already resolved or acknowledged away. Verified with `TestRelatedEndpointCountMatchesDashboardRows`, which seeds resolved/acknowledged/substring-neighbour alerts and asserts the endpoint's count equals `len(applyDashboardFilters(getStandardAlerts(), ...))` for the exact same label filter — that's a real end-to-end reconciliation, not just an assertion on the handler in isolation.
- **Back navigation**: the old boolean `alertModalPushedHistory` (which swapped instead of stacking on a chained open) is replaced with an `alertModalHistoryDepth` counter plus an `alertModalClosing` flag, so opening a related alert from inside the modal pushes a real entry and Back walks the chain instead of closing straight to `/dashboard`. `dashboard_modal_history.test.mjs` evals the actual shipped `<script>` block and exercises the chain-open, chained-close, deep-link, and Forward-after-close cases — I ran it (`node --test`) and all four pass.
- **`Filter dashboard on this` substring bug**: there's now a real exact-match `LabelFilters` path (`DashboardFilters.LabelFilters` / `alertMatchesLabelFilters` server-side, `filters.labels` client-side) instead of stuffing the label value into the free-text search box, so `cluster=prod` can no longer catch `cluster=prod-eu` by substring. Covered on both sides (`TestAlertMatchesLabelFilters`, plus the JS predicate in `dashboard_data.templ`).

I also checked the couple of things a fix like this tends to regress: `dashboard_resolved_alerts.templ` now clears `relatedAlerts`/`relatedError` too (belt-and-suspenders next to the fingerprint-comparison guard from the prior round), the new `/related` route sits inside the authenticated `dashboard` group in `router.go`, and label values with an embedded `=` are cut consistently on both the Go (`strings.Cut`) and JS (`indexOf('=')`) sides. `go build ./...`, `go vet ./...`, `go test ./...`, and `make webui-templates` (zero diff after regeneration, confirming nothing was hand-edited) all came back clean locally.

This was a genuinely thorough fix round — nice work closing out all three QA findings with regression tests that actually pin the behavior instead of just re-asserting the fix. No blocking or non-blocking findings from me this round.



SoulKyu:
## 🧪 QA Report

Env: `docker compose` stack rebuilt from this SHA (webui image rebuilt — the previous one was stale at `5f53667`), webui :9081, fake Alertmanager :10093. Real browser (Playwright/Chromium) + raw HTTP. Peak population **3 054 alerts in the classic view**.

|---|---|---|
| `Related` on an alert that is the only one firing shows an explicit empty state | ✅ | Alert with wholly unique labels (incl. a unique `severity` so nothing correlates) → 0 groups, pane renders “No other alerts are currently firing with a matching label — this looks isolated.”, no spinner, no empty box |
| During a storm: groups ordered by count, **counts match the dashboard when the same filter is applied** | ❌ | Clean case is now correct (see below). **Breaks with a saved filter preset active**: preset `QA noisy-cluster` (hidden rule `alertname =~ QAe2blastA1.*`) → Load → open `QAe2blastA07` → Related → badge reads **“18 firing”** → `Filter dashboard on this` → table reads **“Showing 10 alerts”**, header “Total 10”. `filterDashboardOnRelatedLabel()` resets `filters`/`activePresetName` but leaves `filterHiddenRules` (1 rule still in state after the click), and `/related` only excludes *globally* hidden alerts — so 8 alerts are dropped from the table the badge just promised, with the preset indicator cleared and no chip to explain it |
| Clicking a related alert opens it in the modal and **Back returns to the previous alert** | ✅ | A (`depth=1`) → Related → click B (`depth=2`, URL `/dashboard/alert/<B>`); **Back → A restored, modal still open**; Back again → `/dashboard?search=…` (pre-modal URL, filters intact); Forward → A reopens. 3-deep chain A→B→C (`depth=3`) closed with the ✕ → one step to `/dashboard?search=…`. Deep link (`depth=0`) → related child (`depth=1`) → Back → the deep-linked alert → close → `/dashboard`, never off-site |
| `Filter dashboard on this` closes the modal and leaves the table filtered on that label value | ✅ | Modal closes, URL `?label=qacluster%3Dblast-a-c1`, chip `qacluster=blast-a-c1` in the active-filter bar, `searchQuery` emptied, `displayMode` forced to classic. Panel 17 = table 17, **0 stray rows**: the `blast-a-c1b` substring neighbours are gone (free-text search used to drag them in). Survives a page reload; the chip’s ✕ clears it (17 → 50 rows). *(The preset case above lands you on an under-populated table — same defect, counted once under the criterion above.)* |
| Cache-only (no Alertmanager call, no backend RPC), sub-100ms on a few thousand alerts | ✅ | **3 054 active alerts**: 12 consecutive calls at 7.4 / 7.9 / 8.0 / 8.0 / 9.0 / 9.3 / 10.2 / 10.4 / 10.5 / 13.4 / 13.4 / 13.8 ms. 40 `/related` calls in 0.43 s produced **1** Alertmanager API request in the same window (the 5 s background cache poll) |
| A unit test covers the correlation helper (degenerate labels dropped, ordering by count) | ✅ | `go test ./internal/webui/handlers/` → 10 tests pass, incl. `TestComputeRelatedAlertGroupsCountsOnlyTheClassicView`, `TestRelatedEndpointCountMatchesDashboardRows`, `TestRelatedGroupCountMatchesLabelFilteredDashboard`, `TestAlertMatchesLabelFilters`. `node --test dashboard_modal_history.test.mjs` → 4/4 pass |

**Adversarial pass — what I threw at each criterion**

- **The previous FAIL cases, re-run (all fixed).** Acknowledged alerts: acked 4 of a 10-alert group → panel 10 → **6**, table 6. Resolved: no resolved alert ever appears in a count now (`isClassicViewAlert` is shared with `getStandardAlerts`). Substring: `cluster=prod` vs `cluster=prod-eu` equivalent → exact filter yields 10 vs 4 where free-text yielded 12. Back navigation: restores the parent alert instead of closing the modal. Blank error pane: `/related` forced to HTTP 500 via `fetch` interception → red “Unable to load related alerts (HTTP 500).” + **Retry**, and Retry recovers (screenshot). 404 → “no longer in the live cache”.
- **Boundary & empty** — 1 other sharer → group dropped; exactly 2 → kept; label value shared by ~100 % of the estate (`qaempty=""`) → dropped as degenerate; 40-alert group → list capped at 20, footer “Showing first 20 of 39 other alerts.” (counts the *others*, not the total); 7 correlating labels → exactly 5 groups returned, ordered 9/8/7/6/5 desc, **every one matching its table count**; 3 054-alert estate.
- **Malformed & hostile** — unknown fingerprint, `../../etc/passwd`, `' OR 1=1--`, `%00`, a space, a 3 000-char fingerprint, `<script>x</script>` → 404, no 500, no stack. Attacker-influenceable **label value** `` rendered through `x-text` → `window.__qaXss` undefined, 0 `img[src=x]`; the same value round-trips through the URL filter and the chip intact. Label value containing `=` (`qanote=k=v`) → server `strings.Cut` on the first `=` matches the client parser, panel 3 = table 3. Value with `a&b#c d%e+f/g?h` → panel 4 = table 4. Hand-crafted `?label=` params: `''`, `'='`, `'=value'`, `'key'` are ignored; `?label=key=` is *not* — it filters on the empty value and so matches every alert lacking that label (not reachable from the panel, which drops such groups; cosmetic).
- **Error & unavailable path** — covered above; also full network loss (masked by the app-wide maintenance overlay) and a per-request 500 with everything else healthy.
- **Auth & scope** — unauthenticated → 401; forged `session_id` cookie → 401. Cross-user: user `qa` hid 4 of a 10-alert group → its panel reads 6 while user `qa2` still reads 10 on the same alert, and `qa2`’s table for the same label filter reads 10. No leak either way.
- **Adjacent flows** — the label filter combines with search and severity (`label` only 5 → `+search` 0 → `+severity=critical` 0); `Clear All` and the chip’s ✕ both drop it; a page reload rebuilds it from the URL; typing `/` into the search box still works. Live updates with the filter on: injected 6 matching + 6 non-matching alerts → **6 matching rows arrived, 0 non-matching leaked** (the client-side SSE predicate mirrors the server). History/Comments tabs still load after Related.
- **State that outlives the cache / persisted state** — acknowledgements persist in Postgres and are honoured on the next poll (10 → 6 above); per-session hidden alerts persist and are honoured; a **saved filter preset** persisted in the backend is the one that breaks it — see the ❌.

Root cause for the fixer: `filterDashboardOnRelatedLabel()` (`internal/webui/templates/scripts/dashboard_modal.templ`) rebuilds `this.filters` and nulls `activePresetName` but never clears `this.filterHiddenAlerts` / `this.filterHiddenRules` / `activeFilterPresetId`, while `HandleGetRelatedAlerts` counts over global-hidden-only. Either clear the preset’s hidden state along with the preset, or make `/related` aware of it.

**Verdict: FAIL** — 1 of 6 acceptance criteria fails. The fixer has been summoned.

SoulKyu (commented):
@SoulKyu Reviewed this round's fix commit (69322ffd) adversarially against the QA failure it targets (preset-scoped hidden rules leaking into the blast-radius count).

`HandleGetRelatedAlerts` (`internal/webui/handlers/alert_related.go`) no longer hand-rolls its own "active alerts" predicate; it now calls the exact same `getStandardAlerts()` + `applyDashboardFilters()` path the dashboard table itself uses, fed with `filterHiddenAlerts`/`filterHiddenRules` parsed from the query string via the newly extracted `parseFilterHiddenState()` helper (shared with `parseDashboardFilters` in `dashboard_handlers.go`, so both endpoints can no longer drift apart). Client-side, `appendFilterHiddenParams()` in `dashboard_utilities.templ` centralizes building those query params and is now used by all four call sites (`loadDashboardData`, `loadDashboardIncremental`, `loadAlertColors`, and the new `loadRelatedAlerts`), so the Related panel's fetch carries the same preset-hidden state the table's own requests do. I traced this end-to-end and it holds together — no endpoint is left with a stale hand-copied predicate.

`TestRelatedEndpointHonoursPresetHiddenRules` is a real regression test for the exact bug QA found: it seeds a mixed population, applies a preset hidden rule (`alertname=NoisyProbe`), asserts the muted alerts are absent from the panel's group, and then independently reconciles the panel's count against `applyDashboardFilters` called with the same label + hidden-rule state "Filter dashboard on this" would land on — a genuine cross-check against the table's own filter chain, not just an assertion on the handler in isolation.

I also checked the UI change in `NewDashboard.templ` splitting the hidden-state badge out from the preset-name indicator (so it now shows independently once a preset's hidden alerts/rules outlive the preset via "Filter dashboard on this"), and confirmed `filterDashboardOnRelatedLabel()` deliberately leaves `filterHiddenAlerts`/`filterHiddenRules` untouched when pinning the table to a label filter, consistent with the panel's count already accounting for that hidden state.

Locally: `go build ./...`, `go vet ./...`, and `go test ./internal/webui/handlers/...` all pass, and `make webui-templates` regenerates with zero diff against the committed `_templ.go` files, confirming nothing was hand-edited. Nice, tightly-scoped fix — no blocking or non-blocking findings from me this round.



SoulKyu:
## 🧪 QA Report

Env: `docker compose` stack rebuilt from this SHA (webui image contains `appendFilterHiddenParams` ×5 — verified inside the image, not assumed), webui :9081, fake Alertmanager :10093. Real browser (Playwright/Chromium) + raw HTTP. Peak population **2 770 alerts in the classic view**.

|---|---|---|
| `Related` on an alert that is the only one firing shows an explicit empty state | ✅ | `QAloneOne` (alertname/job/instance/severity/team/qacluster all unique) → 0 groups, pane renders “No other alerts are currently firing with a matching label — this looks isolated.”, no spinner inside the related pane, no empty box |
| During a storm: groups ordered by count, **counts match the dashboard when the same filter is applied** | ✅ | Panel = table in **14** configurations at 2 770 active alerts. **The escalation case is fixed**: preset `QA noisy-cluster` (hidden rule `alertname=QAnoisyProbe`) → open `QAkeepA00` → Related badge **15 firing** → `Filter dashboard on this` → table **15**, `0` stray rows (was 18 vs 10 at `f0729cd`). Also: mixed preset (severities+hidden rule) 24=24, 400 preset-hidden fingerprints in a 49 KB query 782=782, session-global hidden 32=32, 6 acknowledged 26=26, regex rule 26=26, disabled rule 36=36, empty-value rule → groups collapse and still match |
| Clicking a related alert opens it in the modal and **Back returns to the previous alert** | ✅ | A `QAkeepA00` → Related → expand `qacluster=blast-a` (20 entries) → click `QAnoisyProbe/probe-a07`: modal opens B, `depth=2`, URL `/dashboard/alert/<B>`; **Back → A restored, modal still open** (`alertDetails.fingerprint == A`); Forward → B reopens |
| `Filter dashboard on this` closes the modal and leaves the table filtered on that label value | ✅ | Modal closes, URL `?label=qacluster%3Dblast-a`, `searchQuery` emptied, `severities` dropped, `displayMode` forced to classic, preset-hidden rules kept → badge=table in every case (15 / 60 / 24 / 4). The kept hidden state is now explained by a standalone amber `1 hide rule active` chip whose ✕ reveals the muted rows (15 → 26). Reload keeps the label filter and the exact label value, including hostile and `=`/`&`/`#`/`%`-laden values |
| Cache-only (no Alertmanager call, no backend RPC), sub-100ms on a few thousand alerts | ✅ | **2 770 active alerts**: 12 consecutive calls at 6.4 / 6.5 / 6.9 / 7.6 / 8.1 / 8.2 / 8.4 / 8.5 / 8.5 / 8.8 / 12.9 / **13.3 ms**; 30.8 ms with a 49 KB hidden-state payload. 40 `/related` calls in 0.37 s → **0** Alertmanager `GET /api/v2/alerts` in the window. Handler touches `GetLiveAlert` + `getStandardAlerts()`(`GetAllAlerts`) only; no RPC surface added vs. the previous revision |
| A unit test covers the correlation helper (degenerate labels dropped, ordering by count) | ✅ | `go test ./internal/webui/handlers/` → **72 pass**, incl. the new `TestRelatedEndpointHonoursPresetHiddenRules` (asserts group count == `applyDashboardFilters` rows with the same hidden rules) plus `TestRelatedEndpointCountMatchesDashboardRows`, `TestRelatedEndpointIsolatedAlert`, `TestRelatedEndpointUnknownFingerprint` |

**Adversarial pass — what I threw at each criterion**

- **The `f0729cd` FAIL, re-run and broken open further.** Not just the reported preset+hidden-rule case: exact rule, RE2 regex rule, disabled rule, empty-`label_value` rule ("match all"), 400 preset-hidden **fingerprints**, session-global hidden alerts, a server-side global hidden rule, a rule that mutes the **open alert itself**, and every combination of fingerprints+rules — panel and table agree in all of them. Also re-ran the three earlier regressions: 6 acknowledged alerts (32 → 26 both sides), resolved alerts (never counted), substring neighbours (`blast-a` vs `blast-a-c1`-style: exact label filter, 0 stray rows).
- **Boundary & empty** — isolated alert → empty state; group of exactly 2 kept, 1 other dropped; 60-alert group → 20 listed + “Showing first 20 of 59 other alerts.”; 7 correlating labels → exactly **5** groups, ordered count-desc with label-key-asc tiebreak (`job, qacap1..qacap4`); label covering 40/41 of the visible estate → dropped as degenerate (note: the ratio is now relative to the *filtered* population, so an aggressive preset can retire a label from the panel — deliberate, and the count it would have shown still matches the table).
- **Malformed & hostile** — fingerprints: unknown, `../../etc/passwd`, `' OR 1=1--`, `%00`, `..%2f..%2fadmin`, `<script>alert(1)</script>`, 4 000 chars → 404, never 500, no stack. Hidden-state params: garbage JSON, JSON object instead of array, wrong types (`label_key:123`), `null`, `[null,null]`, invalid regex `([a-`, JS-only lookahead `blast-(?=a)`, catastrophic `(a+)+$` → **panel and table agree on all 8**, ≤11 ms, no ReDoS (RE2). Attacker-influenceable **label value** `` renders through `x-text` → `window.__qaXss` undefined, 0 `img[src=x]`, and survives the filter round-trip + reload still inert. Value `k=v&x#y z%p+q/r?s` (`=`, `&`, `#`, space, `%`, `+`, `/`, `?`) → panel 4 = table 4, intact after reload.
- **Error & unavailable path** — `/related` forced to HTTP 500 → red “Unable to load related alerts (HTTP 500).” + **Retry**, and Retry recovers (5 groups render). 404 → “This alert is no longer in the live cache, so its blast radius cannot be computed.” Backend container stopped mid-flow → `/related` **503**, exactly like `/data` (auth path) — no blank pane sold as success. No uncaught page errors and no failed requests in the steady-state flow.
- **Auth & scope** — unauthenticated → 401; forged `session_id`/`session` cookie → 401. Cross-user: `qa` hides 4 alerts globally → its panel reads 32 while `qa2` reads 36 on the same alert, and `qa2`'s table for the same label reads 36. No leak either way.
- **Adjacent flows** — `/` still focuses the search box and typing into it works (no hotkey hijack); `j`/`k`/`o` cursor navigation still opens the modal; History and Comments still load after Related and Related re-renders after switching back; the tab badge tracks the group count; the green preset chip's ✕ still clears everything (159 → 170 rows); the new amber ✕ clears only the hidden state. **Live updates with a preset active**: injected 5 alerts matching the hidden rule + 5 not → **5 arrived, 0 muted leaked** (the refactored `appendFilterHiddenParams` is wired into the incremental path correctly). Panel opened from the **acknowledged** view (whose rows are not in the counted population) → click still lands on classic + label with badge 881 = table 881.
- **State that outlives the cache / persisted state** — acknowledgements persisted in Postgres are honoured on the next poll; per-session hidden alerts persisted in the backend are honoured and not leaked; a **saved filter preset** persisted in the backend — the exact thing that failed last round — now flows into the count. One residual, non-blocking: a preset's hidden state is client-side only, so after `Filter dashboard on this` a page reload drops it (table 15 → 26) *together with* the amber chip that explained it — the view stays internally coherent and this is the pre-existing behaviour of any loaded preset, not something this PR introduces. Cosmetic: the amber ✕ leaves `activeFilterPresetId` set, so “Hide in Filter” stays offered with no visible preset name.

Out of scope, pre-existing, reproduced without any of this PR's code paths: the `Total` header counter does not follow incremental updates (36 rows → 40 rows, header still 36), and logging in as a non-existent user 500s on a nil `resp.User` (`internal/webui/client/backend_client.go:173`).

**Verdict: PASS**

SoulKyu:

SoulKyu (commented):
@SoulKyu Reviewed the blast-radius Related tab at this head (1ac990a5) after re-checking against my three prior review rounds on this PR. This commit turned out to be a rebase onto main's newly-merged auth-interceptor PR (#172) with zero changes to the feature's own files — `internal/webui/handlers/alert_related.go`, its tests, the `dashboard_handlers.go` filter-chain plumbing, and the `dashboard_modal.templ` history/staleness fixes are byte-identical to the commit I approved as clean in my last round.

I still ran the full pass rather than trusting that: re-read `computeRelatedAlertGroups` and `HandleGetRelatedAlerts` end to end, re-verified the endpoint routes through `getStandardAlerts()` + `applyDashboardFilters()` (the same path the dashboard table itself uses) so the panel's counts can't diverge from what "Filter dashboard on this" would show, checked the exact-match `LabelFilters` on both the Go and JS sides for the substring-leak bug from an earlier round, and re-read the history-depth/staleness-guard logic in `dashboard_modal.templ` for the two prior blocking bugs (Back-navigation and cross-alert data leak). All of it holds. Locally: `go build ./...`, `go vet ./...`, and `go test ./internal/webui/...` are all green, `node --test` on `dashboard_modal_history.test.mjs` passes all 4 cases, and `make webui-templates` regenerates the PR's own files with zero diff (a stray diff I saw across unrelated files turned out to be a local `templ` CLI version mismatch, not anything from this branch).

No blocking or non-blocking findings from me this round. Nicely converged feature — thanks for grinding through all three QA rounds with regression tests that actually pin the fixed behavior instead of just re-asserting it.



SoulKyu:
🔀 Rebase (conflicts resolved): rebased onto `main` (2ae143f).

**Conflicts:** all 3 conflicting commits collided only in generated `*_templ.go` files — the `.templ` sources auto-merged cleanly. Resolved per repo convention: verified the merged `.templ` sources, then regenerated with `make webui-templates` (templ v0.3.906). No hand-edits to `*_templ.go`.

- `289d8b6` blast-radius panel → `dashboard_core_templ.go`
- `04e6a8b` panel/table filter agreement → `dashboard_core_templ.go`, `dashboard_data_templ.go`, `dashboard_utilities_templ.go`
- `1ac990a` blast radius through the filter chain → `NewDashboard_templ.go`, `dashboard_data_templ.go`, `dashboard_utilities_templ.go`

**Semantic checks after merge** (both sides preserved, no interleaving):
- label-filter plumbing intact end to end (`filters.labels`, `removeLabelFilter`, `label=` query params, client-side `alertMatchesFilters`, preset reset)
- `appendFilterHiddenParams` single builder still used by all 4 callers (dashboard data, updates, alert colors, blast radius)
- `NewDashboard.templ` filter bar: active-preset indicator and the new standalone hidden-state badge coexist alongside main's saved-filters/freshness markup

**Extra commit:** `main` was already failing `go vet` before this rebase — `internal/backend/auth_interceptor_test.go` still called the 3-arg `NewAuthServiceGorm` after #158 added `allowRegistration`. Fixed in a separate commit (`true`, matching the config default) so the gate is green; unrelated to the blast-radius work and safe to cherry-pick to `main`.

**Verification:** `go build ./...` ✅ · `go vet ./...` ✅ · `go test ./internal/webui/...` ✅ (175 tests, 11 packages)

SoulKyu (commented):
@SoulKyu Re-reviewed the blast-radius Related tab at this head (a5f202a8). The only change since my last clean pass on 1ac990a5 is the single-line fix commit `a5f202a8`, which updates `internal/backend/auth_interceptor_test.go` to pass the new 4th `allowRegistration` argument to `services.NewAuthServiceGorm(...)` after main's registration-gate change landed on this branch through a rebase. I checked it against every other call site in the tree (`server.go`, `registration_gate_test.go`, `update_timezone_test.go`, `auth_impersonation_test.go`) and the signature and argument order match exactly, so this is a mechanical fix, not a behavior change.

The feature's own files (`internal/webui/handlers/alert_related.go` and its tests, the dashboard filter-chain plumbing in `dashboard_handlers.go`, and the history/staleness fixes in `dashboard_modal.templ`) are byte-identical to the commit I already reviewed clean across three prior rounds — no further changes to verify there. Locally: `go build ./...` succeeds and `go test ./internal/backend/...` passes all 106 tests, confirming the interceptor test compiles and runs against the real signature.

No blocking or non-blocking findings from me this round. Nicely converged feature — thanks for grinding through all the QA rounds and keeping the rebases clean.



SoulKyu:
## 🧪 QA Report

Re-test after the rebase (requested by the rebaser). The PR's own 27 files are **byte-identical** to the previously-QA'd `69322ff` (`git diff 69322ff HEAD` touches only files from the new base: #172's gRPC auth interceptor, `config.go` mapstructure tags, `NOTIFICATOR_SERVICE_TOKEN`). So this pass re-ran the full battery on the **new base**, with the deny-by-default gRPC interceptor now in front of every backend RPC the panel's view state depends on (hidden alerts, presets, acknowledgements).

Env: `docker compose` stack rebuilt from this SHA — webui/backend images rebuilt locally and verified inside the image (`filterDashboardOnRelatedLabel` ×1, `appendFilterHiddenParams` in `dashboard_data/_modal/_utilities_templ.go`), webui :9081, fake Alertmanager :10093. Real browser (Playwright/Chromium) + raw HTTP. Peak population **2 409 alerts in the classic view**.

|---|---|---|
| `Related` on an alert that is the only one firing shows an explicit empty state | ✅ | `QAGlone` (alertname/qasolo/severity/job/instance/team all unique) → 0 groups, pane renders "No other alerts are currently firing with a matching label — this looks isolated." — screenshot confirms **no spinner and no empty box inside the pane** |
| During a storm: groups ordered by count, **counts match the dashboard when the same filter is applied** | ✅ | Panel = table in **34** same-instant configurations. Clean case: `qacluster=blast-B` 30=30, `severity` 47=47, `job` 38=38. Preset case (the old FAIL): exact rule 29=29, regex rule 20=20, disabled rule 30=30, rule on the correlating label itself 0=0, rule muting the **open** alert 29=29, empty `label_value` (match-all) 0=0, 50 rules 0=0, 7 preset-hidden fingerprints 23=23, 407 hidden fps (34 KB query) 23=23, fps+regex combined 15=15. Ordering count-desc with label-key-asc tiebreak (9/8/7/6/5) |
| Clicking a related alert opens it in the modal and **Back returns to the previous alert** | ✅ | A `QAEblast06` (depth 1) → Related → click B `QAEsub07` → depth 2, URL `/dashboard/alert/<B>`; **Back → A restored, modal still open** (`alertDetails.fingerprint == A`); Forward → B. 3-deep chain (depth 1→2→3) closed with ✕ → lands exactly on the pre-modal `/dashboard?search=QAMblast`. Deep link (depth 0) → related child (depth 1) → Back → deep-linked alert restored → close → `/dashboard`, never off-site |
| `Filter dashboard on this` closes the modal and leaves the table filtered on that label value | ✅ | Modal closes, URL `?label=qacluster%3Dblast-F`, chip `qacluster=blast-F`, `searchQuery` emptied, `displayMode` forced to classic, preset cleared. Panel 30 = table 30, **0 stray rows** (the `blast-F-c1` substring neighbours the old free-text search dragged in are gone). Survives reload; chip ✕ clears it (30 → 415) |
| Cache-only (no Alertmanager call, no backend RPC), sub-100ms on a few thousand alerts | ✅ | **2 409 active alerts**: 12 consecutive calls at 7.8 / 8.6 / 8.6 / 9.0 / 9.3 / 9.5 / 9.8 / 10.2 / 10.2 / 11.1 / 11.5 / 24.1 ms; 40 calls in 0.39 s, 60 calls in 0.60 s. Alertmanager access log over that window: **1** `GET /api/v2/alerts` — the 5 s background cache poll, none attributable to `/related`. Handler touches `GetLiveAlert` + `getStandardAlerts()`/`applyDashboardFilters` only |
| A unit test covers the correlation helper (degenerate labels dropped, ordering by count) | ✅ | `go test ./internal/webui/handlers/` → **72 pass**, incl. `TestComputeRelatedAlertGroups`, `…OrdersByCountDesc`, `…EmptyWhenAlone`, `…CapsAlertsPerGroup`, `…CountsOnlyTheClassicView`, `TestRelatedEndpointHonoursPresetHiddenRules`, `TestRelatedGroupCountMatchesLabelFilteredDashboard`. `node --test dashboard_modal_history.test.mjs` → 4/4. Full `go test ./...`, `go build ./...`, `go vet ./...` clean on the rebased base |

**Adversarial pass — what I threw at each criterion**

- **The rebase itself (new base = #172's deny-by-default gRPC interceptor).** The panel's population depends on backend RPCs (hidden alerts, saved presets, acknowledgements) that are now behind the interceptor. Re-ran every state-dependent case end-to-end: acknowledging 6 alerts through `bulk-action` (13 → 6, panel and table both), per-user hidden alerts (5 hidden → 25 vs 30), preset load + `Filter dashboard on this`, cross-user isolation. No RPC failed, no count diverged, no `Unauthenticated` in the backend log. `go test ./...` clean on the merged tree.
- **Boundary & empty** — isolated alert → empty state; a label shared with exactly **1** other alert → dropped; 30-sharer group → 20 listed + "Showing first 20 of 29 other alerts."; 7 correlating labels → exactly **5** groups, 9/8/7/6/5 desc, each matching its table; a label value covering ~100 % of the estate (`qaempty=""`, 6 alerts + every alert lacking the key) → dropped as degenerate, so the empty-value trap never reaches the URL filter; 2 409-alert estate.
- **Malformed & hostile** — fingerprints: empty (400), unknown, `../../etc/passwd`, `' OR 1=1--`, `%00`, `..%2f..%2fadmin`, `<script>alert(1)</script>`, a single space, `null`, 4 000 chars → 404 or gin's 404, **never a 500, never a stack**. Hidden-state params: `{{notjson`, a JSON object instead of an array, wrong types (`label_key:123`), `null`, `[null,null]`, `][`, invalid regex `([a-`, JS-only lookahead `QANblast(?=0)`, catastrophic `(a+)+$` → panel and table agree on all of them, ≤ 11 ms, no ReDoS (RE2). Attacker-influenceable **label value** `` renders through `x-text` → `window.__qaXss` undefined, `0` `img[src=x]`, the literal text is shown; it round-trips through `?label=…`, the chip, and a reload still inert, panel 7 = table 7. Value `k=v&x#y z%p+q/r?s` (`=`, `&`, `#`, space, `%`, `+`, `/`, `?`) → server `strings.Cut` on the first `=` matches the client parser, panel 8 = table 8. Hand-crafted `?label=` params: `''`, `'='`, `'=value'`, `'key'` are ignored; `?label=key=` filters on the empty value (matches alerts lacking the key) — not reachable from the panel, which drops such groups; cosmetic.
- **Error & unavailable path** — `/related` forced to **500** → red "Unable to load related alerts (HTTP 500)." + **Retry**, and Retry recovers (4 groups render); **503** → same; **200 with `success:false`** → the server's message + Retry; **network abort** → "check your connection and retry."; **404** → "This alert is no longer in the live cache, so its blast radius cannot be computed." In every case the empty state stays hidden — no blank pane sold as "nothing correlates". `docker compose stop backend` mid-flow with a live session → `/related` returns **503 "Authentication service unavailable"**, byte-identical to `/dashboard/data`; recovers on restart. A **401** on `/related` triggers the app-wide session handler (page reset), not a silent empty pane.
- **Auth & scope** — unauthenticated → 401; forged `session_id=forged-deadbeef` → 401; 512-char forged cookie → 401. Cross-user: `qa` hid 5 of a 30-alert group → its panel/table read 25/25 while `qa2` read 30/30 on the same alert. No leak either way; acknowledgements (backend-persisted) are correctly shared, per-session hides are not.
- **Adjacent flows** — the label filter combines with severity (`label` alone 30 → `+severity=critical` 5) and clears via `Clear All`, the chip ✕, and the amber hidden-state ✕ independently; a reload rebuilds it from the URL. `j`/`k`/`o` cursor navigation still opens the modal; the search box still accepts typing (incl. `/`, which is not a hotkey in this app). History and Comments still load after Related, and Related re-renders on switching back (4 groups, no error). The per-alert refetch guard from `361f295` holds: after A → B the payload fingerprint equals the open alert's. **Live updates with the label filter on**: injected 6 matching + 6 non-matching alerts → the matching ones arrived, **0 non-matching leaked** (the client predicate mirrors `alertMatchesLabelFilters`). Entering from the **acknowledged** view (whose rows are not in the counted population) → `Filter dashboard on this` forces classic, panel 6 = table 6.
- **State that outlives the cache / persisted state** — acknowledgements persisted in Postgres are honoured on the next poll (13 → 6, both sides), and `/related` on an *acknowledged* alert still agrees with the table in all 4 of its groups; no resolved or acknowledged alert ever appeared in a listed group (25 listed entries re-fetched individually, 0 leaks); per-user hidden alerts persist and are honoured; a **saved filter preset** persisted in the backend — the case that failed at `f0729cd` — flows into the count in all 11 rule variants tried.

Non-blocking, unchanged from the previous pass: a preset's hidden state is client-side only, so a reload after `Filter dashboard on this` drops it (table 16 → 23) *together with* the amber chip that explained it — the view stays internally coherent, and it is the pre-existing behaviour of any loaded preset, not something this PR introduces.

**Verdict: PASS**

PR #175, opened 2026-08-02
docs(spec): plan @mentions in alert comments (#164)

### LINKED ISSUE #164: feat(collaboration): @mentions in alert comments — pull a teammate into the alert they need to see
## User & need

Notificator is a *team* incident console: acks, comments and an activity feed already exist. But a
comment is a message nobody is told about. At 3am the on-call writes `this is the payment DB, @marie
owns it` in an alert comment — and Marie learns about it only if she happens to open that alert.

Today the only push channel is `notification_settings.templ`: browser/sound notifications for **new
alerts, filtered by severity**. There is zero notification for collaboration events. The workaround
is to leave the tool (Slack, phone) to say "look at alert X", which is exactly the context switch
the console is supposed to remove.

## Proposed feature

- Typing `@` in the comment box of the alert modal offers usernames; the stored comment keeps the
  plain `@username` text.
- Rendered comments highlight mentions, and the one addressed to *you* is visually distinct.
- A **Mentions** indicator in the top bar (next to the activity link) shows the count of comments
  mentioning you that you have not seen yet; clicking it opens the activity feed filtered to your
  mentions, each row deep-linking to its alert modal (the existing `/dashboard/alert/:id` route).
- If browser notifications are enabled, a new mention of you raises one — regardless of severity
  filters, because a human explicitly asked for you.

## Approach

Deliberately additive, no schema change: mentions are plain text inside the existing `comments`
table, detected by matching against the current user's username.

- **Detection / feed**: `GET /dashboard/activity` (`activity_handlers.go`) already returns comment
  events with `content`, `username` and `alertName`, and `matchesActivitySearch` already filters on
  content. Add a `mentionsUser` query flag (or reuse `search=@<me>`) so the feed can return only
  mentions; the current user comes from the session, so no new gRPC RPC and no `make proto`.
- **Poll**: the base layout polls the activity endpoint on the existing dashboard refresh cadence
  and feeds new mention events into `templates/scripts/notification_service.templ`, which already
  owns permission handling, sound and de-duplication.
- **Autocomplete source**: needs a non-admin "list usernames" endpoint —
  `ListUsersForImpersonation` is admin-only. Add a minimal authenticated
  `GET /api/v1/users/mentionable` returning `{id, username}` only. If that proves contentious,
  ship phase 1 without autocomplete (free-text `@name` still works end to end).
- **UI**: comment box and comment rendering live in `components/alert_modal_shared.templ` and
  `scripts/dashboard_modal.templ`; the header lives in `layouts/Base.templ`. Edit `.templ` sources
  and regenerate with `make webui-templates` — never touch `*_templ.go`.
- **Seen state**: a `lastSeenMentionAt` timestamp in `localStorage`. Deliberately per-browser: no
  new table, no new RPC. If teams ask for cross-device read state later, it becomes one column on
  the user preferences the backend already stores.

## Acceptance criteria

- Posting a comment containing `@bob` on an alert produces, for the user `bob` and no one else:
  a mention badge increment, a row in the mentions view, and (if browser notifications are enabled)
  one browser notification linking to that alert.
- The mention notification fires even when the alert's severity is excluded from the user's
  severity notification preferences.
- Clicking a mention opens the alert modal for the right alert, and the badge count drops.
- A comment that mentions nobody produces no notification and no badge change.
- Mention matching is case-insensitive and does not match a substring of a longer username
  (`@bob` must not match `bobby`) — covered by a unit test next to
  `handlers/activity_handlers_test.go`.
- Self-mentions do not notify the author.

## Out of scope

- Mentioning groups/teams (`@sre`), and mentioning users who have never logged in.
- Email, Slack or any external delivery — browser notification only, using the existing service.
- Editing a comment to add or remove a mention (comments are create/delete today).
- Cross-device read state for the mention badge.
- Mentions in acknowledgement reasons.

SoulKyu:
## Summary
- Adds `specs/2026-08-02-164-feat-collaboration-mentions-in.md`, a verified implementation plan for #164 (@mentions in alert comments).
- Grounds the issue's proposed approach against the current tree and corrects five inaccuracies: comment box/list location (`modal_components.templ`, not `alert_modal_shared.templ`), the missing global refresh loop to hook a badge poll into, the nonexistent `/api/v1/users` route group, insufficient `search=@<me>` reuse for word-boundary matching, and the alert-shaped notification path that can't bypass severity filters as-is.

## Test plan
- [ ] Spec reviewed for accuracy against the codebase (file:line references, function signatures)
- [ ] No code changes in this PR — plan-only

Closes #164

### DIFF
--- a/specs/2026-08-02-164-feat-collaboration-mentions-in.md
+++ b/specs/2026-08-02-164-feat-collaboration-mentions-in.md
@@ -37,7 +37,7 @@ phone), which is exactly the context switch the console should remove.
 
 ## Corrections to the issue's proposed approach
 
-Verified against the current branch; three of the issue's approach claims don't
+Verified against the current branch; several of the issue's approach claims don't
 match what's actually in the tree, so the plan below routes around them instead of
 building on them:
 
@@ -105,6 +105,24 @@ building on them:
    *bypass* severity filtering per the issue's own requirement. The plan adds a
    sibling method instead of stretching the alert-shaped one over a different
    event type.
+7. **There is no page-independent JS handle on the viewer.** `window.currentUser`
+   and `window.currentUserId` do not exist anywhere in the tree (0 occurrences
+   outside this document). The viewer's identity is reachable in exactly three
+   page-scoped ways: `dashboardUtilitiesMixin.getCurrentUser()`
+   (`internal/webui/templates/scripts/dashboard_utilities.templ:355`, populated by
+   `loadCurrentUser()` at :297) — dashboard only; `currentUserId`, server-injected
+   into the statistics Alpine root
+   (`internal/webui/templates/pages/StatisticsDashboard.templ:43`) — statistics
+   only; and `GET /api/v1/auth/me`. Likewise
+   `internal/webui/templates/scripts/dashboard_utilities.templ` defines a *mixin
+   object* (`window.dashboardUtilitiesMixin`, :5), not free functions, and is
+   included only by `NewDashboard.templ:1006` — nothing in it is reachable from
+   the statistics page. **Rule this plan follows:** cross-page JS helpers live in
+   `internal/webui/templates/layouts/Base.templ` (the only script host every page
+   renders — all 10 pages call `@layouts.Base`), and viewer identity is *passed in*
+   as an argument from a scope each call site actually has, never read from an
+   assumed global. The scope table in the Verification ledger enumerates every JS
+   symbol this plan names, where it is defined, and on which pages it exists.
 
 ## Approach
 
@@ -230,9 +248,20 @@ Only the live modal (`modal_components.templ`) gets this — the statistics
 resolved-alert comment view (`alert_modal_shared.templ:484-508`,
 `AlertModalCommentsWritable`) is also "writable" in principle but per the issue's
 scope this ships for the primary alert modal; the shared writable component can
-gain it later with the same pattern if wanted (no changes needed there for v1
-since it delegates to the same `addComment()`/`newCommentContent` the dashboard
-mixin provides).
+gain it later with the same pattern if wanted. No changes are needed there for v1.
+
+Note the provenance, because it matters for where a later autocomplete would have
+to be wired: `AlertModalCommentsWritable`'s `newCommentContent` / `addComment()`
+are **not** the dashboard mixin's. `@scripts.DashboardModal()`
+(`window.dashboardModalMixin`, `scripts/dashboard_modal.templ:5`, `addComment` at
+:245) is included only by `NewDashboard.templ:1007`. The statistics page's root
+scope is `{ ...statisticsDashboardPage(), ...statisticsViewsMixin(),
+currentUserId }` (`StatisticsDashboard.templ:43`), and `statisticsDashboardPage()`
+declares its own `newCommentContent`/`commentSubmitting`
+(`StatisticsDashboard.templ:1802-1803`) and its own `addComment()`
+(`StatisticsDashboard.templ:4359`). Two independent implementations, one shared
+markup component — so a later autocomplete there is a second wiring job, not a
+free ride on the dashboard one.
 
 ### 4. Frontend: mention highlighting in rendered comments
 
@@ -241,17 +270,55 @@ possible):
 - `modal_components.templ:1525` — `<p ... x-text="comment.content"></p>`
 - `alert_modal_shared.templ:467` — `<p ... x-text="comment.content"></p>`
 
-Switch both to `x-html` bound to a new global helper, e.g.
-`window.renderCommentHtml(content, viewerUsername)`, defined once in
-`internal/webui/templates/scripts/dashboard_utilities.templ` (next to the other
-small formatting helpers like `getCurrentUser`/`canDeleteComment`,
-`dashboard_utilities.templ:355-382`) and referenced from both templates as
-`x-html="renderCommentHtml(comment.content, getCurrentUser()?.username)"`
-(`alert_modal_shared.templ`'s simpler modal doesn't have `getCurrentUser()` in
-scope — for that one, pass `comment.content` and let the helper compare against
-`window.currentUser?.username` read internally, so both call sites work without
-threading extra Alpine state through a shared component's string-built
-expressions).
+Switch both to `x-html` bound to a new helper
+`window.renderCommentHtml(content, viewerUsername)`.
+
+**Where it lives.** In a new `<script>` block in
+`internal/webui/templates/layouts/Base.templ`, after the impersonation block
+(`Base.templ:110-305`, i.e. inserted at :306, before the footer at :307) and
+alongside the other page-independent globals already defined there
+(`window.darkModeState` :58, `window.impersonationState` :112). Not in
+`scripts/dashboard_utilities.templ`: that file is one `window.dashboardUtilitiesMixin`
+object literal (`:5`) whose members are Alpine *methods*, not callable globals, and
+it is included only by `NewDashboard.templ:1006` — a helper defined there is
+invisible to the statistics page, which is one of the two call sites. `Base.templ`
+is the only script host every page renders.
+
+**How each call site supplies the viewer.** The username is passed explicitly;
+the helper reads no globals. The two call sites below are the complete set
+(`AlertModalCommentsReadonly` is reached only via `AlertModalCommentsWritable`,
+`alert_modal_shared.templ:486`, itself used only at `alert_modal_shared.templ:748`
+inside `AlertModalReadonly`, whose only caller is
+`StatisticsDashboard.templ:1690`):
+
+- `modal_components.templ:1525` →
+  `x-html="renderCommentHtml(comment.content, getCurrentUser()?.username)"`.
+  `getCurrentUser()` is `dashboardUtilitiesMixin`'s
+  (`dashboard_utilities.templ:355`) and is in scope here — the same scope already
+  calls `canDeleteComment(comment)` two lines down (`modal_components.templ:1532`),
+  another member of that mixin (`dashboard_utilities.templ:376`).
+- `alert_modal_shared.templ:467` →
+  `x-html="renderCommentHtml(comment.content, currentUsername)"`. That symbol does
+  not exist yet: add it to the statistics root `x-data`
+  (`StatisticsDashboard.templ:43`) next to the `currentUserId` already injected
+  there the same way — `currentUsername: '` + `data.User.Username` + `'`.
+  `StatisticsDashboardData.User` is a `pages.ProfileUser` with a `Username` field
+  (`Profile.templ:15-24`) that the handler already fills
+  (`statistics_handlers.go:457-462`), so this is a template-only change: no
+  handler, model, or route edit. Alpine's nested
+  `x-data="{ currentTab: 'overview' }"` (`alert_modal_shared.templ:604`) inherits
+  the root scope, so `currentUsername` resolves inside the modal.
+
+No new `templ` parameter is threaded through `AlertModalReadonly` /
+`AlertModalCommentsWritable` / `AlertModalCommentsReadonly`: the whole chain has a
+single render path (enumerated above) rooted in the statistics page, so the bare
+`currentUsername` reference is unambiguous. If a second page ever renders
+`AlertModalReadonly`, *that* is the moment to add a `viewerVar string` parameter —
+adding it now would be a parameter with one possible value.
+
+The helper tolerates a missing second argument (`''`/`undefined` → every mention
+renders as a neutral chip, no `mention-self`) so a future third call site degrades
+to "no self-highlight" instead of throwing inside an Alpine expression.
 
 The helper:
 1. HTML-escapes `content` (manual `textContent`-via-`document.createElement`
@@ -340,9 +407,17 @@ function mentionsBadge() {
 }
 ```
 
+The badge resolves its own user via `GET /api/v1/auth/me` rather than a
+server-injected value (the route this plan uses everywhere else) because
+`PageNavigator(activePage string)` (`PageNavigator.templ:5`) takes no page data,
+and one of its four hosts — `NewDashboardContent()` (`NewDashboard.templ:13`,
+navigator at `:123`) — takes no page data at all. The `userId` only keys
+`localStorage`, so the one-request-per-page-load cost buys per-user read state; a
+failed fetch degrades to the `'anon'` key rather than breaking the badge.
+
 Rendered on all four pages that include `PageNavigator`
-(`StatisticsDashboard.templ`, `Activity.templ`, `Silences.templ`,
-`NewDashboard.templ`), 30s poll matching the existing `Activity.templ:282` and
+(`StatisticsDashboard.templ:62`, `Activity.templ:89`, `Silences.templ:34`,
+`NewDashboard.templ:123`), 30s poll matching the existing `Activity.templ:282` and
 `Base.templ:262` cadences. `window.notificationService` only exists when
 `NewDashboard.templ:998`'s `@scripts.NotificationService()` has run — i.e. only
 while the dashboard page is open — so `notifyNew`'s guard means mention *browser
@@ -413,12 +488,21 @@ naive push into the alert-shaped queue.
   `mentionsOnly` field; when true, default `windowMinutes = 10080` before the
   first `load()` (so a mention from yesterday isn't hidden by the normal 60-minute
   default).
+- The seen-marking below needs the viewer's id, and this page does not expose one
+  today (`x-data="activityPage()"`, `Activity.templ:74`). Inject it server-side
+  rather than adding a global: change that attribute to
+  `x-data={ "{ ...activityPage(), currentUserId: '" + data.User.ID + "' }" }`,
+  byte-for-byte the pattern `StatisticsDashboard.templ:43` already uses (spread +
+  injected id). `ActivityContent(data)` already receives `data`
+  (`Activity.templ:12`) and the handler already fills `User`
+  (`activity_handlers.go:173-175`), so no handler change — and no second
+  `/api/v1/auth/me` round-trip on this page.
 - `load()` (`Activity.templ:286-314`): when `mentionsOnly`, add
-  `params.set('mentionsUser', '1')` and mark `localStorage` seen (same
-  `storageKey('seen')` computation as `mentionsBadge()`, keyed by the same
-  `notificator_mentions_seen_<userId>` — fetch `/api/v1/auth/me` once in `init()`
-  here too, or read `window.currentUserId` if the page already exposes it) so
-  returning to the badge shows 0 without waiting for the next 30s poll.
+  `params.set('mentionsUser', '1')` and mark `localStorage` seen, using the same
+  key `mentionsBadge()` writes — `'notificator_mentions_seen_' + (this.currentUserId || 'anon')`
+  — so returning to the badge shows 0 without waiting for the next 30s poll. Both
+  sides must derive the key from the same expression; if they disagree the badge
+  never clears, so keep the `'anon'` fallback identical too.
 - Add a small "Showing your mentions — Clear" chip above the filter bar when
   `mentionsOnly` is true, clicking clears the query param (`history.replaceState`)
   and reloads with normal defaults — cheap escape hatch back to the full feed.
@@ -449,9 +533,17 @@ naive push into the alert-shaped queue.
 - `go test ./internal/webui/handlers/... -run 'TestMatchesMention|TestMatchesActivitySearch|TestBuildActivityFeed'`
   — new + existing activity tests green.
 - `make webui-templates` after editing every `.templ` file touched above
-  (`modal_components.templ`, `alert_modal_shared.templ`, `dashboard_utilities.templ`,
-  `notification_service.templ`, `PageNavigator.templ`, `Activity.templ`) — never
-  hand-edit the generated `*_templ.go` files.
+  (`layouts/Base.templ`, `components/modal_components.templ`,
+  `components/alert_modal_shared.templ`, `components/PageNavigator.templ`,
+  `pages/StatisticsDashboard.templ`, `pages/Activity.templ`,
+  `scripts/notification_service.templ`) — never hand-edit the generated
+  `*_templ.go` files.
+- Browser check that costs nothing and would have caught the class of bug this
+  plan had to be corrected for twice: on **both** `/dashboard` and `/statistics`,
+  evaluate `typeof window.renderCommentHtml` → `"function"`, and confirm a comment
+  mentioning the logged-in user renders a `mention-self` chip on each page (the
+  dashboard modal and the statistics resolved-alert modal use *different* Alpine
+  scopes; a helper that works on one proves nothing about the other).
 - `make test` (docker-compose full stack) manual pass:
   1. As user `alice`, comment `hi @bob check this` on an alert. As `bob` (second
      browser/session), confirm: badge increments within 30s, `/activity?mentions=1`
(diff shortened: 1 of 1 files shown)


specs/2026-08-02-164-feat-collaboration-mentions-in.md
```diff
@@ -0,0 +1,497 @@
...
+   badge.** `internal/webui/templates/layouts/Base.templ` (the layout wrapped
+   around every page) has no polling loop. The adaptive refresh loop that exists
+   (`startAutoRefresh`/`loadDashboardIncremental`,
+   `internal/webui/templates/scripts/dashboard_core.templ:597-602`) only runs on
+   the dashboard page's Alpine component — Silences, Statistics and Activity don't
+   have it, and the top bar with the Activity tab
+   (`internal/webui/templates/components/PageNavigator.templ`) renders on all four
+   pages (`StatisticsDashboard.templ`, `Activity.templ`, `Silences.templ`,
+   `NewDashboard.templ`). The badge therefore needs its own small, page-independent
+   poll living in `PageNavigator.templ`, not a hook into the dashboard's loop.
```
SoulKyu, line 74: **"Base.templ has no polling loop" overclaims — it does have one, and the doc contradicts itself about this elsewhere.**

Problem: this section states as fact that `internal/webui/templates/layouts/Base.templ` "has no polling loop," and the Verification ledger further down cites having "read `Base.templ` in full" to back that claim up.

Why it matters: it's the load-bearing justification for building an entirely new poll in `PageNavigator.templ` (section 5) instead of considering the existing pattern. `Base.templ` already contains `connectedUsersDropdown()` (~lines 241-274), which runs `this.pollInterval = setInterval(() => this.refresh(), 30000)` at line 262 — a small, page-independent poll living in the layout wrapped around every page, polling `/api/admin/connected-users` for the admin connected-users badge. That's exactly the shape this section claims doesn't exist. The spec even cites this same code elsewhere (`Base.templ:210-232`/`:262` in section 3, as the "fetch once, filter client-side" pattern for the mention-autocomplete cache) as an existing 30s poll — so the doc says "no polling loop" here and treats it as an existing pattern to reuse two sections later.

Evidence: `Base.templ:241-274` (`connectedUsersDropdown`), `Base.templ:262` (`setInterval(() => this.refresh(), 30000)`), vs. this section's unqualified "has no polling loop."

Suggested fix: narrow the claim, e.g. "Base.templ's only existing poll (`connectedUsersDropdown`, :241-274) is admin-gated and fetches connected-users data, unrelated to activity/mentions — there's no *dashboard-refresh-shaped* loop to hook a mentions badge into, so it still needs its own poll." This keeps the proposed architecture (separate poll in `PageNavigator.templ`) intact, it just stops asserting something the tree doesn't support.


SoulKyu (commented):
@SoulKyu — reviewed this spec-only PR (it adds `specs/2026-08-02-164-feat-collaboration-mentions-in.md`, no code changes). I independently re-verified all 23 file:line claims the spec makes against the current tree — comment tab location in `modal_components.templ`, the `router.go` route groups, `AuthService.ListUsers`'s session-only check, `matchesActivitySearch`'s substring behavior, `notification_service.templ`'s severity gating, the `Comment`/`ActivityEvent` field shapes, etc. — and 22 of them check out exactly as described. Nice grounding work overall, and the corrections to the original issue (comment markup location, missing `/api/v1/users` group, `strings.Contains` vs word-boundary matching) are accurate and genuinely useful for whoever implements this next. One inline note below on a factual slip in the "no polling loop" claim that's worth tightening before this becomes the implementation source of truth — doesn't change the proposed approach, just the stated justification for it.



specs/2026-08-02-164-feat-collaboration-mentions-in.md
SoulKyu: @SoulKyu This has been fixed in c71ee93.

Reworded the claim at specs/2026-08-02-164-feat-collaboration-mentions-in.md:61-64 to acknowledge Base.templ's existing connectedUsersDropdown poll (admin-gated, :242-274, setInterval at :262) instead of denying any loop exists, and updated the matching Verification ledger entry at line 474-476 for consistency.




SoulKyu:
**Looper fixer round complete** — c71ee93

- ✅ Review comment on `specs/2026-08-02-164-feat-collaboration-mentions-in.md:74` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/175#discussion_r3699907088)
  - Reworded the claim at specs/2026-08-02-164-feat-collaboration-mentions-in.md:61-64 to acknowledge Base.templ's existing connectedUsersDropdown poll (admin-gated, :242-274, setInterval at :262) instead of denying any loop exists, and updated the matching Verification ledger entry at line 474-476 for consistency.


SoulKyu:
## 🧪 QA Report

Spec/docs-only PR on a `looper/planner/*` branch → graded as a **spec-accuracy review**, not against #164's acceptance criteria (the feature is intentionally unimplemented here). Every codebase reference was checked at `c71ee93`, and the runtime-checkable claims were exercised against the booted stack (`make test`, webui :9081, backend :9080/:51051).

|---|---|---|
| Correction 1: comment box/list is in `modal_components.templ`, not `alert_modal_shared.templ` | ✅ | Line-exact: comments tab `1434`, textarea+`x-ref="commentInput"` `1449-1453`, `x-text="comment.content"` `1525`, delete `1532`; second renderer at `alert_modal_shared.templ:454-508`, `x-text` at `467`. Repo-wide grep finds exactly these two `x-text="comment.content"` sites. |
| Correction 2: `dashboard_modal.templ` is JS-only (behavior yes, markup no) | ✅ | 0 occurrences of `<div>`/`<template>`/`<textarea>`; `addComment` `245`, `deleteComment` `294`, `refreshComments` `332`. |
| Correction 3: no dashboard-refresh-shaped loop for a global badge | ✅ | `Base.templ` sole poll = `connectedUsersDropdown` (`242`, `setInterval(…30000)` `262`, fetches `/api/admin/connected-users` `276`, admin-gated); `startAutoRefresh` `dashboard_core.templ:598` is Alpine-scoped to the dashboard; `PageNavigator` renders on exactly the 4 named pages. |
| Correction 4: no `/api/v1/users` group; endpoint belongs on `dashboard` group | ✅ | Live: `GET /api/v1/users` → **404**, `/api/v1/users/mentionable` → **404**, `/api/v1/dashboard/mentionable-users` → **404**, `/api/v1/dashboard/activity` → **401** (exists, auth-gated). `dashboard.GET("/activity")` is at `router.go:304`, exactly where the spec says to insert. |
| Correction 4b: `AuthService.ListUsers` RPC has **no** admin check; the gate is handler-side | ✅ | Live gRPC as **non-admin bob** (session only) → returned all 3 users + `totalCount`. Same user against the webui handler `/api/impersonate/users` → **403 "You are not allowed to impersonate users"**. Exactly the contrast the spec's design relies on. `BackendClient.ListUsers(sessionID, limit, offset)` at `backend_client.go:1069` matches the proposed call verbatim. |
| Correction 5: `search=` uses `strings.Contains`, so `@bob` matches `@bobby` | ✅ | **Live-proven**: alice posted `@bobby please check this cert`; `GET /activity?search=%40bob` returned that event (1 hit). The stated acceptance-criterion violation is real, so a dedicated matcher is genuinely required. |
| Correction 6: `shouldNotify`/`showNotificationImmediate` are alert-shaped and can't be reused | ✅ | `shouldNotify` `410` gates on `enabledSeverities` `433` + `matchesFilterPreset` `438`; `showNotificationImmediate` `669` re-calls `shouldNotify` at `674` (so it truly cannot be reused for a filter-bypassing event); `processNotificationQueue` `652` does call `showNotificationImmediate` unconditionally at `659`, confirming the queue-branch fix is needed. |
| Proposed `matchesMention` regex actually meets the criteria | ✅ | Ran the spec's Go function verbatim over 15 adversarial cases — all pass: `@bobby`→no match, `user@bob.com`→no match, `@bob.smith`/`@bob_smith`→no match, case-insensitive, empty username, newline/paren/punctuation boundaries, and `QuoteMeta` proven (username `a.b` does not match `@axb`). |
| Spec's design targets exist at runtime (`windowMinutes`, notification helpers, event shape) | ✅ | Live `windowMinutes=10080` → **200** (code comment confirms no upper bound); event JSON keys are exactly `id, alertKey, kind, username, content, createdAt`; `getNotificationIcon('info')`→`info-icon.png`, `playNotificationSound`, `recordNotification`, `canShowNotification`, `notificationQueue` all exist. |
| Ledger claims: `/auth/me` shape, `Comment` shape, 1000-char cap, 200-event risk | ✅ | Live `/api/v1/auth/me` → `data.user.{id,username,email}`; comment of 1001 chars → **400 "cannot exceed 1000 characters"**, 1000 chars → **200**; the 200 cap is `ORDER BY comments.created_at DESC LIMIT 200` (`gorm_db.go:405-406`, `activityMaxLimit=200`), so the spec's "oldest dropped first, fresh mention never dropped" risk note is correct. |
| §4: shared helper can read the viewer's username from `window.currentUser?.username` | ❌ | **`window.currentUser` does not exist anywhere in the repo** (0 hits repo-wide) and is `undefined` at runtime on **both** `/dashboard` and `/statistics`. Following §4 as written, `mention-self` never applies in the `alert_modal_shared.templ` renderer — silently, no error — so Goal 2 ships half-done in the one call site the spec explicitly says "Both need the highlighting change". |
| §3: `AlertModalCommentsWritable` delegates to "the **dashboard mixin**'s" `addComment()`/`newCommentContent` | ❌ | Wrong provenance: the statistics page scope is `{...statisticsDashboardPage(), ...statisticsViewsMixin(), currentUserId}` (`StatisticsDashboard.templ:43`), which defines its **own** `newCommentContent` (`1802`) and `addComment()` (`4359`). Live: `window.dashboardModalMixin` is `undefined` on `/statistics`. The conclusion ("no changes needed there for v1") still holds — attribution only. |

### Detail on the blocking finding

Spec §4 instructs, for the `alert_modal_shared.templ` call site:

> *"let the helper compare against `window.currentUser?.username` read internally, so both call sites work"*

Its premise is right — I confirmed `getCurrentUser` **is** out of scope there (`typeof d.getCurrentUser === "undefined"` on the live statistics Alpine root). But the proposed replacement is a global that was never established. Live Alpine root scope on `/statistics`:

```json
{ "currentUserId": "ihazk55e0v7us9z1omrn4wdlp5numjh9",
  "typeof_getCurrentUser": "undefined",
  "typeof_addComment": "function", "has_newCommentContent": true }
globals: { "window_currentUser": "undefined",
           "window_notificationService": "undefined",
           "window_dashboardModalMixin": "undefined" }
```

The page does expose a viewer handle — `currentUserId` — but it is an **ID**, while the spec's helper compares by **username**, so it is not a drop-in either. The fix is small (have the helper fall back to a username the page actually exposes, or state that `window.currentUser` must be set in `Base.templ` as part of this work). Notably this claim is the one §4 assertion **absent from the spec's own Verification ledger** — every other reference I checked and that ledger lists came back exact.

**Adversarial pass** — what I tried to break, and what happened:

- **Boundary/empty:** ran the spec's `matchesMention` over empty username, single-char boundaries, start/end-of-string anchors, newline-adjacent mentions, glued mentions (`@alice@bob`, `@bob@alice`) — behaves as specified, no crash, no false positive.
- **Malformed & hostile:** username containing regex metacharacters (`a.b`) — `QuoteMeta` holds, `@axb` correctly does not match, so an attacker-chosen username can't turn into a wildcard. Oversized comment (1001 chars) → clean 400. Checked the §4 escape-then-match ordering for XSS: escaping first and *then* injecting the `<span>` is the correct order (matching after escaping would let the injected markup be escaped, and matching before would let hostile comment text survive into `x-html`) — the spec got this right.
- **Error/unavailable path:** `/api/v1/dashboard/activity` unauthenticated → 401, not an empty 200 (so the badge poll can't render "0 mentions" as a fake success).
- **Auth & scope:** the decisive one — called `AuthService.ListUsers` as **non-admin bob** with only a session, and it returned every user. This *confirms* the spec (no admin gate on the RPC) but also means the new endpoint deliberately exposes the roster to any authenticated user; the spec's trimmed `{id, username}` projection correctly drops the `email` the RPC returns. Verified `/api/impersonate/users` still 403s for bob, i.e. the spec's plan doesn't propose weakening the impersonation gate.
- **Adjacent flows:** confirmed the existing `search=`, `scope=mine`, and `kinds=` filters are applied *after* feed resolution, and the spec's insertion point (after the `kinds` block `143-151`, before `available` `153-158`) is exactly correct and would not disturb them.
- **Persisted/pre-existing state:** posted comments *before* any mentions feature exists and queried them back through the live activity endpoint — the spec requires no schema change and no migration, and stored comments are plain text (`@bobby …` round-tripped verbatim), so pre-existing rows work under the plan.
- **Where the attack landed:** probing the *runtime existence* of every JS global the spec names. `window.notificationService` — real (dashboard only, exactly as documented). `window.dashboardModalMixin` — real, but not in scope where §3 implies. `window.currentUser` — does not exist. That is the failure above.

**Verdict: FAIL**

One material defect (§4's nonexistent global, which would silently ship half of Goal 2) plus one incorrect attribution (§3). Everything else — all six corrections, the proposed matcher, the endpoint design, the risk analysis, and ~38 line-level references — checked out exactly, several proven against the running app rather than just read. This is a narrow fix, not a rewrite.

<sub>Stack torn down (`docker compose down -v`). Ports: webui 9081 / backend 9080 / gRPC 51051 (QA slot 1).</sub>

SoulKyu:
## 🧪 QA Report

Spec/docs-only PR on a `looper/planner/*` branch → graded as a **spec-accuracy review**, not against #164's acceptance criteria. Re-verified from scratch at `6263b04` against the booted stack (webui :10081, backend :10080/:52051, alertmanager :11093). The new commit rewrites §4/§7 to fix the previous round's `window.currentUser` finding — that fix is correct, but it introduces a new defect in its replacement.

|---|---|---|
| Correction 7 *(new)*: `window.currentUser`/`window.currentUserId` exist nowhere | ✅ | `rg` under `internal/` → **0 hits**. Live: `typeof window.currentUser` / `window.currentUserId` = `"undefined"` on **all four** pages (`/dashboard`, `/statistics`, `/activity`, `/silences`). |
| Correction 7: viewer identity reachable in exactly 3 page-scoped ways | ✅ | `dashboardUtilitiesMixin.getCurrentUser()` `dashboard_utilities.templ:355` (mixin `:5`, filled by `loadCurrentUser()` `:297` from `/api/v1/auth/profile`); `currentUserId` at `StatisticsDashboard.templ:43`; `GET /api/v1/auth/me` → **200**. |
| §4 *(new)*: `Base.templ` is the only script host every page renders | ✅ | 10 `layouts.Base` hits in `pages/*.templ`, one per page. Live: `window.darkModeState`+`window.impersonationState` = `object` on all 4 pages, while `dashboardUtilitiesMixin`/`dashboardModalMixin`/`notificationService` are `undefined` on 3 of 4. Insertion point checked for conditionals: the impersonation block `:110-305` sits directly in `<body>` with **no** wrapping templ `if`/`for`, so `:306` is unconditional — "page-independent by construction" holds. |
| §4 *(new)*: dashboard call site — `getCurrentUser()?.username` is in scope at `modal_components.templ:1525` | ✅ | Live, comment node inside the open alert modal: `typeof getCurrentUser` = `"function"`, `typeof canDeleteComment` = `"function"`, and evaluating **the spec's exact expression** `getCurrentUser()?.username` → `"alice"`. `<textarea x-ref="commentInput">` (§3's target) also present. |
| §4 *(new)*: nested modal `x-data` inherits the statistics root scope | ✅ | Live on `/statistics`: the nested `x-data="{ currentTab: 'overview' }"` (`alert_modal_shared.templ:604`) resolves the root's `currentUserId` → `1lsmkecy...`, while keeping its own `currentTab: 'overview'`. So a `currentUsername` added at `:43` **would** resolve inside `AlertModalCommentsReadonly`. |
| §4 *(new)*: single render path justifies a bare scope reference over a templ param | ✅ | `AlertModalCommentsReadonly` def `:454`, used only `:486` (inside `AlertModalCommentsWritable` `:484`), which is used only `:748` (inside `AlertModalReadonly` `:602`), whose only page call site is `StatisticsDashboard.templ:1690`. Repo-wide grep confirms no other reachability. |
| §4 *(new)*: **inject `data.User.Username` into the root `x-data` "the same way" as `currentUserId`** | ❌ | **Breaks the page for any username containing `'`.** See detail below — reproduced live end-to-end. |
| §3 *(new)*: statistics comment state is the page's own, not the dashboard mixin's | ✅ | `StatisticsDashboard.templ:43` root scope, own `newCommentContent`/`commentSubmitting` `:1802-1803`, own `addComment()` `:4359`; `@scripts.DashboardModal()` only at `NewDashboard.templ:1007`. Live on `/statistics`: `window.dashboardModalMixin` = `undefined`, root `typeof addComment` = `"function"`. Provenance now stated correctly. |
| §7 *(new)*: inject `currentUserId` on `/activity` from `data.User.ID`; key must match `mentionsBadge()`'s | ✅ | `Activity.templ:74` is `x-data="activityPage()"`; `ActivityContent(data)` receives `data`; `ActivityPage` fills `User` from `GetCurrentUserFromContext`. Key consistency **verified live**: `/api/v1/auth/me` `data.user.id` = `1lsmkecy3g8fup0po7it6f5jbymxnswx` = the `currentUserId` rendered into `/statistics`' `x-data` — same source, so `notificator_mentions_seen_<id>` agrees on both sides. Value is a system-generated ID, so the injection is safe here. |
| §5: `PageNavigator` takes no page data; 4 call sites; `/auth/me` fallback justified | ✅ | `PageNavigator(activePage string)` `:5`; call sites line-exact at `NewDashboard.templ:123`, `StatisticsDashboard.templ:62`, `Activity.templ:89`, `Silences.templ:34`; `NewDashboardContent()` `:13` takes no data. Live: nav present on all 4 pages. |
| Corrections 1–3 (markup locations, `dashboard_modal.templ` JS-only, no global refresh loop) | ✅ | Line-exact at this SHA: comments tab `1434`, textarea `1449-1453`, `x-text="comment.content"` `1525`, delete `1532`; second renderer `alert_modal_shared.templ:454-508`, `x-text` `:467`. `Base.templ`'s only poll is `connectedUsersDropdown` (`:242`, `setInterval(…30000)` `:262`, admin-gated `/api/admin/connected-users`); `startAutoRefresh` `dashboard_core.templ:598` is dashboard-Alpine-scoped. |
| Correction 4: no `/api/v1/users`; endpoint belongs on the `dashboard` group | ✅ | Live: `/api/v1/users` → **404**, `/api/v1/users/mentionable` → **404**, `/api/v1/dashboard/mentionable-users` → **404**, `/api/v1/dashboard/activity` → **200** authed / **401** unauth. `dashboard.Use(RequireAuth())` at `router.go:255`; `/dashboard/alert/:id` page route → **200**. `BackendClient.ListUsers(sessionID, limit, offset) ([]*User, int32, error)` at `:1069` matches the spec's `users, _, err := …(sessionID, 1000, 0)` verbatim; `AuthService.ListUsers` `services.go:361` validates session only — the admin gate is handler-side (`impersonation_handlers.go:144`). |
| Correction 5: `search=` is `strings.Contains`, so `@bob` matches `@bobby` | ✅ | **Re-proven live at this SHA**: alice commented `@bobby please check this cert` on `7bcc2ef5…`; `GET /activity?windowMinutes=10080&search=%40bob` → **1 event**, that comment. The dedicated word-boundary matcher is genuinely required. |
| Correction 6: `shouldNotify`/`showNotificationImmediate` are alert-shaped | ✅ | `shouldNotify` `:410`, `canShowNotification` `:632`, `processNotificationQueue` `:652-666` (loop calls `showNotificationImmediate(alert)` unconditionally → the `__mention` branch §6 adds is genuinely needed), `showNotificationImmediate` `:669`. |
| Proposed `matchesMention` Go matcher meets goal 5 | ✅ | Ran the spec's function verbatim over **23** adversarial cases, all correct: `@bobby`→no, `user@bob.com`→no, `@bob.smith`/`@bob_smith`/`@bob-smith`→no, `@bob0`→no, `...@bob`→no, `a@bob`→no, empty username→no, case-insensitive yes, newline/paren/tab/`!`/`@@` boundaries yes, and `QuoteMeta` proven (username `a.b` does not match `@axb`). |
| §4's JS token regex is equivalent to the backend matcher | ✅ | Ran `/(?:^|[^\w.-])@([A-Za-z0-9_.\-]+)/g` over the same 22-case corpus: **0 divergences** from the Go matcher (`[^\w.-]` ≡ `[^A-Za-z0-9_.\-]`). |
| §4's escape-then-highlight order is XSS-safe | ✅ | Ran the described helper against 6 payloads (``, `<script>`, `"><svg onload>`, `@`, pre-escaped `&amp;`/`&lt;`): all inert, only the `@token` becomes a `<span>`. Missing/`undefined`/`''` viewer → neutral chip, no throw, exactly as §4 promises; empty content → `""`. |
| Ledger claims (`/auth/me` shape, `Comment` shape, 1000-char cap, 200-event ceiling) | ✅ | Live `/api/v1/auth/me` → `data.user.{id,username,email}`; comment of **1001** chars → **400 "cannot exceed 1000 characters"** (`dashboard_handlers.go:1547`); `GetRecentActivity(sessionID, since, 200)` at `activity_handlers.go:95`; `ActivityEvent` JSON keys are exactly `id, alertKey, kind, username, content, createdAt`; response payload key is `events`; `windowMinutes=10080` accepted. |

### Detail on the blocking finding

§4 says, for the statistics call site:

> *add it to the statistics root `x-data` (`StatisticsDashboard.templ:43`) next to the `currentUserId` already injected there **the same way** — `currentUsername: '` + `data.User.Username` + `'`*

The existing `currentUserId` injection is safe only because a user ID is a system-generated alphanumeric. `data.User.Username` is **user-controlled** and unvalidated — the register handler checks `username == ""` and nothing else (`handlers.go:176`). Live: registering `o'brien` → **HTTP 201**. `x-data` is a templ attribute, so the value goes through `templ.EscapeString` = `html.EscapeString` — HTML escaping, which does **not** make a value safe inside a JS string literal. §4 would be the first place a username enters an `x-data` (`rg 'x-data=\{.*User\.Username'` → 0 existing hits), so the "same pattern" argument is exactly where it breaks.

Reproduced end-to-end in a real browser on the live `/statistics` page:

```
username stored:   o'brien                        (register → HTTP 201)
server emits:      currentUsername: &#39;o&#39;brien&#39;
browser decodes:   currentUsername: 'o'brien'
Alpine:            PAGEERROR: Unexpected identifier 'brien'
                   PAGEERROR: currentUsername is not defined
component init:    false          (control with 'alice' → renders "alice")
```

Because this is the **root** `x-data` at `:43`, the failure is not limited to the mention chip — `statisticsDashboardPage()` and `statisticsViewsMixin()` never initialize, so the whole statistics page (charts, filters, the resolved-alert modal) is dead for that user. And with a crafted username the expression silently executes:

```
username:  x', qaPwned: (window.__QA_PWNED=1), z: '
result:    window.__QA_PWNED === 1  →  true      (no page error at all)
```

Scope is self-inflicted (the injected username is the viewer's own), so this is self-XSS rather than a cross-user hole — but the **page-break** case needs no attacker at all: `o'brien` is an ordinary name. Fix is one line — `templ.JSONString(data.User.Username)` (exists in the pinned `templ@v0.3.906`), or inject via `templ.JSONScript` / a `data-` attribute read by the helper, instead of string-concatenating into the expression. §7's `currentUserId` needs no change.

### Non-blocking notes for the implementer

- **`Activity.templ:12`** is cited for "`ActivityContent(data)` already receives `data`". `ActivityContent` is at `:73`; `:12` is `templ Activity(data ActivityData)`. Both take `data`, so it does not misdirect — but it is the one line number in the document that is off.
- **Impersonation**: §1 filters the feed with `middleware.GetEffectiveUser` (correct — matches the existing `scope=mine` at `activity_handlers.go:131-132`), while §5/§7 key `localStorage` off `/auth/me` / `data.User.ID`, which are `GetCurrentUserFromContext` (the **real** user). An admin who impersonates someone sees that person's mentions but writes the "seen" cutoff under their own key, clobbering their own badge afterwards. Not claimed by the spec; worth one sentence in §5.
- **Clock skew**: the badge compares server `createdAt` against a client `Date.now()` cutoff; a fast client clock suppresses fresh mentions. Standard, but the spec's "badge shows 0" wording assumes agreement.

**Adversarial pass** — the break-it cases actually run, per area:

- **Malformed / hostile input into the new `x-data` injection (§4)** — username with `'` (`o'brien`), and a username crafted as a JS expression breakout (`x', qaPwned: …, z: '`). Both accepted by registration (HTTP 201); both reproduced against the live `/statistics` Alpine runtime through the real HTML parser (entity-decode path), one yielding a dead root component, the other silent code execution. **→ flipped §4 to ❌.** The quote-free control (`alice`) rendered correctly, isolating the cause.
- **Boundary / empty on the proposed matchers** — empty username, empty content, `@` alone, token at string start/end, adjacent `@@`, tab/newline/paren/apostrophe/punctuation boundaries, `@bob0`, `...@bob`, `a@bob`, email-shaped `user@bob.com`, `QuoteMeta` regex-metachar username (`a.b` vs `@axb`), case folding both directions. 23 Go cases + 22 JS cases, **no incorrect verdicts**; back-to-back `@a@b` behaves identically in both engines (neither matches), so the "same pattern" claim survives.
- **XSS / escaping order (§4 helper)** — 6 injection payloads through escape-then-regex, including pre-escaped entities (`&amp;`, `&lt;`) to test double-escaping and payloads placed adjacent to the `@token` so the `<span>` splice is the seam. All inert. Also probed the degradation contract: absent / `undefined` / `''` viewer argument — neutral chip, no throw inside the Alpine expression, as §4 promises.
- **Scope claims tested where they'd silently no-op** — this is the class the spec was corrected for twice, so every symbol in the scope table was resolved on the live page rather than grepped: mixins probed on all four pages (dashboard-only ones confirmed `undefined` on the other three), Base.templ globals confirmed `object` on all four, `window.currentUser`/`currentUserId` confirmed `undefined` everywhere, and the two §4 call-site expressions evaluated against real Alpine scopes (`getCurrentUser()?.username` → `"alice"`; nested modal scope → root `currentUserId`). Also checked the `Base.templ:306` insertion point is not inside a templ `if`/`for` that would make "all pages" false.
- **Auth & scope** — `/api/v1/dashboard/activity` unauthenticated → **401**; the three endpoints the spec asserts absent → **404** each, so §2's routing rationale is not built on a phantom route; confirmed the `ListUsers` RPC's missing admin check is real and the gate is handler-side, which is the whole basis for §2's "no new RPC" conclusion.
- **Error / unavailable path** — oversized comment (1001 chars) → **400** with the exact message the ledger quotes, 1000 → accepted; malformed fingerprint → **404 "Alert not found"** rather than a silent success.
- **Adjacent flows** — verified §4's change target does not disturb the neighbours it shares scope with: `canDeleteComment(comment)` still resolves in the same comment scope, `comment` is in scope, `x-ref="commentInput"` textarea still present for §3; and that `AlertModalCommentsReadonly` truly has no second render path that a bare `currentUsername` reference would break.
- **Persisted state** — key derivation compared across the two independent producers (`/auth/me` vs server-injected `data.User.ID`) on a live session: identical value, so the `notificator_mentions_seen_*` key agrees. Pushed further into impersonation, where they diverge — reported as a non-blocking note since the spec makes no impersonation claim.

**Verdict: FAIL** (1 of 18 criteria ❌ — §4's `currentUsername` injection.)

Everything else in this revision checks out, including the `window.currentUser` fix that prompted it: the Base.templ relocation, the explicit-argument rule, the scope table, and the §3 provenance correction are all accurate against the running app. The blocker is one line in §4 — swap the string concatenation for `templ.JSONString(data.User.Username)`.

SoulKyu:
Closing: this spec served its purpose (issue #164 is closed). The spec text remains on `looper/planner/164-feat-collaboration-mentions-in`.

PR #176, opened 2026-08-02
feat(collaboration): @mentions in alert comments

SoulKyu:
## Summary

Implements #164: `@username` mentions in alert comments, so a comment addressed to a teammate actually reaches them instead of depending on them opening that specific alert.

- **Detection**: `mentionsUsername()` in `activity_handlers.go` does a case-insensitive match on `@username` with a word-boundary check, so `@bob` doesn't fire for `@bobby`. A new `mentionsOnly=true` flag on `GET /api/v1/dashboard/activity` narrows the feed to comments mentioning the requesting session's user, excluding their own comments (self-mentions never notify the author).
- **Highlighting**: `static/js/mention_text.js` is a pure, HTML-escaping renderer that wraps `@tokens` in a highlight span, with a distinct style for the mention addressed to the viewer. Wired into both comment UIs — the main dashboard modal (`modal_components.templ`) and the statistics resolved-alert modal (`alert_modal_shared.templ`) — via `x-html`.
- **Mentions badge**: `PageNavigator.templ` gets a self-contained `mentionsWatcher()` Alpine component (present on every page that renders the nav) that polls the mentions feed every 30s, shows an unseen-count badge next to the Activity tab, and deep-links to `/activity?mentions=1` on click (which also marks mentions as seen). A separate "notified" watermark, defaulted to "now" on first run, drives a one-time browser notification per new mention via a new `NotificationService.showMentionNotification()` — deliberately bypassing the existing severity/filter-preset scoping, since a mention is a direct ask from a teammate, not a fired alert.
- **Activity page**: a `Mentions` toggle plus `?mentions=1` deep-link support in `Activity.templ`.

## Scope cut

Ships **without** `@`-autocomplete in the comment box, and without the `/api/v1/users/mentionable` endpoint the issue proposed to feed it — the issue explicitly allows this fallback ("If that proves contentious, ship phase 1 without autocomplete"), and free-text `@name` still works end to end for detection, highlighting, badge and notification. Adding an endpoint with no caller would be dead code.

## Testing

- `go build ./...`, `go vet ./...`, `go test ./internal/webui/...` — all pass (188 tests).
- `node --test internal/webui/static/js/mention_text.test.js` — 5 new tests covering escaping, highlighting, case-insensitivity and self-mention styling.
- New `TestMentionsUsername` in `activity_handlers_test.go` covers case-insensitivity, the `@bob`-vs-`@bobby` boundary case, and trailing punctuation.

## Test plan

[...]

PR #179, opened 2026-08-02
spec(webui): plan fix for fabricated /profile data and broken Copy-ID button

### LINKED ISSUE #177: fix(webui): profile page renders fabricated account/session data and its Copy-ID button copies a literal template string
## Problem

The `/profile` page (linked from the dashboard user menu, `NewDashboard.templ:324`) presents invented data as fact and ships a broken button:

1. **Fabricated account/session data** — `internal/webui/handlers/profile_handlers.go:45-52` hardcodes every date it renders:
   - `CreatedAt: time.Now().AddDate(0, -3, -15)` → "Member Since" is always *3.5 months ago*, for every user, forever.
   - `LastLogin: time.Now().Add(-2 * time.Hour)` → "Last Login" is always *2 hours ago*.
   - `SessionInfo.CreatedAt: time.Now().Add(-30 * time.Minute)` / `ExpiresAt: time.Now().Add(7 * 24 * time.Hour)` → "Session Started", "Expires" and the "Duration" line (always exactly 30 minutes) are all fiction.
   - `EmailVerified: user.Email != ""` (line 47) → the green "Verified" badge means "has an email string", not verified (no verification flow exists).
   - `Stats: {TotalAlerts: 156, ActiveAlerts: 12, ResolvedAlerts: 144}` (lines 54-59) — hardcoded demo numbers (currently unrendered, but a trap for whoever wires the template to them).

2. **Copy-ID button copies a literal template string** — `Profile.templ:242` puts `userId: '{ data.User.ID }'` inside the page's `<script>`. templ does **not** interpolate inside `<script>` elements, so the generated `Profile_templ.go` (WriteString call for the script block) ships the literal text `{ data.User.ID }` to every browser. Clicking the "ID" button copies that garbage string to the clipboard. There is also no feedback either way: `copyUserId()` sets `showToast`, but no element in the template binds it.

The real data exists end-to-end: `proto/auth.proto:114-115` already carries `created_at` and `last_login` on the `User` message, and the backend model has both (`internal/backend/models/models.go:18-20`). The webui gRPC client simply drops them (`internal/webui/client/backend_client.go:51-58`), and the handler papers over the gap with `time.Now()` arithmetic.

## Impact

- Admins configuring impersonation (`config.Admin.ImpersonationAllowedUsers`) need user IDs; the page's one-click copy for that workflow silently puts `{ data.User.ID }` in their clipboard.
- Any operator who reads "Member Since", "Last Login", "Session Expires" or the "Verified" badge is being lied to — e.g. checking "when did this account last log in?" during an access review returns a fabricated timestamp that always looks fresh.

## Proposed approach

- Extend `client.User` (and the `ValidateSession`/`GetProfile` unmarshalling) with `CreatedAt` / `LastLogin` from the proto, and render those.
- Pass the user ID to the script via a `data-*` attribute (or `templ.JSONScript`) instead of an interpolation that templ never performs, and either render the toast markup or swap the icon inline like the alert modal's Copy-link button does.
- For the session card: render only what the backend actually knows (e.g. expiry from the session record), and drop the rest rather than inventing it.
- Delete the hardcoded `UserStats` block or wire it to the statistics service; don't leave demo numbers in a production handler.
- Respect the templ workflow: edit `.templ` + `make webui-templates`, never `*_templ.go`.

## Acceptance criteria

- [ ] "Member Since" and "Last Login" show the account's real `created_at` / `last_login` from the backend (verifiable by registering a fresh user).
- [ ] The "ID" button puts the actual user ID in the clipboard, with visible feedback on success.
- [ ] No date, badge or counter on `/profile` is derived from `time.Now()` arithmetic or hardcoded constants in `profile_handlers.go`.
- [ ] The "Verified" badge is either backed by a real flag or removed.

SoulKyu:
## Summary
- Planning spec for #177: `/profile` invents Member Since/Last Login/session timing/Verified badge via `time.Now()` arithmetic and hardcoded constants, and its Copy-ID button copies a literal `{ data.User.ID }` string because templ doesn't interpolate inside `<script>`.
- Spec at `specs/2026-08-02-177-fix-webui-profile-page.md`: wires real `created_at`/`last_login`/`email_verified` through proto → backend `ValidateSession` → webui client → handler (the only path `/profile` actually reads from), sources session start time from a new login-time cookie key mirroring the existing `ImpersonationStartedAt` pattern (no new DB/RPC round trip), deletes the dead `UserStats` block, and fixes the Copy-ID button via the already-shipped attribute-interpolation + icon-swap-feedback patterns used elsewhere in this codebase.
- No source code changed in this PR — spec only.

## Test plan
- [ ] Implementer runs `make proto && make webui-templates && go build ./... && go test ./...` per the spec's Validation section
- [ ] Manual `/profile` verification steps listed in the spec (fresh classic user, OAuth user, Copy-ID clipboard + feedback, page-source check for the literal script bug)

Closes #177

### DIFF
--- a/specs/2026-08-02-177-fix-webui-profile-page.md
+++ b/specs/2026-08-02-177-fix-webui-profile-page.md
@@ -6,19 +6,19 @@
 
 ## Problem
 
-`internal/webui/handlers/profile_handlers.go:39-60` (`ProfilePage`) invents
+`internal/webui/handlers/profile_handlers.go:16-63` (`ProfilePage`) invents
 every date and counter it renders instead of reading real data:
 
 ```go
-CreatedAt:     time.Now().AddDate(0, -3, -15),          // "Member Since": always 3.5 months ago
-LastLogin:     &[]time.Time{time.Now().Add(-2 * time.Hour)}[0], // "Last Login": always 2h ago
-EmailVerified: user.Email != "",                          // green "Verified" badge means "has an email string"
+CreatedAt:     time.Now().AddDate(0, -3, -15),                    // :45  "Member Since": always 3.5 months ago
+LastLogin:     &[]time.Time{time.Now().Add(-2 * time.Hour)}[0],   // :46  "Last Login": always 2h ago
+EmailVerified: user.Email != "",                                  // :47  green "Verified" badge means "has an email string"
 ...
 SessionInfo: pages.SessionInfo{
-    CreatedAt: time.Now().Add(-30 * time.Minute),          // "Session Started": always 30m ago
-    ExpiresAt: time.Now().Add(7 * 24 * time.Hour),          // "Expires": always "in 7 days"
+    CreatedAt: time.Now().Add(-30 * time.Minute),                 // :51  "Session Started": always 30m ago
+    ExpiresAt: time.Now().Add(7 * 24 * time.Hour),                // :52  "Expires": always "in 7 days"
 },
-Stats: pages.UserStats{TotalAlerts: 156, ...},              // hardcoded demo numbers, currently unrendered
+Stats: pages.UserStats{TotalAlerts: 156, ...},                    // :54-59 hardcoded demo numbers, currently unrendered
 ```
 
 Separately, `internal/webui/templates/pages/Profile.templ:242` puts
@@ -27,27 +27,26 @@ not interpolate `{ }` expressions inside `<script>` elements (confirmed
 against the working precedent below), so every browser receives the literal
 JS source `userId: '{ data.User.ID }'`. Clicking the "ID" button copies that
 literal string, and there is no visible feedback either way (`showToast` is
-set but no element in the template reads it).
+set at `Profile.templ:241,246,248,258,260` but no element in the template
+reads it).
 
-The real data already exists:
-- `proto/auth.proto:110-119` (`User` message) has `created_at` (field 4) and
-  `last_login` (field 5) already.
+Real data exists for the *account* dates, and only for those:
+
+- `proto/auth.proto:110-119` (`User` message) already carries `created_at`
+  (field 4) and `last_login` (field 5).
 - `internal/backend/models/models.go:13-27` (`models.User`) has real
-  `CreatedAt`, `LastLogin *time.Time`, and `EmailVerified bool` columns.
-  `EmailVerified` is genuinely meaningful for OAuth users — GitHub/Google/
-  Microsoft logins set it from the provider's own verified-email flag
-  (`internal/backend/services/oauth_service.go:229,257,288`, wired through
-  `internal/backend/database/oauth_db.go:16-24`) — it is only always-`false`
-  for classic username/password accounts, which have no verification flow.
-- `internal/backend/services/services.go:217-247` (`ValidateSession`) and
-  `:250-276` (`GetProfile`) both already build an `authpb.User` from the real
-  `user` row but only ever set `Id/Username/Email/CreatedAt(/Timezone)` —
-  `LastLogin` and email-verified state are silently dropped even though the
-  proto and the model both carry them.
+  `CreatedAt` and `LastLogin *time.Time` columns.
+- `internal/backend/services/services.go:217-246` (`ValidateSession`) builds
+  its `authpb.User` from the real `user` row loaded by `GetUserBySession`
+  (`internal/backend/database/gorm_db.go:297-306`, which selects the whole
+  row) but only ever sets `Id/Username/Email/CreatedAt(/Timezone)` —
+  `LastLogin` is silently dropped even though the proto and the model both
+  carry it, and even though the same file already populates it correctly two
+  functions later (`services.go:350-352`, `:395`).
 - `internal/webui/client/backend_client.go:51-58` (`client.User`) drops
   `CreatedAt`/`LastLogin` entirely when unmarshalling `ValidateSession`'s
-  response (`:298-314`), even though `resp.User.CreatedAt` is already
-  present on the wire and simply never read.
+  response (`:298-302`), even though `resp.User.CreatedAt` is already on the
+  wire and simply never read.
 
 `ProfilePage` sources its `user` from `middleware.GetCurrentUserFromContext`
 (`profile_handlers.go:17`), which is populated by `RedirectIfNotAuth`
@@ -58,6 +57,40 @@ fixing `ValidateSession`'s unmarshalling is enough to fix everything
 `ProfilePage` renders about the user — no new RPC round trip needed on that
 path.
 
+### The "Verified" badge has no truth to read
+
+There is no email-verification flow in this product. `models.User.EmailVerified`
+(`models.go:26`) is written in exactly one place — OAuth signup
+(`internal/backend/database/oauth_db.go:23,73,98`) — from
+`OAuthUserInfo.EmailVerified`, which each provider parser fills as follows
+(`internal/backend/services/oauth_service.go`):
+
+| Parser | Line | Value | Real verification signal? |
+|---|---|---|---|
+| GitHub | `:229` | `githubUser.Email != ""` | **No** — the exact "has an email string" heuristic issue #177 calls a lie, just moved to signup time |
+| Google | `:257` | `googleUser.EmailVerified` | Yes — the provider's real `email_verified` claim |
+| Microsoft | `:288` | `true` | **No** — hardcoded for every Microsoft user |
+| Generic OIDC | `:325-330` | `email_verified` claim if the provider sends a bool, else the zero value `false` | Only when the provider sends it |
+
+Classic username/password accounts never write the column at all, so they
+keep the GORM default `false` (`models.go:26`).
+
+So a badge "backed by the real `EmailVerified` flag" would still be a lie for
+GitHub and Microsoft users, and would only *disappear* for the classic
+accounts that are the majority. Sourcing it honestly would mean fixing two
+provider parsers (GitHub's `/user` response has no verified-email field at
+all — it needs a second call to `/user/emails`; Microsoft Graph exposes none),
+which is a different feature than "stop lying on `/profile`".
+
+Issue #177 AC4 allows either backing the badge with a real flag or removing
+it. **This spec removes it.** That keeps the page honest without inventing a
+verification feature, and it drops the entire proto/service/client plumbing
+that would otherwise exist only to carry `email_verified` to a badge that
+cannot tell the truth. `EmailVerified` stays in the DB model untouched for a
+future real verification flow; it just stops being rendered.
+
+### Out of reach: session timing
+
 The session "Started"/"Expires"/"Duration" numbers have no backend session
 lookup to source from at all today (`GetUserBySession`,
 `internal/backend/database/gorm_db.go:297-306`, never selects
@@ -67,12 +100,14 @@ method). Building one would mean either adding a second DB round trip to
 `/profile`) or wiring a new RPC end-to-end for a single page. Both are
 disproportionate to a "stop lying" issue. The session's real 7-day lifetime
 is already fixed and known (`services.go:165,1777`,
-`internal/webui/middleware/session.go:25`), and the codebase already has a
+`internal/webui/middleware/session.go:24`), and the codebase already has a
 precedent for stashing a real, request-time timestamp straight into the
 webui's own session cookie: `ImpersonationStartedAt`
-(`middleware/session.go:82,121-129`, `int64` Unix timestamp, nil-safe
+(`middleware/session.go:17,82,121-129`, `int64` Unix timestamp, nil-safe
 getter). This spec reuses that exact pattern for a `session_started_at` key
 set at login, instead of adding backend plumbing.
+
+### Dead code nearby
 
 `internal/webui/handlers/profile_handlers.go:65-101` (`GetProfileData`) has
 its own hardcoded stats (`acknowledged_alerts: 42`, `comments: 17`,
@@ -88,8 +123,9 @@ non-fabricating handler at `internal/webui/handlers/handlers.go:252-266`).
 
 1. "Member Since" and "Last Login" show the account's real `created_at` /
    `last_login`.
-2. The "Verified" badge reflects the real `EmailVerified` flag from the
-   backend, not `user.Email != ""`.
+2. The "Verified" badge is **removed** (`Profile.templ:109-116`) — no flag in
+   this codebase actually means "this email was verified" (see above), so
+   issue #177 AC4's "or removed" option is taken.
 3. "Session Started" / "Expires" / "Duration" show a real, request-derived
    session start time (or are hidden if one isn't available), never a
    `time.Now()`-relative fabrication.
@@ -100,6 +136,15 @@ non-fabricating handler at `internal/webui/handlers/handlers.go:252-266`).
 
 ## Non-goals
 
+- **No proto change at all.** With the badge gone, `created_at` (field 4) and
+  `last_login` (field 5) already carry everything `/profile` needs, so
+  `proto/auth.proto` and `internal/backend/proto/**` are untouched — no
+  `make proto` run, no regenerated `.pb.go` in the diff.
+- **No email-verification feature.** `models.User.EmailVerified` and the
+  OAuth parsers that fill it (`oauth_service.go:229,257,288,325-330`) are
+  left exactly as they are; they simply stop being rendered. Making that flag
+  trustworthy (GitHub `/user/emails`, a classic-auth verification mail flow)
+  is a separate issue.
 - `GetProfileData` (`profile_handlers.go:65-101`) and its own hardcoded
   stats — unrouted dead code, not reachable from `/profile` or anywhere
   else. Left untouched.
@@ -113,43 +158,23 @@ non-fabricating handler at `internal/webui/handlers/handlers.go:252-266`).
   card is powered by the webui's own login-time cookie value instead.
 - No design changes to the OAuth-provider badge, Quick Actions, or
   "Change Password" (still a stub `alert(...)`, unrelated to this issue).
-- No new `google.protobuf` message shape changes beyond the one new `User`
-  field.
 
 ## Approach
 
-### 1. Proto: add `email_verified` to `User`
+### 1. Backend: populate `LastLogin` in `ValidateSession`
 
-`proto/auth.proto:110-119`, inside `message User { ... }`, after
-`string timezone = 8;`:
-
-```proto
-  bool email_verified = 9;
-```
-
-Run `make proto` (`scripts/generate_proto.sh` — requires `protoc`,
-`protoc-gen-go`, `protoc-gen-go-grpc` on `PATH`) to regenerate
-`internal/backend/proto/auth/auth.pb.go` and `auth_grpc.pb.go`. The script
-wipes and regenerates `internal/backend/proto/{auth,alert}` wholesale, but
-`alert.proto` is unchanged so only the new `EmailVerified` field (and
-associated getters/marshal code) should show up in `git diff`.
-
-### 2. Backend: populate `LastLogin` and `EmailVerified` in `ValidateSession`
-
-`internal/backend/services/services.go:233-246`, extend the response the
-same way `LastLogin` is already populated elsewhere in this file (e.g. the
-nil-check pattern at `services.go:350-351`):
+`internal/backend/services/services.go:233-246`, add the nil-check that the
+same file already uses for `LastLogin` at `:350-352`:
 
 ```go
 resp := &authpb.ValidateSessionResponse{
 	Valid:   true,
 	Message: "Session is valid",
 	User: &authpb.User{
-		Id:            user.ID,
-		Username:      user.Username,
-		Email:         user.Email,
-		CreatedAt:     timestamppb.New(user.CreatedAt),
-		EmailVerified: user.EmailVerified,
+		Id:        user.ID,
+		Username:  user.Username,
+		Email:     user.Email,
+		CreatedAt: timestamppb.New(user.CreatedAt),
 	},
 }
 if user.LastLogin != nil {
@@ -161,12 +186,14 @@ if user.Timezone != nil {
 return resp, nil
 ```
 
-`GetProfile` is intentionally **not** touched (see Non-goals) — it's dead
-code and changing it doesn't affect anything `/profile` renders.
+That is the only backend change in this spec. `GetProfile` is intentionally
+**not** touched (see Non-goals) — it's dead code and changing it doesn't
+affect anything `/profile` renders.
 
-### 3. WebUI client: stop dropping `CreatedAt`/`LastLogin`/`EmailVerified`
+### 2. WebUI client: stop dropping `CreatedAt`/`LastLogin`
 
-`internal/webui/client/backend_client.go:51-58`, extend `User`:
+`internal/webui/client/backend_client.go:51-58`, extend `User` (the file
+already imports `time`, line 9):
 
 ```go
 type User struct {
@@ -178,21 +205,20 @@ type User struct {
 	Timezone      *string    `json:"timezone,omitempty"`
 	CreatedAt     time.Time  `json:"created_at"`
 	LastLogin     *time.Time `json:"last_login,omitempty"`
-	EmailVerified bool       `json:"email_verified"`
 }
 ```
 
-`ValidateSession` (`backend_client.go:298-302`), read the new fields off
-`resp.User` (already imports `timestamppb`, whose `*Timestamp` has
-`.AsTime()` — same idiom as `internal/webui/services/color_service.go:180-181`):
+In `ValidateSession` (`backend_client.go:277-315`), read the new fields off
+`resp.User` at `:298-302` (the file already imports `timestamppb`, line 14;
+`*Timestamp.AsTime()` is the same idiom as
+`internal/webui/services/color_service.go:180-181`):
 
 ```go
 user := &User{
-	ID:            resp.User.Id,
-	Username:      resp.User.Username,
-	Email:         resp.User.Email,
-	CreatedAt:     resp.User.CreatedAt.AsTime(),
-	EmailVerified: resp.User.EmailVerified,
+	ID:        resp.User.Id,
+	Username:  resp.User.Username,
+	Email:     resp.User.Email,
+	CreatedAt: resp.User.CreatedAt.AsTime(),
 }
 if resp.User.LastLogin != nil {
 	t := resp.User.LastLogin.AsTime()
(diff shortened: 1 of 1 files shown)


SoulKyu (commented):
@SoulKyu Reviewed this spec against the actual codebase, adversarially checking every file/line reference and code claim rather than trusting the write-up.

This PR touches only `specs/2026-08-02-177-fix-webui-profile-page.md` (no source changes). I verified every factual claim the spec makes:

- `profile_handlers.go:39-60`'s fabricated `time.Now()` arithmetic and hardcoded `UserStats` block match exactly what's in the code today.
- `Profile.templ`'s `<script>` block does contain the literal `userId: '{ data.User.ID }'` — confirmed templ does not interpolate inside `<script>`, and that the proposed attribute-interpolation fix (`x-data={ "profilePage('" + data.User.ID + "')" }`) mirrors the exact shipped pattern in `StatisticsDashboard.templ:43` and `filter_components.templ:7`.
- The proto/backend/client data-dropping chain (`auth.proto` → `ValidateSession`/`GetProfile` → `backend_client.go`'s `User` struct) checks out field-by-field, and the new `email_verified = 9` proto field doesn't collide with any other field in the `User` message (field 9 elsewhere is `LoginResponse.error`, a different message).
- The `ImpersonationStartedAt` cookie precedent the plan reuses for `session_started_at` is real and already working the same way (int64 Unix timestamp, nil-safe getter) — confirmed exactly two `session_id` write sites (`Login`, `OAuthCallback`) that both need the new key.
- I specifically hunted for an injection risk in the naive single-quote JS embedding fix: traced `models.User.ID` generation through `BeforeCreate` (`models.go:36-40`) and confirmed *every* user-creation path (classic `CreateUser` and OAuth `CreateOAuthUser`) relies on the same `GenerateID()` → 32-char lowercase-alphanumeric charset, with no caller ever pre-setting `ID`, so the safety claim holds even for OAuth-created accounts.
- Backward compatibility looks solid: `authpb.User{...}` is only ever constructed with keyed struct literals across the codebase, so the new proto field is safe to add; the existing `ValidateSession` test only asserts `Timezone` and stays unaffected; pre-existing sessions degrade gracefully (session card hides itself) instead of crashing.

I couldn't find a gap, feasibility problem, or unverified claim worth blocking on. Nice work being this thorough about grounding every design decision in what the code actually does instead of assuming it — the verification ledger at the bottom made this an unusually fast, high-confidence review. Approving.



SoulKyu:
## 🧪 QA Report

Spec-only PR on a planner branch (`specs/2026-08-02-177-fix-webui-profile-page.md`, +530 lines, no source change), so per QA policy I graded **the spec**, not issue #177's acceptance criteria. Stack booted from this SHA (`docker compose` @ webui :10081) and every codebase reference was checked against the running app and the code at this SHA.

|---|---|---|
| Fabricated-data problem statement is real and accurately cited (`profile_handlers.go:45-52,54-59`) | ✅ | Registered `qa179user` at 19:29:14 UTC; `/profile` rendered "Member Since Apr 17, 2026", "Last Login Aug 2 5:29 PM", "Session Started 7:00 PM", "Expires Aug 9 7:30 PM" — DB says `created_at = last_login = 2026-08-02 19:29:14`. Exactly the `time.Now()` offsets the spec quotes. |
| Copy-ID `<script>` bug is real (`Profile.templ:242`, templ does not interpolate in `<script>`) | ✅ | Live page source ships `userId: '{ data.User.ID }'`; generated `Profile_templ.go:308` contains the same literal. Playwright: clicking the "ID" button called `clipboard.writeText("{ data.User.ID }")` while the page displays `8h94fiexdwdkj3s68uz1ke00ddcggeeg`. Button text stayed "ID" — no feedback, `showToast` bound nowhere (`Profile.templ:241,246,248,259,261` only). |
| Proposed fix pattern (attribute interpolation) actually works in this codebase | ✅ | `/statistics` (the cited precedent `StatisticsDashboard.templ:43`) renders `currentUserId: &#39;8h94fiexdwdkj3s68uz1ke00ddcggeeg&#39;` — real value, HTML-escaped, un-escaped by the browser before Alpine parses. `filter_components.templ:7` same idiom. Icon-swap feedback precedent confirmed at `modal_components.templ:1121-1130` + `dashboard_modal.templ:151-156`. |
| Data-availability claims (proto / model / service / client) | ✅ | `proto/auth.proto:110-119` has `created_at`(4)/`last_login`(5), next free field is 9 as the spec assumes. `models.go:13-27` has `CreatedAt`/`LastLogin`/`EmailVerified`. `services.go:233-246` (ValidateSession) and `:264-274` (GetProfile) set only Id/Username/Email/CreatedAt(/Timezone). `backend_client.go:51-58` + `:298-314` drop them. `GetUserBySession` (`gorm_db.go:297-306`) loads the full user row, so `LastLogin`/`EmailVerified` are available with no extra query. |
| Auth-path claim: fixing `ValidateSession` alone is sufficient for `/profile` | ✅ | `router.go:381` → `RedirectIfNotAuth` (`auth.go:86-122`) → `ValidateSession` → `c.Set("user")` → `GetCurrentUserFromContext`. Live: unauthenticated `/profile` → 302 `/login`; backend stopped → 503 "Service temporarily unavailable" with session preserved, exactly as `auth.go:104-108` describes. |
| Session-timing claims + cookie-pattern reuse | ✅ | No `GetSessionByID` exists; `GetUserBySession` never selects `sessions.created_at/expires_at` (confirmed — the columns *do* exist in Postgres, they are just not exposed). 7-day lifetime confirmed at `services.go:165,1777` and `session.go:25`; DB row: `created_at 19:29:14.75` → `expires_at 2026-08-09 19:29:14.75`, so `startedAt + 7d` is truthful, not a fabrication. `ImpersonationStartedAt` precedent verified at `session.go:18,82,121-129`; `SetSessionValue` does return `error` as the snippets assume; exactly two `session_id` write sites (`handlers.go:126`, `oauth_handlers.go:124`), both covered, and `oauth_handlers.go` already imports `time` (line 9). |
| Dead-code / non-goal claims | ✅ | `GetProfileData` has no route (only its own definition); the real `/api/v1/profile` is `handlers.GetCurrentUser` (`router.go:206`, `handlers.go:252-266`). `BackendClient.GetProfile` is never called from webui (only its own body at `:352` and the unrelated health check at `:143`). `UserStats`/`Stats` are unrendered — confirmed live: no `156`/`144` anywhere in the served HTML. No test touches `ProfilePage`/`ProfileData`; `update_timezone_test.go:76-78` only asserts `Timezone`. |
| Single-quote embedding safety claim for user IDs | ✅ | `models.go:36-41` `BeforeCreate` assigns `GenerateID()` whenever `ID == ""`, and `CreateOAuthUser` (`oauth_db.go:17-24`) never sets an ID — so OAuth users get the same 32-char `[a-z0-9]` IDs, not a provider-controlled string. Live ID matched the charset. |
| **`EmailVerified` justification for keeping the "Verified" badge** | ❌ | Spec lines 36-41: *"GitHub/Google/Microsoft logins set it from the provider's own verified-email flag (`oauth_service.go:229,257,288`)"*. Only Google does (`:257`, real `email_verified` claim). **`:229` is `EmailVerified: githubUser.Email != ""`** — the identical "has an email string" heuristic the issue calls a lie, merely relocated to signup time. **`:288` is a hardcoded `EmailVerified: true`** for every Microsoft user. This false premise is the sole basis for Goal 2 ("badge reflects the real `EmailVerified` flag") over issue AC4's other option (remove the badge). |

### Adversarial pass

- **Boundary / empty** — brand-new account (age: seconds) still renders "Member Since Apr 17, 2026": the fabrication the spec targets is confirmed at the boundary, not just anecdotally. `LastLogin` nil path already guarded (`Profile.templ:145`) and untouched by the spec ✓. `resp.User.CreatedAt` can never be nil on the ValidateSession path (always `timestamppb.New`), so the proposed `.AsTime()` cannot yield a 1970 date ✓.
- **Malformed / hostile** — attacked the spec's quoting-safety claim by looking for a user ID that is *not* `GenerateID()`-derived (an OAuth-provider-supplied ID would let a crafted value break the `'…'` wrapping in `x-data`): `BeforeCreate` + `CreateOAuthUser` close that hole, and templ escapes `'` to `&#39;` in attributes anyway. Claim survives.
- **Error / unavailable path** — stopped `qa179-backend` mid-session and hit `/profile`: 503 with the session intact (not a wiped-but-successful render), matching the spec's reading of `auth.go`. Restarted, session still valid.
- **Auth & scope** — unauthenticated `/profile` → 302 `/login` ✓. Checked the impersonation angle the spec flags: `GetEffectiveUser` (`auth.go:176-190`) builds a bare `&client.User{ID,Username}`, but `/profile` uses `GetCurrentUserFromContext` only, so the spec's "admin sees their own profile" note is correct ✓.
- **Adjacent flows** — the spec's `x-data` rewrite is the same construct `/statistics` and the filter components already ship; loaded both live, no console breakage, real IDs interpolated. Removing `UserStats` touches nothing else (`grep` + live HTML both clean).
- **Persisted state** — pre-existing sessions have no `session_started_at` key; the spec explicitly degrades to hiding the started/expires/duration block rather than crashing or inventing (Risks §1), which is the right call. Verified `int64`-through-cookie round-trip is an already-shipped pattern (`ImpersonationStartedAt`), not a new assumption.
- **Where it broke** — the OAuth `EmailVerified` claim. An implementer following this spec ships a badge that, for a GitHub user, still means exactly `email != ""`, and for a Microsoft user is unconditionally `true`, then closes AC4 ("the Verified badge is either backed by a real flag or removed") believing it is satisfied. Live DB confirms the visible half of this: `email_verified = f` for the classic account, so the only user-visible effect for the majority is the badge disappearing, while the two providers where it *does* show remain fabricated.

**Note (not a failure):** §5 keeps a hardcoded `7 * 24 * time.Hour` in `profile_handlers.go`, which is in tension with issue AC3's letter ("no … hardcoded constants in `profile_handlers.go`"). The value is truthful (verified against `sessions.expires_at`) and the spec discloses the duplication with a `ponytail:` comment naming the upgrade path, so I did not fail it — flagging so the implementer is not surprised by a later review.

### Fix needed

Correct lines 36-41 and Goal 2: only Google (`:257`) and generic OIDC (`:328`) source a real verified flag. Either scope the badge to providers that supply one, or take AC4's second option and remove it. Everything else in the spec checked out line-for-line.

**Verdict: FAIL**

SoulKyu:
## 🧪 QA Report

**Spec/docs-only PR** (`looper/planner/177-*`, diff = `specs/2026-08-02-177-fix-webui-profile-page.md` only, +641/-0). Per the spec-PR rule the linked issue's ACs are **not** graded as implemented — the spec itself is graded for reference accuracy and whether following it would misdirect the implementer. All checks below were run against the app booted from this clone (webui :9081, backend :9080/:51051, postgres :6432).

|---|---|---|
| Fabrication diagnosis is real (`profile_handlers.go:45,46,47,51,52,54-59`) | ✅ | Registered `qauser179` at 00:53, `/profile` rendered **Member Since Apr 17 2026**, **Last Login Aug 2 8:53 PM**, **Session Started 10:23 PM**, **Duration 30m**; DB truth `users.created_at=2026-08-02 22:53:20`, `last_login=22:53:21`, `sessions.created_at=22:53:21` → every value invented, exactly as described |
| templ does not interpolate `{ }` inside `<script>` (`Profile.templ:242`) | ✅ | Generated `Profile_templ.go:308` emits the literal `userId: '{ data.User.ID }'`; served `/profile` HTML contains the same literal. Real ID appears exactly once (the User ID display field), never in the script |
| `showToast` is written but bound to no element (`:241,246,248,…`) | ✅ | `rg showToast Profile.templ` → 5 hits, all inside `<script>`; `rg 'x-show\|x-text' Profile.templ` → **0** bindings |
| `created_at` already on the wire, `LastLogin` silently dropped by `ValidateSession` | ✅ | grpcurl `notificator.auth.AuthService/ValidateSession` → `{"createdAt":"2026-08-02T22:53:20.863278Z"}` present, **`lastLogin` absent** despite a non-null DB value — precisely the spec's §1/§2 diagnosis |
| The fix idiom §1 tells the implementer to copy (`services.go:350-352`) actually works | ✅ | Same proto `User` message via `SearchUsers`/`ListUsers` returns `"lastLogin":"2026-08-02T22:53:21.047187Z"` matching the DB → confirms §1's 3-line nil-check is sufficient and the "no proto change at all" non-goal is correct |
| "Verified" badge premise (`EmailVerified` is untrustworthy) | ✅ | Classic user has `email_verified = f` in Postgres yet the badge **renders** (badge is driven by `user.Email != ""`). Parser lines confirmed verbatim: GitHub `oauth_service.go:229` `githubUser.Email != ""`, Google `:257` real claim, Microsoft `:288` literal `true`, OIDC `:325-330` conditional; writes only at `oauth_db.go:23,73,98` |
| §3 cookie-timestamp precedent + "exactly two `session_id` write sites" | ✅ | `ImpersonationStartedAt` const/setter/nil-safe getter present in `middleware/session.go`; `rg '"session_id"'` across `internal/webui/` → only 2 cookie writes (`handlers.go:126`, `oauth_handlers.go:124`), all other hits are gin-context `c.Get`/`c.Set`. The two differing error styles (JSON 500 vs log+redirect) match the real handlers; `oauth_handlers.go` does import `time` (:9) and `log` (:5) |
| §5 attribute-interpolation fix mechanism works at runtime | ✅ | Live `/statistics` serves `x-data="{ …, currentUserId: &#39;4hq60tujgiqny8jearpa1fjfuhaiv7vw&#39; }"` — the real ID, via the exact `StatisticsDashboard.templ:43` pattern the spec prescribes. Copy-link icon-swap precedent confirmed at `modal_components.templ:1120-1131` + `dashboard_modal.templ:151-156` |
| Dead-code claims (out-of-scope justifications) | ✅ | `GetProfileData` → only its own definition, no route; `BackendClient.GetProfile` (`:340`) → no caller (only the unrelated health check at `:143`); `UserStats`/`EmailVerified` readers → exactly the sites the spec names |
| Routing claim: `/profile` → `RedirectIfNotAuth` → `ValidateSession` | ✅ | `router.go:377,381`; live: unauthenticated and garbage-cookie requests both 302 → `/login`; `auth.go:102,118` sets the context user from `ValidateSession` on every request, so the §2 client fix alone does reach `ProfilePage` |
| ID-escaping safety claim (naive single-quote embedding) | ✅ | `generateRandomString(32)` charset `[a-z0-9]`; issued ID `4hq60tujgiqny8jearpa1fjfuhaiv7vw`; every explicit `models.User{ID: …}` in the tree is in `*_test.go` — no production path takes an ID from input |
| Line/symbol references accurate enough to follow | ✅ | ~40 references re-derived; all land on the named symbol. Drift of 1–2 lines on a few (`ImpersonationStartedAt` cited `:17`/actual `:18`; `MaxAge` `:24`/`:25`; `showToast` `:258,260`/`:259,261`; test assertion `:70-78`/`:72-78`) — each also names the symbol, so none misdirects |
| Issue #177 AC coverage | ✅ | AC1→goal 1, AC2→goal 5, AC3→goals 3/4, AC4→goal 2 (removal, which AC4 explicitly permits). AC3's letter-vs-spirit gap (the mirrored `7*24*time.Hour` constant staying in `profile_handlers.go`) is **self-disclosed** in Risks & trade-offs rather than glossed over |

**Adversarial pass** — cases constructed to break the spec, and what happened:

- **Error/unavailable path:** stopped the `backend` container mid-session and hit `/profile` → **503**, no page rendered. No fabricated or stale-success render, and §4 keeps the `user == nil` guard that the pass-through branch (`auth.go:89`) relies on. Nothing to fix.
- **Auth & scope:** `/profile` with no cookie → 302 `/login`; with `Cookie: notificator-session-v2=garbage` → 302 `/login`. Confirms the spec's premise that `ProfilePage` never runs without a `ValidateSession`-populated user.
- **Missing/absent state (the spec's own degradation claim):** `GetSessionValue` is `session.Get(key)` → returns `nil` for a key that was never set, so §3's getter returns `nil` and §5's `CreatedAt != nil` guard drops the block. No panic path; the pre-existing-session case the spec calls "self-healing" is genuinely safe.
- **Completeness of §3 (the likeliest way this spec misdirects):** hunted for a *third* place a webui session is established that the spec forgot — grepped every `"session_id"` occurrence, not just the `SetSessionValue` idiom, and checked `Register` for auto-login (it returns `redirect:/login`, no session write). Only the two documented sites exist. Claim holds.
- **Struct-widening blast radius:** §2 adds fields to `client.User` — checked all 4 construction sites (`backend_client.go:298,357,1094`, `auth.go:182`); all use named fields, so none breaks. `GetEffectiveUser`'s bare `&client.User{ID,Username}` is correctly identified as irrelevant to `/profile`.
- **Nil-timestamp → 1970 date:** attacked §2's unconditional `resp.User.CreatedAt.AsTime()`. `AsTime()` on nil yields 1970 rather than panicking, so I checked the wire: `ValidateSession` already emits `createdAt` **today**, before any change — so even an un-upgraded backend during a rolling deploy can't produce the 1970 render. The spec's justification is correct for the right reason.
- **Hostile value in the interpolated attribute:** tried to find an ID that would break out of `'…'` in `x-data`. No production path assigns a user ID from external input (OAuth signup deliberately leaves it empty so `BeforeCreate` generates it), and the charset is `[a-z0-9]` only. templ additionally escapes `'` → `&#39;`, verified in the live `/statistics` markup. The "no escaping helper needed" shortcut is safe as specified.
- **Adjacent flow:** verified removing the badge doesn't take the OAuth-provider/"Classic Auth" badge with it — that block is `Profile.templ:96-107`, a separate sibling inside the same wrapper `div` (`:95-117`), sourced from real data.
- **Not manufactured as failures:** the ±1–2 line drifts listed above, and "Last Login" reading as *just now* right after login (truthful, and what AC1 asks for). Neither is a user-visible wrong outcome.

Could not break any claim. The one deviation from issue #177's letter (AC3's hardcoded 7-day constant) is disclosed by the spec itself with its rationale and upgrade path, not hidden.

**Verdict: PASS**

SoulKyu:

SoulKyu:
## 🧪 QA Report

**Spec/docs-only PR** on planner branch `looper/planner/177-fix-webui-profile-page` (diff = `specs/2026-08-02-177-fix-webui-profile-page.md` only, +530). Per the spec-PR rule I grade **the spec's accuracy at this SHA**, not issue #177's ACs. Stack booted from this clone (webui :12081, backend :12080/:54051, postgres :9432); ~45 file:line references re-derived by hand.

This head (`4dbd2ce`, *"drop the Verified badge instead of plumbing email_verified"*) correctly fixes the `EmailVerified` premise that failed the earlier review — that part now checks out exactly. But the branch was **rebased onto `c0dc4f5` (#176)**, and one citation was not re-derived against that new base.

|---|---|---|
| Fabrication diagnosis real & cited verbatim (`profile_handlers.go:16-63`, `:45,46,47,51,52,54-59`) | ✅ | Registered `qa179u` at **03:39:15 UTC**; live `/profile` rendered **Member Since Apr 18 2026**, **Last Login Aug 3 2:09 AM**, **Session Started 3:39 AM**, **Duration 30m**. Postgres truth: `users.created_at = last_login = 2026-08-03 03:39:15`. Every value invented, exactly the offsets quoted. |
| Copy-ID `<script>` bug + unbound `showToast` (`Profile.templ:242`, `:241,246,248,…`) | ✅ | Served HTML contains the literal `userId: '{ data.User.ID }'`. Playwright: clicking **Copy User ID** called `clipboard.writeText("{ data.User.ID }")` while the page displays `5avzy7q2rwf0puu1aej5wpqvmkrn3wmc`; button text stayed `"ID"` — no feedback. |
| Data-availability chain (`proto/auth.proto:110-119`, `models.go:13-27`, `services.go:217-246`, `backend_client.go:51-58,298-302`) | ✅ | All four exact. `created_at`=4/`last_login`=5 present → the "**no proto change at all**" non-goal is correct. `ValidateSession` sets only `Id/Username/Email/CreatedAt(/Timezone)`; the same file's `:350-352` and `:395` nil-check idiom the fix copies is verbatim correct. `GetUserBySession` (`gorm_db.go:297-306`) loads the whole row → no extra query needed. |
| Auth-path claim: fixing `ValidateSession` alone reaches `/profile` | ✅ | `router.go:381` → `RedirectIfNotAuth` (`auth.go:86-121`) → `ValidateSession` → `c.Set("user")`. Live: no cookie → 302 `/login`; `notificator-session-v2=garbage` → 302 `/login`; **backend stopped mid-session → 503 "Service temporarily unavailable"** with the session intact (not a wiped-state success), matching `auth.go:104-108`. |
| "Verified" badge premise + the decision to **remove** it | ✅ | Table at spec lines 68-73 is exact: GitHub `oauth_service.go:229` = `githubUser.Email != ""`, Google `:257` = real claim, Microsoft `:288` = literal `true`, OIDC block real (at `:328-330`). Writes only at `oauth_db.go:23,73,98`; classic signup never writes it. **Live proof the flag is not what the badge shows**: `qa179u` has `email_verified = f` in Postgres yet the green **Verified** badge renders. AC4's "or removed" option is a sound reading. |
| Session-timing rationale + cookie-pattern reuse (`session.go` `ImpersonationStartedAt`, two `session_id` write sites) | ✅ | No `GetSessionByID` exists; `GetUserBySession` never selects `sessions.created_at/expires_at`. `grep 'SetSessionValue(c, "session_id"'` → exactly `handlers.go:126` (JSON 500) and `oauth_handlers.go:124` (log + redirect) — the **two differing error styles §3 spells out are both correct**, and `oauth_handlers.go` does import `time` (:9) and `log` (:5). 7-day lifetime truthful: DB `sessions.created_at 03:39:15` → `expires_at 2026-08-10 03:39:15`. `GetSessionValue` = `session.Get` → nil for an unset key, so §3's getter degrades instead of panicking. |
| Dead-code / non-goal claims | ✅ | `GetProfileData` unrouted; real `/api/v1/profile` is `router.go:206` → `handlers.go:252-266` (exact). `services.go:250-276` + `backend_client.go:340-374` `GetProfile` uncalled except the health check at `:143` (exact). `UserStats`/`Stats` read by no template markup — confirmed by reading `Profile.templ` end-to-end and by the served HTML (no `156`/`144`). |
| Struct/template surgery is safe (`Profile.templ:9-37,109-116,95-117,170-193,122-131`) | ✅ | Every one of those ranges is line-exact. `SessionInfo`/`ProfileUser` constructed only in `profile_handlers.go` → pointer-ising the timestamps breaks nothing. `templates.FormatDate/FormatDuration` are `utils.go:20,24` and take values, as stated. Removing the badge leaves the sibling OAuth/"Classic Auth" block (`:96-107`) intact. Icon-swap precedent (`modal_components.templ:1120-1131` + `dashboard_modal.templ:151-156`) verified. |
| **§5's Copy-ID fix cites a precedent that no longer exists at this SHA — and prescribes the idiom the parent commit removed for security** | ❌ | Spec lines 420-430 quote, as *"already a shipped, working pattern in this codebase at `StatisticsDashboard.templ:43`"*: `x-data={ "{ …, currentUserId: '" + data.User.ID + "' }" }`. **At this SHA line 43 is `<div`**, and `:45-48` are `data-user-id={ data.User.ID }` / `data-username=…` / `x-data="{ …statisticsViewsMixin() }"` / `x-init="currentUserId = $el.dataset.userId; …"`. `c0dc4f5` (#176 — the very commit this branch was rebased onto) removed the quoted line, stating: *"Replace JS-string-concatenated x-data on StatisticsDashboard with data-user-id/data-username attributes hydrated via x-init, **closing a JS-injection vector through an unvalidated username**."* Issue #177's own proposal says the same thing (*"Pass the user ID to the script via a `data-*` attribute (or `templ.JSONScript`)"*). The spec tells the implementer to write `x-data={ "profilePage('" + data.User.ID + "')" }` and justifies skipping escaping as *"matching the existing convention above"* — that convention was deleted one commit ago, for security. |

### Adversarial pass — what I tried

- **The failing case, isolated on the live app.** The stack's webui image predates the rebase, so I used that as a control: the **pre-rebase** container serves `currentUserId: &#39;5avzy7q2rwf0puu1aej5wpqvmkrn3wmc&#39;` on `/statistics` — the spec's quote, verbatim — while the tree at this SHA compiles to `data-user-id` (2 occurrences in the committed `StatisticsDashboard_templ.go`, 0 occurrences of the concat). Three independent sources agree the citation describes the *old* tree: the `.templ` source, the generated `.go`, and `git show c0dc4f5`.
- **Tried to argue the finding away, twice, and could not.** (a) *Does the prescribed code still work?* Yes — user IDs are `GenerateID()` → `[a-z0-9]{32}` (`models.go:36-41,326-336`; `oauth_db.go:17-24` never sets an ID, so OAuth users get the same), and templ escapes `'`→`&#39;` in attributes, so no injection **for this field**. (b) *Is the mechanism claim itself dead?* No — `filter_components.templ:7` still ships the concat idiom, so "templ interpolates in attributes" is true. That is why this is a spec-accuracy failure, not a runtime one: the fix compiles and copies the right ID, but the implementer is sent to a line that doesn't contain the quoted code, and told to reintroduce the pattern a security fix had just replaced — with the escaping shortcut justified by a convention that no longer exists.
- **Boundary/empty:** an account seconds old still renders "Member Since Apr 18, 2026" — the fabrication holds at the boundary. `resp.User.CreatedAt` is unconditionally `timestamppb.New(...)`, so §2's `.AsTime()` cannot yield 1970 even mid-rollout. Existing `if data.User.LastLogin != nil` guard (`:145`) is untouched.
- **Persisted state:** sessions created before this ships have no `session_started_at`; verified `GetSessionValue` returns nil for an unset key, so §5's `CreatedAt != nil` guard drops the block rather than crashing. `CreatedAt`/`ExpiresAt` are set together in §4, so the `*ExpiresAt` deref inside the guard is safe.
- **Blast radius of the struct change:** `client.User` gains two fields — all construction sites use named fields, so none breaks. `GetEffectiveUser` (`auth.go:176-190`) builds a bare `&client.User{ID,Username}`, but `/profile` only calls `GetCurrentUserFromContext`, so the spec's impersonation note is right.
- **Error / auth paths:** covered in the table (503 with session preserved; 302 on absent and on forged cookie).
- **Test blast radius:** the only `ValidateSession` test is `update_timezone_test.go` and it asserts `Timezone` only — adding `LastLogin` cannot break it.
- **Not counted as failures** (line drift; each also names its symbol, so none misdirects): `session.go` `ImpersonationStartedAt` cited `:17`→actual `:18`, `MaxAge` `:24`→`:25`, `SetSessionValue` `:32-36`→`:33-37`, const block `:14-18`→`:15-19`; `Profile.templ` `showToast` `:258,260`→`:259,261`, `handleLogoutResponse` `:271-275`→`:273-277`; OIDC block `:325-330`→`:328-330`; `GetProfileData` `:65-101`→`:65-99`; test `:70-78`→`:72-78`. Also §5's *"`grep EmailVerified internal/webui/` returns exactly those two sites"* actually returns 5 (plus `profile_handlers.go:47` and the generated `Profile_templ.go`) — both handled by §4 and by regeneration. And §5's mirrored `7 * 24 * time.Hour`, which the spec self-discloses in Risks and which I verified is truthful against `sessions.expires_at`.

### Fix needed

Re-derive §5's Copy-ID precedent against the post-`c0dc4f5` tree. Either cite `filter_components.templ:7` (the concat idiom that *does* still exist) instead of the deleted `StatisticsDashboard.templ:43` snippet, or — better, and what both #176 and issue #177 point at — switch the prescription to `data-user-id={ data.User.ID }` + `x-init="…$el.dataset.userId"`, matching the pattern the codebase standardized on one commit ago. Everything else in the spec, including the reworked badge analysis, checked out line-for-line.

**Note on the run:** `make test` fails on this host before it reaches docker-compose — `webui-full-rebuild` deletes every `*_templ.go` and then dies with `templ: No such file or directory` (Error 127). Not caused by this PR (the target has no dependency guard), but it leaves the working tree stripped of generated files. I restored them with `git checkout` and booted via `make docker-build-all` + `docker compose up -d`. That build was still queued behind three parallel QA slots after 85 min, so `/statistics` at this SHA was verified from source, the committed generated `.go`, and `git show` rather than a live render — the pre-rebase container above served as the control. All `/profile` evidence is live and valid: both `Profile.templ` and `profile_handlers.go` are unchanged since that image was built (last touched 2025-10-19 and 2026-07-24).

**Verdict: FAIL**

SoulKyu:
Closing: this spec served its purpose (issue #177 is closed). The spec text remains on `looper/planner/177-fix-webui-profile-page`.

PR #181, opened 2026-08-02
fix(webui): profile page renders fabricated account/session data and its Copy-ID button copies a literal template string

### LINKED ISSUE #177: fix(webui): profile page renders fabricated account/session data and its Copy-ID button copies a literal template string

SoulKyu:
## Summary

The `/profile` page hardcoded every date it rendered (`time.Now()` arithmetic in `profile_handlers.go`), showed an `EmailVerified` badge that only checked for a non-empty email string, carried a dead hardcoded `UserStats` block, and shipped a Copy-ID button that copied the literal text `{ data.User.ID }` because templ doesn't interpolate inside `<script>` blocks.

- `ValidateSession`/`GetProfile` (backend `AuthServiceGorm`) now also return `LastLogin` on the proto `User` (they already returned `CreatedAt`), matching the pattern already used in `SearchUsers`/`ListUsers`. No proto changes needed — `created_at`/`last_login` were already on the wire format, just dropped by the handlers.
- webui `client.User` now carries `CreatedAt`/`LastLogin`; `profile_handlers.go` renders those directly instead of inventing "3.5 months ago" / "2 hours ago".
- Removed the `EmailVerified` badge — there's no verification flow behind it, so a fabricated flag isn't a fair substitute.
- Removed the hardcoded `UserStats` block (`156`/`12`/`144`), which was already unrendered — dead trap for whoever wires the template to it next.
- The session card now shows only the `Session ID`, which is real. "Session Started" / "Expires" / "Duration" were sourced from constants (30 min ago, +7 days) with no backend record backing them, so they're dropped rather than invented — there's no RPC currently exposing the actual session row's timestamps.
- Copy-ID button: the user ID now travels via a `data-user-id` attribute (read in Alpine's `init()`), not a `<script>` interpolation that templ silently skips. The button swaps to a checkmark + "Copied" on success, mirroring the alert modal's copy-link button (`modal_components.templ`).

Closes #177

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/backend/services/... ./internal/webui/...` (225 passed)
- [x] Added `TestProfileDatesAreReal` covering `ValidateSession`/`GetProfile` returning real `CreatedAt`/`LastLogin`
- [x] `make webui-templates` regenerated `Profile_templ.go` from the edited `.templ`

### DIFF
--- a/internal/backend/services/profile_dates_test.go
+++ b/internal/backend/services/profile_dates_test.go
@@ -1,0 +1,42 @@
+package services
+
+import (
+	"context"
+	"testing"
+
+	authpb "notificator/internal/backend/proto/auth"
+)
+
+// TestProfileDatesAreReal guards against #177 regressing: ValidateSession and
+// GetProfile must return the account's actual created_at/last_login instead
+// of the webui handler inventing them with time.Now() arithmetic.
+func TestProfileDatesAreReal(t *testing.T) {
+	svc, sessionID := setupAuthServiceWithSession(t)
+	ctx := context.Background()
+
+	validate, err := svc.ValidateSession(ctx, &authpb.ValidateSessionRequest{SessionId: sessionID})
+	if err != nil {
+		t.Fatalf("ValidateSession: %v", err)
+	}
+	if validate.User.CreatedAt == nil {
+		t.Fatal("ValidateSession: CreatedAt is nil, want the account's real creation time")
+	}
+	if validate.User.LastLogin != nil {
+		t.Fatalf("ValidateSession: LastLogin = %v, want nil (never logged in)", validate.User.LastLogin)
+	}
+
+	if err := svc.db.UpdateLastLogin(validate.User.Id); err != nil {
+		t.Fatalf("UpdateLastLogin: %v", err)
+	}
+
+	profile, err := svc.GetProfile(ctx, &authpb.GetProfileRequest{SessionId: sessionID})
+	if err != nil {
+		t.Fatalf("GetProfile: %v", err)
+	}
+	if profile.User.CreatedAt == nil {
+		t.Fatal("GetProfile: CreatedAt is nil, want the account's real creation time")
+	}
+	if profile.User.LastLogin == nil {
+		t.Fatal("GetProfile: LastLogin is nil after UpdateLastLogin, want it populated")
+	}
+}
--- a/internal/backend/services/services.go
+++ b/internal/backend/services/services.go
@@ -243,6 +243,9 @@ func (s *AuthServiceGorm) ValidateSession(ctx context.Context, req *authpb.Valid
 	if user.Timezone != nil {
 		resp.User.Timezone = *user.Timezone
 	}
+	if user.LastLogin != nil {
+		resp.User.LastLogin = timestamppb.New(*user.LastLogin)
+	}
 	return resp, nil
 }
 
@@ -271,6 +274,9 @@ func (s *AuthServiceGorm) GetProfile(ctx context.Context, req *authpb.GetProfile
 	}
 	if user.Timezone != nil {
 		resp.User.Timezone = *user.Timezone
+	}
+	if user.LastLogin != nil {
+		resp.User.LastLogin = timestamppb.New(*user.LastLogin)
 	}
 	return resp, nil
 }
--- a/internal/webui/client/backend_client.go
+++ b/internal/webui/client/backend_client.go
@@ -49,12 +49,14 @@ type AuthResult struct {
 }
 
 type User struct {
-	ID            string  `json:"id"`
-	Username      string  `json:"username"`
-	Email         string  `json:"email"`
-	OAuthProvider *string `json:"oauth_provider,omitempty"`
-	OAuthID       *string `json:"oauth_id,omitempty"`
-	Timezone      *string `json:"timezone,omitempty"`
+	ID            string     `json:"id"`
+	Username      string     `json:"username"`
+	Email         string     `json:"email"`
+	OAuthProvider *string    `json:"oauth_provider,omitempty"`
+	OAuthID       *string    `json:"oauth_id,omitempty"`
+	Timezone      *string    `json:"timezone,omitempty"`
+	CreatedAt     time.Time  `json:"created_at"`
+	LastLogin     *time.Time `json:"last_login,omitempty"`
 }
 
 // IsOAuthUser returns true if the user was created via OAuth
@@ -310,6 +312,13 @@ func (c *BackendClient) ValidateSession(sessionID string) (*User, error) {
 	if resp.User.Timezone != "" {
 		user.Timezone = &resp.User.Timezone
 	}
+	if resp.User.CreatedAt != nil {
+		user.CreatedAt = resp.User.CreatedAt.AsTime()
+	}
+	if resp.User.LastLogin != nil {
+		lastLogin := resp.User.LastLogin.AsTime()
+		user.LastLogin = &lastLogin
+	}
 
 	return user, nil
 }
@@ -368,6 +377,13 @@ func (c *BackendClient) GetProfile(sessionID string) (*User, error) {
 	}
 	if resp.User.Timezone != "" {
 		user.Timezone = &resp.User.Timezone
+	}
+	if resp.User.CreatedAt != nil {
+		user.CreatedAt = resp.User.CreatedAt.AsTime()
+	}
+	if resp.User.LastLogin != nil {
+		lastLogin := resp.User.LastLogin.AsTime()
+		user.LastLogin = &lastLogin
 	}
 
 	return user, nil
--- a/internal/webui/handlers/profile_handlers.go
+++ b/internal/webui/handlers/profile_handlers.go
@@ -42,20 +42,11 @@ func ProfilePage(c *gin.Context) {
 			Email:         user.Email,
 			OAuthProvider: oauthProvider,
 			OAuthID:       user.OAuthID,
-			CreatedAt:     time.Now().AddDate(0, -3, -15),
-			LastLogin:     &[]time.Time{time.Now().Add(-2 * time.Hour)}[0],
-			EmailVerified: user.Email != "",
+			CreatedAt:     user.CreatedAt,
+			LastLogin:     user.LastLogin,
 		},
 		SessionInfo: pages.SessionInfo{
 			SessionID: sessionID,
-			CreatedAt: time.Now().Add(-30 * time.Minute),
-			ExpiresAt: time.Now().Add(7 * 24 * time.Hour),
-		},
-		Stats: pages.UserStats{
-			TotalAlerts:    156,
-			ActiveAlerts:   12,
-			ResolvedAlerts: 144,
-			LastActivity:   &[]time.Time{time.Now().Add(-5 * time.Minute)}[0],
 		},
 	}
 
--- a/internal/webui/templates/pages/Profile.templ
+++ b/internal/webui/templates/pages/Profile.templ
@@ -7,9 +7,8 @@ import (
 )
 
 type ProfileData struct {
-	User         ProfileUser
-	SessionInfo  SessionInfo
-	Stats        UserStats
+	User        ProfileUser
+	SessionInfo SessionInfo
 }
 
 type ProfileUser struct {
@@ -20,20 +19,10 @@ type ProfileUser struct {
 	OAuthID       *string
 	CreatedAt     time.Time
 	LastLogin     *time.Time
-	EmailVerified bool
 }
 
 type SessionInfo struct {
-	SessionID  string
-	CreatedAt  time.Time
-	ExpiresAt  time.Time
-}
-
-type UserStats struct {
-	TotalAlerts      int
-	ActiveAlerts     int
-	ResolvedAlerts   int
-	LastActivity     *time.Time
+	SessionID string
 }
 
 templ Profile(data ProfileData) {
@@ -41,7 +30,7 @@ templ Profile(data ProfileData) {
 }
 
 templ ProfileContent(data ProfileData) {
-	<div class="min-h-screen bg-gray-50 dark:bg-dark-bg-primary py-8" x-data="profilePage()">
+	<div class="min-h-screen bg-gray-50 dark:bg-dark-bg-primary py-8" x-data="profilePage()" data-user-id={ data.User.ID }>
 		<div class="max-w-4xl mx-auto px-4 sm:px-6 lg:px-8">
 			<!-- Header -->
 			<div class="mb-8">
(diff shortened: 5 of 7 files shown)


SoulKyu (commented):
@SoulKyu Reviewed the profile page fix for #177: fabricated dates, the bogus EmailVerified badge, the dead UserStats block, and the Copy-ID button's `<script>` interpolation bug.

Verified against the diff:
- `AuthServiceGorm.ValidateSession`/`GetProfile` now populate `LastLogin` on the proto `User` the same way `SearchUsers`/`ListUsers` already do (internal/backend/services/services.go), and `CreatedAt` was already set unconditionally, so both fields flow through cleanly with no wire/schema changes.
- `webui/client.User` and `profile_handlers.go` propagate those real values end-to-end; the `RequireAuth`/`OptionalAuth` middleware path that populates `GetCurrentUserFromContext` goes through `ValidateSession`, so the profile page actually receives real data rather than invented timestamps.
- `EmailVerified` and `UserStats`/session-duration fields are fully removed from `pages.ProfileUser`/`ProfileData`, and I confirmed no other file in the repo still references them (grep for `EmailVerified`/`UserStats`/`FormatDuration` only turns up unrelated OAuth model fields).
- The Copy-ID fix replaces the `<script>` block's `userId: '{ data.User.ID }'` (which templ never interpolates) with a `data-user-id` attribute read via Alpine's `init()` — the attribute value goes through templ's normal HTML-escaping path, so this closes the bug without introducing an XSS gap.
- `internal/webui/templates/utils.go`: `FormatDuration` is deleted along with its only caller, no dangling references.
- Regenerated templates with `make webui-templates` locally — `Profile_templ.go` matches the committed version exactly (no diff), confirming the generated file is in sync with the `.templ` source.
- Ran `go build ./...`, `go vet ./...`, and `go test ./internal/backend/services/... ./internal/webui/...` — all green (225 tests passing), including the new `TestProfileDatesAreReal` regression test.

Nice, thorough fix — appreciate that it went after the root cause on the backend (missing `LastLogin` passthrough) instead of just patching the template, and that the dead `UserStats`/session-duration code was removed rather than left as another trap. No actionable findings.



SoulKyu:
## 🧪 QA Report

Env: full docker-compose rebuild at `1e5fc9a`, webui on :10081, postgres on :7432. Values cross-checked against the `users` table in postgres. Browser flows driven with Playwright (chromium).

|---|---|---|
| "Member Since" / "Last Login" show the real `created_at` / `last_login` | ✅ | Registered `qauser181`; DB row `created_at=2026-08-02 20:14:51Z`, `last_login=20:16:11Z` → page renders `Aug 2, 2026 at 8:14 PM` / `at 8:16 PM`. Re-login bumped "Last Login" to the new value. |
| "ID" button copies the actual user ID, with visible feedback | ✅ | Stubbed `navigator.clipboard.writeText` captured exactly `hwaffyq9g82p1fwijcsqdknde1u2wfct` (= `users.id`); `Alpine.$data(el).userId` matches; button swaps to green check + "Copied" and reverts to "ID" after 2s. No `{ data.User.ID }` anywhere in the served HTML (grep count 0). |
| No date/badge/counter on `/profile` from `time.Now()` or hardcoded constants | ✅ | Rendered HTML grep: `Verified`=0, `Session Started`=0, `Expires`=0, `Duration`=0, `156`=0, `144`=0, `showToast`=0. Session card renders only the real Session ID. Proven dynamic: DB-forced `created_at=2024-01-15 09:05Z` / `last_login=2025-06-30 23:45Z` rendered verbatim (`Jan 15, 2024 at 9:05 AM` / `Jun 30, 2025 at 11:45 PM`) — impossible with the old `time.Now()` arithmetic. |
| "Verified" badge backed by a real flag or removed | ✅ | Removed: absent from the DOM and from the served HTML; `EmailVerified` gone from `ProfileUser`. Screenshot shows only the auth-method badge ("Classic Auth"). |

**Adversarial pass:**

- **Boundary / empty** — `last_login = NULL` (account that never logged in / pre-change row): the "Last Login" row is omitted entirely, nothing fabricated in its place. `created_at = NULL` forced directly in postgres renders `Jan 1, 0001 at 12:00 AM` (Go zero time) — only reachable by corrupting the row (GORM always populates it), so noted below rather than scored.
- **Persisted state / migration on an existing DB** — set `created_at`/`last_login` to 2024/2025 values on an existing row with a live session: rendered exactly those, no "3.5 months ago"/"2 hours ago" drift. This is the case the old handler could never pass.
- **Auth & scope** — unauthenticated `GET /profile` → `302 /login`; session forced expired in `sessions.expires_at` → `302 /login` (no stale render). Registered a second user (`qauser181b`): its page shows only its own ID and its own dates, no leak of user A's 2024 timestamp.
- **Error / unavailable path** — `docker compose stop backend` mid-flow, then `GET /profile` with a valid cookie → `503` with a 31-byte body. It does **not** render a profile shell with blank or zero dates, i.e. no wiped state pushed as success.
- **Malformed & hostile** — registered a user named `"` with email `x"><script>alert(2)</script>@…`, then loaded `/profile`: no dialog fired, no `img[src=x]` in the DOM, no page errors, and the Copy-ID button still copied that user's correct ID. The `data-user-id` attribute is templ-escaped.
- **Copy fallback path** — forced `writeText` to reject (headless clipboard permission denied): the `.catch` → `execCommand('copy')` fallback ran and "Copied" feedback still appeared.
- **Adjacent flows** — dashboard user menu → `/profile` still 200; "Sign Out" on the profile page still logs out and lands on `/` (the `handleLogoutResponse` handler survived the script rewrite); "Change Password" still gated on non-OAuth accounts; timezone API (`PUT /api/v1/profile/timezone`) still 200.

Non-blocking observations (not scored against the criteria):
1. `navigator.clipboard` is unguarded, so over plain HTTP on a non-localhost host (insecure context, e.g. `http://192.168.3.61:10081`) clicking "ID" throws an uncaught `TypeError: Cannot read properties of undefined (reading 'writeText')`, copies nothing and shows no feedback — the `.catch` fallback can't fire because the throw is synchronous. Works on localhost/HTTPS. Same unguarded pattern already exists in `StatisticsDashboard.templ` (3 call sites), so this is the app's baseline, not a regression from this PR; a `navigator.clipboard?.writeText(...)` + `try/catch` would make the fallback the author wrote actually reachable.
2. `GetProfileData` in `profile_handlers.go` still returns hardcoded `acknowledged_alerts: 42 / comments: 17 / color_preferences: 3`. It is not registered on any route (`router.go` only wires `GET /profile`), so nothing renders it — dead code, same trap the issue flagged for `UserStats`.
3. Profile dates render in UTC with no zone label, while the footer timezone selector shows e.g. `Europe/Paris`. The values are correct; no page in the app formats timestamps against the saved timezone, so this is app-wide, not specific to this change.

**Verdict: PASS**

SoulKyu:

SoulKyu (commented):
@SoulKyu Reviewed this at head `1d1a6554` (a clean rebase onto `main` of the previously-reviewed `1e5fc9a`, no code delta). Verified independently against the diff and by running the suite locally:

- `AuthServiceGorm.ValidateSession`/`GetProfile` (internal/backend/services/services.go) now set `resp.User.LastLogin` via the same nil-guarded `timestamppb.New(*user.LastLogin)` pattern already used by `SearchUsers`/`ListUsers` — no proto/wire changes needed since `last_login` was already on `authpb.User`.
- `webui/client.User` gained `CreatedAt`/`LastLogin` and both RPC call sites populate them from the proto response; `profile_handlers.go` now passes `user.CreatedAt`/`user.LastLogin` straight through instead of `time.Now().AddDate(...)`/`time.Now().Add(-2*time.Hour)`.
- `EmailVerified` (fake, derived only from `email != ""`) and the entire hardcoded `UserStats`/session-duration block are removed from `pages.ProfileUser`/`ProfileData`, along with the now-dead `FormatDuration` helper — grepped the repo and confirmed nothing else references them.
- Copy-ID fix: `data-user-id={ data.User.ID }` is now a templ-escaped HTML attribute read in Alpine's `init()`, replacing the old `<script>` block's `userId: '{ data.User.ID }'` which templ never interpolated (confirmed templ doesn't expand expressions inside `<script>`, so the old code really did copy the literal string).
- `go build ./...` and `go test ./internal/backend/services/... ./internal/webui/...` both pass locally (256 tests), including the new `TestProfileDatesAreReal` regression guard.

This also lines up with the prior review and the QA agent's adversarial pass on `1e5fc9a` (empty/nil `last_login`, cross-user isolation, XSS via malicious username, clipboard-fallback path, backend-down path) — all PASS, and since this head is only a rebase of that exact commit, those results still hold. Nice fix — going after the root cause on the backend (missing `LastLogin` passthrough) rather than just patching the template, and removing the dead `UserStats`/duration code instead of leaving it as a trap, is exactly the right call. No actionable findings.



SoulKyu:
## 🧪 QA Report

Env: images rebuilt from this clone at `1d1a655` (webui + backend, backend with `--no-cache` after a cached cross-slot image was found serving stale code — the first run showed old behaviour purely because of that image, not the PR). Stack on webui :10081 / postgres :7432. Values cross-checked against the `users` / `sessions` tables in postgres. Browser flows driven with Playwright (chromium).

|---|---|---|
| "Member Since" / "Last Login" show the real `created_at` / `last_login` | ✅ | Fresh user `qa181a`: DB `created_at=2026-08-03 03:26:28Z`, `last_login=03:43:28Z` → page renders `Aug 3, 2026 at 3:26 AM` / `at 3:43 AM`. Re-login moved "Last Login" to `03:45` and the page followed. |
| "ID" button copies the actual user ID, with visible feedback | ✅ | `navigator.clipboard.writeText` stubbed: captured exactly `4ofrv2u8wvxdd0ooak7vzwteovec2qmf` (= `users.id`); `Alpine.$data(el).userId` matches; label swaps `ID → Copied` with the green check (`svg.text-green-500` visible) and reverts after 2s. Served HTML contains `data-user-id="4ofrv…"` and zero occurrences of `{ data.User.ID }`. |
| No date/badge/counter on `/profile` from `time.Now()` or hardcoded constants | ✅ | Rendered HTML: `Verified`=0, `Session Started`=0, `Expires`=0, `Duration`=0, `156`=0, `144`=0, `showToast`=0; session card renders only the real Session ID. Proven dynamic: `created_at` forced to `2019-03-07 14:02Z` and `last_login` to `2021-11-30 23:58Z` in postgres rendered verbatim (`Mar 7, 2019 at 2:02 PM` / `Nov 30, 2021 at 11:58 PM`) — impossible with the old `time.Now()` arithmetic. `time.Now` count in `profile_handlers.go` = 0. |
| "Verified" badge backed by a real flag or removed | ✅ | Removed — absent from DOM and from the served HTML; only the auth-method badge ("Classic Auth") remains (screenshot). |

**Adversarial pass:**

- **Boundary / empty** — `last_login = NULL` (account that never logged in, or a pre-change row): the "Last Login" row is omitted entirely, nothing invented in its place. Dates far in the past (2019/2021) render verbatim rather than being clamped or recomputed.
- **Persisted state / existing DB** — forced 2019/2021 timestamps on an existing row holding a live session, then reloaded `/profile`: exactly those values, no "3.5 months ago" / "2 hours ago" drift. This is the case the old handler could never pass.
- **Auth & scope** — unauthenticated `GET /profile` → `302 /login`; forged cookie `session_id=deadbeef` → `302 /login`; `sessions.expires_at` forced into the past → `302 /login` with a 29-byte body (no stale render). Second account (hostile-named, see below) shows only its own ID (`b0orkgn8sv…`) and its own `last_login`, while user A's was NULL at the same moment — no cross-user leak.
- **Error / unavailable path** — `docker compose stop backend`, then `GET /profile` with a valid cookie → `503 Service temporarily unavailable` (31 bytes). It does **not** render a profile shell with blank/zero dates, i.e. no wiped state pushed as success.
- **Malformed & hostile** — registered `"qa181b` with email `x"><script>alert(2)</script>@example.com`, logged in and loaded `/profile`: no dialog fired, username/email rendered escaped (`&lt;img src=x …`), no page errors, and the Copy-ID button still copied that user's correct ID. The `data-user-id` attribute goes through templ's escaping.
- **Copy fallback path** — forced `clipboard.writeText` to reject: the `.catch` → `document.execCommand('copy')` fallback ran (`exec: ["copy"]`) and "Copied" feedback still appeared.
- **Adjacent flows** — `/dashboard` still links to `/profile` (200); "Sign Out" from the profile page logs out and lands on `/`, and `/profile` afterwards redirects to `/login`; "Change Password" still fires its pre-existing TODO alert; `GET /api/v1/profile/timezone` → 200. No failed requests at all on `/profile` itself (network log empty for status ≥ 400).

Non-blocking observations (not scored against the criteria):
1. `navigator.clipboard` is unguarded: with the API absent (plain HTTP on a non-localhost host = insecure context) the click throws `TypeError: Cannot read properties of undefined (reading 'writeText')`, copies nothing and shows no feedback — the author's `.catch` fallback can't fire because the throw is synchronous. Reproduced by deleting `navigator.clipboard`. Not a regression (the pre-PR code made the same unguarded call) and the same pattern exists in `StatisticsDashboard.templ`; `navigator.clipboard?.writeText(...)` inside a `try/catch` would make the fallback reachable.
2. `GetProfileData` in `profile_handlers.go` still returns hardcoded `acknowledged_alerts: 42 / comments: 17`. It is not wired to any route (`router.go` only registers `GET /profile` → `ProfilePage`), so nothing renders it — dead code, the same trap the issue flagged for `UserStats`.
3. Profile dates render in UTC with no zone label while the footer selector shows e.g. `Europe/Paris`. Values are correct; no page in the app formats timestamps against the saved timezone, so this is app-wide rather than specific to this change.

**Verdict: PASS**

PR #182, opened 2026-08-02
feat(auth): let classic-auth users change their password

### LINKED ISSUE #178: feat(auth): classic-auth users cannot change their password — the profile button is a dead-end alert()

SoulKyu:
## Summary

- Adds a session-gated `ChangePassword` RPC to `AuthService` (`proto/auth.proto`): verifies the old bcrypt hash, rejects OAuth-backed accounts, hashes and persists the new password, and invalidates the user's other sessions on success (current session stays alive).
- Adds `POST /api/v1/auth/change-password` on the WebUI, calling the new RPC through `BackendClient`.
- Replaces the dead `alert('Change password functionality coming soon!')` stub in `Profile.templ` with a real modal (current/new/confirm password, inline error/success states).
- The RPC is unreachable without a valid session — it isn't in the interceptor's public allowlist, so the deny-by-default gate from #172 covers it automatically.

## Why

The Profile page teased a "Change Password" button on every classic-auth account, but clicking it just showed an alert box — there was no backend RPC, no handler, and no way to rotate a password short of writing to the database by hand. That made the quickstart's "log in with `admin:admin` (change it)" instruction impossible to follow in mixed/classic deployments (the default docker-compose and Helm setups).

Closes #178

## Test plan

- [x] `go build ./...`, `go vet ./...`, `go test ./...` (307 passed)
- [x] New unit tests (`internal/backend/services/change_password_test.go`): wrong old password rejected, correct old password rotates the hash (old password stops authenticating, new one works), current session survives while other sessions are invalidated, OAuth-backed accounts are rejected server-side
- [x] `make webui-templates` equivalent (pinned `templ` v0.3.906) regenerated only `Profile_templ.go`
- [ ] Manual browser click-through of `/profile` (not run in this environment — no live docker/browser available)


SoulKyu (commented):
@SoulKyu Thanks for closing out #178 — this is a solid, well-scoped addition.

What I verified:
- `ChangePassword` in `internal/backend/services/services.go` correctly gates on `classicAuthDisabled()`, validates the session via `GetUserBySession`, rejects OAuth-only/passwordless accounts, requires and bcrypt-verifies the old password, enforces a minimum new-password length consistent with `Register`'s existing rule, and on success rotates the hash and invalidates every other session for the user while keeping the current one alive.
- Confirmed the RPC is not present in `publicMethods` in `internal/backend/auth_interceptor.go`, so the deny-by-default gate does cover it, as claimed.
- Ran `go build ./...`, `go vet`, and `go test ./internal/backend/services/... -run TestChangePassword -v`: both new tests (`TestChangePassword`, `TestChangePasswordRejectsOAuthUser`) pass, including the "current session survives, other sessions invalidated" assertion.
- Walked the WebUI path end-to-end (`router.go` → `profile_handlers.go` → `backend_client.go`) — the route sits behind `RequireAuth()`, session ID plumbing matches the pattern already used by `UpdateTimezone`/`ProfilePage`, and the JSON response shape (`success`/`error`) lines up with what `Profile.templ`'s `changePassword()` expects.
- Checked the regenerated `Profile_templ.go` against `Profile.templ` — the diff is consistent (large single-line string literals just make `git diff --stat` look small), no stray manual edits.
- Looked for an edge case where a classic-auth user who later links an OAuth account (`LinkOAuthToExistingUser` keeps `PasswordHash` while setting `OAuthProvider`) would get incorrectly blocked by the `IsOAuthUser()` check — but that linking path is never actually invoked from the OAuth callback flow (`CreateOrUpdateOAuthUser` only creates/updates, never links), so it's not a reachable state today. Worth keeping in mind if that linking path ever gets wired up.

No blocking or non-blocking issues found. Nice work — the modal UX (loading/disabled state, inline error/success) is a clean replacement for the old stub alert.



SoulKyu:
## 🧪 QA Report

Env: full `docker compose` stack rebuilt from this SHA (backend + webui images built from source), postgres volume created fresh, browser flows driven with Playwright/Chromium, gRPC probed directly on `:51051` with a Go client using the repo's generated stubs.

|---|---|---|
| A classic-auth user can change their password from `/profile` with the current password; the new one works on next login, the old one doesn't | ✅ | Browser: `/profile` → "Change Password" → filled current/new/confirm → green "Password changed successfully."; `POST /api/v1/auth/login` with the old password → `401 Invalid credentials`, with the new one → `200`. Current browser session stayed alive (`/api/v1/auth/me` → 200) while the user's two other sessions were deleted (both → `401 Invalid or expired session`, `sessions` table went 3→1 for that user). Another user's session (bob) survived untouched. |
| Wrong current password is rejected with a visible error (no silent success), and the RPC is unreachable without a valid session | ✅ | Modal showed red banner **"Current password is incorrect"**, no success state, password unchanged (old password still logs in). Unauthenticated `POST /api/v1/auth/change-password` → `401 {"error":"Authentication required"}`. Direct gRPC `ChangePassword` with no metadata → `Unauthenticated: /notificator.auth.AuthService/ChangePassword requires a session or service token`; with a garbage `session_id` in the body → `Invalid session`; with an empty one → `Session ID is required`. |
| OAuth-only accounts never see the button and the RPC refuses them server-side | ✅ | Button is templ-gated on `data.User.OAuthProvider == nil` and the OAuth login path sets `auth_method=oauth:<provider>`, so an OAuth session renders no button. Server-side: a user row with an OAuth provider and empty `password_hash` + a live session → RPC returns `success=false "OAuth accounts do not have a password to change"` (both with an empty and a guessed `old_password`). |
| The `admin:admin` bootstrap credential can be rotated end-to-end through the UI | ✅ | `admin`/`admin` (created against the *pre-PR* backend image, then the stack was upgraded to this SHA — no migration needed, `password_hash` is updated in place) was rotated twice through the browser modal (`admin`→`admin2`, later `→adminfinal` submitted with the Enter key). Login with the rotated credential works; the old one is dead. |

**Adversarial pass**

- **Boundary / empty**: 3-char new password → client-side "must be at least 4 characters long" and, bypassing the UI, the server returns the same (parity with `Register`). Exactly-72-byte password → accepted and logs in. **73 bytes → `Internal server error`** (bcrypt's 72-byte cap surfaced as an opaque message) — *not* a fail: the password is left untouched (old one still logs in) and `Register` has the identical behaviour, so no state is reachable that the user cannot re-enter. New password == old password → allowed (rotation still succeeds, sessions still invalidated). Mismatched confirmation → "New password and confirmation do not match", no request sent.
- **Malformed / hostile**: truncated JSON, wrong types (`"old_password":123`, object as new password), missing `new_password`, empty strings, form-encoded body, and a 1 MiB password all → `400` with a clean error, no 5xx, no panic in backend logs. Error text is rendered with `x-text` (escaped), and the only strings the server returns are static.
- **Error / unavailable path**: `docker stop` of the backend with the modal open, then submit → after the gRPC deadline the modal shows red **"Authentication service unavailable"**, never a success banner, and the password was verifiably *not* changed (the attempted new value fails to log in afterwards, the previous one still works). App-level "Maintenance in progress" overlay also appears. On recovery the same submit succeeds.
- **Auth & scope**: covered above; additionally, a second user's sessions are not touched when someone else rotates (`user_id = ? AND id != ?` verified live). CSRF: cookie is `SameSite=Lax`, and the handler only accepts `application/json` (form-encoded cross-site shape → `400`), so a cross-origin form POST cannot drive it.
- **Adjacent flows**: after the change — Sign Out still logs out and redirects, re-login works, `GET`/`PUT /api/v1/profile/timezone` still `200`, no page errors in the console. Modal state is fully reset on close/reopen (all three fields empty, no stale error/success banner). Enter in the confirm field submits.
- **Persisted state**: the users exercised were created by the *previous* backend image and rotated fine after the upgrade; `find -name '*_templ.go' -delete && templ generate` reproduces the committed `Profile_templ.go` byte-for-byte (no stale generated file).

Two things that did **not** break a criterion but are worth the author's eye:

1. `ChangePassword` resolves the user from `req.SessionId` in the request *body* and ignores the caller the interceptor already authenticated. A probe authenticated as `bob` (bob's session in `x-notificator-session`) but carrying **admin's** session id in the body successfully changed admin's password. This only escalates for someone who already holds the victim's session token (itself a full compromise), and every other handler in this service follows the same `req.SessionId` pattern — so it's a hardening gap, not a new hole. A `caller.user.ID == user.ID` check via `backend.AuthenticatedUser(ctx)` would close it.
2. An account holding *both* a password hash and an OAuth link would still see the button while the server always refuses it (dead-end control, the very defect #178 is about). Not reachable today — `CreateOrUpdateOAuthUser` never links an OAuth identity onto an existing classic account, it creates a new user — so this is only a latent trap if account linking is ever added.

**Verdict: PASS**

SoulKyu:

internal/webui/templates/pages/Profile.templ
```diff
@@ -263,9 +335,57 @@ templ ProfileContent(data ProfileData) {
...
+						this.currentPassword = '';
+						this.newPassword = '';
+						this.confirmPassword = '';
+					} catch (err) {
+						console.error('Failed to change password:', err);
+						this.changePasswordError = 'Failed to change password';
+					} finally {
+						this.changePasswordLoading = false;
+					}
 				}
```
SoulKyu, line 385: `changePassword()`'s success path can overwrite modal state after the user has already cancelled, leaving stale "success" state for the next time the modal opens.

Problem: only the submit button is gated on `changePasswordLoading` (`:disabled="changePasswordLoading"` on the button at line 281). The backdrop `@click="closeChangePasswordModal()"` (line 239), the header "X" button (line 244), and the "Cancel" button (lines 291-294) all call `closeChangePasswordModal()` unconditionally, with no check on `changePasswordLoading`. `x-show` only toggles visibility — it doesn't destroy the Alpine component or abort the in-flight `fetch`.

Concrete repro: open the modal, enter a correct current password and a valid new password, click "Change Password", then immediately click "Cancel" (or the backdrop) before the network round trip finishes. `closeChangePasswordModal()` (lines 338-346) resets `showChangePasswordModal`, `changePasswordSuccess`, and the password fields right away. The backend request is not aborted, so the password change still goes through server-side even though the user believes they cancelled. When the promise resolves, lines 379-382 unconditionally set `this.changePasswordSuccess = true` and clear the fields again, overwriting the just-performed reset. Nothing is visibly wrong immediately (the modal is already hidden), but the next time the user opens the modal via the profile button (`showChangePasswordModal = true`, which does not reset any of this state), they land straight on the "Password changed successfully" screen without having submitted anything in that new session.

Why it matters: this both masks that the earlier "cancelled" attempt actually rotated the user's password, and shows a false success confirmation on an unrelated later open of the modal — confusing UI state around a security-sensitive action.

Suggested fix: in the `try` block of `changePassword()`, guard the state mutations on the modal still being open (e.g. `if (!this.showChangePasswordModal) return;` right after `await response.json()`, or capture a local token at call-start and compare before applying results), and/or disable the backdrop click, the "X" button, and the "Cancel" button while `changePasswordLoading` is true, the same way the submit button already is.

SoulKyu (commented):
Reviewed the classic-auth change-password feature end to end: proto (`ChangePassword` RPC), `AuthServiceGorm.ChangePassword` in `internal/backend/services/services.go`, the gRPC interceptor's public-method allowlist, the webui handler/client wiring, and the new Alpine modal in `Profile.templ`.

The backend logic holds up under adversarial checks: wrong old password is rejected, OAuth-only accounts are correctly blocked (`IsOAuthUser()`/`HasPassword()`), the new-password minimum length matches `Register`'s existing rule, the current session survives while other sessions are hard-deleted via `InvalidateUserSessions` (`Session` has no soft-delete field, so this is a real delete, confirmed against `GetUserBySession`'s query), and `ChangePassword` is correctly absent from `publicMethods` in `auth_interceptor.go` so the deny-by-default gate covers it as the PR description claims. Test coverage in `change_password_test.go` matches these paths.

One non-blocking correctness issue found in the frontend modal's async state handling — see inline comment.



internal/webui/templates/pages/Profile.templ
SoulKyu: @SoulKyu This has been fixed in 0afebf3.

Added `if (!this.showChangePasswordModal) return;` right after awaiting the JSON response in changePassword() (Profile.templ), and disabled the backdrop click, X button, and Cancel button while changePasswordLoading is true, mirroring the existing submit-button guard.




SoulKyu:
**Looper fixer round complete** — 0afebf3

- ✅ Review comment on `internal/webui/templates/pages/Profile.templ:390` (@SoulKyu) — [thread](https://github.com/SoulKyu/notificator/pull/182#discussion_r3701057036)
  - Added `if (!this.showChangePasswordModal) return;` right after awaiting the JSON response in changePassword() (Profile.templ), and disabled the backdrop click, X button, and Cancel button while changePasswordLoading is true, mirroring the existing submit-button guard.


SoulKyu:
## 🧪 QA Report

Env: full `docker-compose` stack rebuilt from this SHA (webui :9081, backend gRPC :51051, postgres :6432). Real browser flow driven with Playwright/Chromium, plus raw `curl` and direct `grpcurl` against the backend.

|---|---|---|
| Classic-auth user changes their password from `/profile`; new one works on next login, old one doesn't | ✅ | Registered `admin:admin`, logged in through the real login form, opened the `/profile` modal, submitted `admin` → `NewPass123`, got the green "Password changed successfully." banner. `POST /api/v1/auth/login` with `admin` → **401 Invalid credentials**, with `NewPass123` → **200**. Current session stayed alive (`/profile` still 200); the two other pre-existing sessions were killed (`/api/v1/auth/me` → 401, `sessions` rows dropped 3 → 1). |
| Wrong current password rejected with a visible error (no silent success), RPC unreachable without a valid session | ✅ | Modal with a wrong current password → red **"Current password is incorrect"**, no success banner, HTTP 400, password unchanged. gRPC `ChangePassword` with **no metadata** → `Unauthenticated: … requires a session or service token`; with a **bogus session** → `Unauthenticated: invalid or expired session`; over HTTP with no cookie / a garbage cookie → **401**. |
| OAuth-only accounts never see the button and the RPC refuses them server-side | ✅ | Seeded an OAuth-only user (`o_auth_provider=github`, empty `password_hash`) + a live session; gRPC `ChangePassword` → **"OAuth accounts do not have a password to change"**. Button visibility is driven by the session's `auth_method` (`oauth:<provider>`, set only by the OAuth callback), so a real OAuth session never renders it. |
| `admin:admin` bootstrap credential rotated end-to-end through the UI | ✅ | The rotation above was done entirely through the browser modal, no DB writes. Survived `docker compose restart backend webui` on the **existing** volume: `admin` → 401, rotated password → 200. No re-seeding of the default credential on restart. |

**Adversarial pass** — what I tried to break each criterion with:

- **Boundary / empty**: empty body, missing `new_password`, `new_password:""` → 400 "current and new password are required". 3-char password → "must be at least 4 characters" (client *and* server). New password identical to the current one → accepted (200), harmless. Exactly-72-byte password → works.
- **Malformed / hostile**: truncated JSON `{"old_password":` → 400. Wrong types (`old_password:123`, `new_password:["a"]`) → 400, no panic. New password `'; DROP TABLE users;--` → stored literally, logs in fine afterwards, `users` table intact (4 rows). 1 MB `old_password` → 400, app still serving. Zero `panic`/`fatal` in backend logs across the whole run.
- **Error / unavailable path**: `docker compose stop backend` mid-flow → the modal sits on "Changing…" for ~10s (gRPC deadline) and then renders **"Authentication service unavailable"** in red — no fake success, no silent close. Password confirmed unchanged after the backend came back (`WhileDown1` → 401, previous password → 200).
- **Auth & scope**: unauthenticated POST → 401. Forged cookie → 401. **Expired session** (`expires_at` backdated in postgres) → 401 "Invalid or expired session", password unchanged. Replay of a successful request with the now-stale old password → 400 "Current password is incorrect". Cross-user: `bob` rotating his own password left admin's password and admin's sessions untouched.
- **Adjacent flows**: `/dashboard` 200, `/profile` 200, `GET`/`PUT /api/v1/profile/timezone` 200, logout 200 then `/api/v1/auth/me` 401 — all still fine after the change. Client-side guards still fire correctly: empty fields → "All fields are required", mismatched confirm → "New password and confirmation do not match".
- **Persisted state**: rotation verified on a database that already held pre-change users and sessions, and re-verified after a full backend+webui restart.

Non-blocking observations (none of these break a criterion):

1. **`session_id` is taken from the request body, not from the authenticated caller.** With `bob`'s session in gRPC metadata but **admin's** session id in the body, the RPC operated on **admin** and returned `success: true` once admin's current password was supplied. It is not a privilege escalation — you need the victim's live session id *and* their current password, i.e. an already-compromised account — but the handler never cross-checks against `backend.AuthenticatedUser(ctx)`, which the interceptor populates exactly for this. Worth tightening.
2. **New password > 72 bytes → "Internal server error"** (`bcrypt.ErrPasswordTooLong`, x/crypto v0.39.0), password unchanged. Misleading message for a user error, but `Register` behaves identically on the same input, so this is pre-existing product-wide behaviour, not a regression from this PR.
3. **OAuth-linked accounts that also have a password** are refused too ("OAuth accounts do not have a password to change"). The issue only requires OAuth-*only* accounts to be refused, so this is in spec, but such accounts cannot rotate their classic password.
4. Build note: `make test` needs `templ` on `PATH` and its `clean` step deletes every `*_templ.go` first — it failed with `Error 127` here until templ was installed. Regenerating with the pinned `templ v0.3.906` produced output **byte-identical** to the committed `Profile_templ.go`.

**Verdict: PASS**

SoulKyu:
## 🧪 QA Report

Env: full `docker compose` stack rebuilt from this SHA (webui :9081, backend gRPC :51051, postgres :6432, fresh volume). Browser flows driven with Playwright/Chromium, plus raw `curl` and `grpcurl` straight at the backend. Re-test of the fixer commit `0afebf3` on top of the previously-passing `fa65f11`.

|---|---|---|
| A classic-auth user can change their password from `/profile` with the current password; the new one works on next login, the old one doesn't | ❌ | The rotation itself works (registered `admin:admin`, logged in through the real form, `admin` → `NewPass123` via the modal, green banner; `login` with `admin` → **401**, with `NewPass123` → **200**). **But** on a connection with normal latency, a second <kbd>Enter</kbd> while the request is in flight makes a *successful* change report **"Current password is incorrect"** in red, next to the green success banner. Reproduced 3/3. Details below. |
| Wrong current password rejected with a visible error (no silent success), RPC unreachable without a valid session | ✅ | Wrong current password in the modal → red **"Current password is incorrect"**, HTTP 400, no success banner, password unchanged. gRPC with **no metadata** → `Unauthenticated`; **bogus session** → `Unauthenticated: invalid or expired session`; **expired session** (`expires_at` backdated in postgres) → rejected on both gRPC and HTTP (401), password unchanged. HTTP with no cookie / forged cookie → **401**. |
| OAuth-only accounts never see the button and the RPC refuses them server-side | ✅ | Seeded a real OAuth-only user (`o_auth_provider=github`, `o_auth_id`, empty `password_hash`) + live session → gRPC `ChangePassword` → **"OAuth accounts do not have a password to change"**. Also verified the *other* refusal branch (OAuth user that also has a password hash → same refusal), so both `IsOAuthUser()` and `!HasPassword()` gates fire. Button visibility is server-side (`ProfilePage` derives `OAuthProvider` from the session's `auth_method` `oauth:` prefix, set only by the OAuth callback) — see caveat below. |
| The `admin:admin` bootstrap credential can be rotated end-to-end through the UI | ✅ | Done entirely through the browser modal, zero DB writes. Survived `docker compose restart backend webui` on the **existing** volume: `admin` → 401, rotated password → 200, no re-seeding of the default credential. |

### ❌ The failure: `@keydown.enter` has no in-flight guard

`Profile.templ:277` binds `@keydown.enter="changePassword()"` on `#confirm-password`. This commit added a `changePasswordLoading` guard to the backdrop, the X button and Cancel — **but not to this Enter handler**, and not to `changePassword()` itself. So Enter still fires a second request while the first is in flight.

Exact repro (3/3):
1. 400 ms RTT (CDP `Network.emulateNetworkConditions` — no request interception, no mocking; any non-localhost deployment qualifies).
2. `/profile` → Change Password → fill current/new/confirm.
3. Focus the confirm field, press <kbd>Enter</kbd>, wait ~300 ms (still showing "Changing…"), press <kbd>Enter</kbd> again.

Observed: two real requests. `#1` → `200 {"success":true}`; `#2` is evaluated against the *already-rotated* password → `400 {"error":"Current password is incorrect"}`. The stale guard `if (!this.showChangePasswordModal) return;` doesn't help — the modal is still open — so the modal renders **both** banners at once:

> 🔴 Current password is incorrect
> 🟢 Password changed successfully.

The password *is* changed, but the user is told their current password was wrong. Anyone trusting the red banner concludes it failed and keeps using the old password, which no longer works. This is the same stale-response class the commit set out to fix, on the one path it didn't cover. Also reachable as a second variant: when response #1 lands first it clears the fields, and the second Enter then paints **"All fields are required"** over the success banner (seen at gap 400–600 ms).

Not reproducible on localhost (sub-ms RTT): both requests land inside the same read-modify-write window and both return 200, which is why it survived the fix.

### Adversarial pass

- **The head commit's own fix — verified working.** With the backend SIGSTOPped mid-request: Cancel `disabled=true`, X `disabled=true`, force-clicking Cancel + X + backdrop all inert, modal stayed open, and on unpause the response landed correctly on the still-open modal. No stale overwrite via those three paths.
- **Boundary / empty**: empty body, missing `new_password`, `new_password:""` → 400 "current and new password are required". 3-char → "must be at least 4 characters" (client *and* server). Password identical to the current one → 200, harmless. 73-byte password → "Internal server error" (bcrypt's 72-byte limit) — `Register` behaves identically on the same input, so pre-existing product-wide, not a regression.
- **Malformed / hostile**: truncated JSON `{"old_password":`, wrong types (`old_password:123`, `new_password:["a","b"]`, `null`) → 400, no panic. 1 MB `old_password` → 400, app still serving. New password `x'; DROP TABLE users;--` → stored literally, logs in fine afterwards, `users` table intact. Zero `panic:`/`fatal error` in backend and webui logs across the entire run.
- **Error / unavailable path**: `docker compose stop backend` mid-flow → modal renders red **"Authentication service unavailable"**, no fake success, no silent close; password confirmed unchanged after the backend returned.
- **Auth & scope**: unauthenticated → 401. Forged cookie → 401. Expired session → 401 on HTTP and `Unauthenticated` on gRPC, password unchanged. Cross-user: `bob` rotating his own password left admin's password and admin's sessions untouched. Session invalidation: the three pre-existing admin sessions all → 401 after rotation while the acting browser session stayed alive (`/api/v1/auth/me` 200); bob's session untouched.
- **Adjacent flows**: `/dashboard`, `/profile`, `/api/v1/auth/me`, `GET`+`PUT /api/v1/profile/timezone`, `/api/v1/alerts` all 200 after the change; logout 200 then `me` → 401. Client-side guards still fire: empty fields → "All fields are required", mismatched confirm → "New password and confirmation do not match".
- **Persisted state**: rotation exercised on a database already holding users and sessions created before the change, and re-verified after a full `backend`+`webui` restart on the existing volume.
- **Generated templates**: committed `Profile_templ.go` matches the `.templ` source at this SHA (the image rebuilds templates itself, and the diff is consistent).

### Non-blocking observations

1. **`session_id` is read from the request body, not from the authenticated caller** (`services.go:334`). With `bob`'s session in gRPC metadata and **admin's** session id in the body plus admin's correct current password, the RPC rotated **admin's** password and returned `success: true`. Not a privilege escalation — you already need the victim's live session id *and* their current password — but the handler never cross-checks `backend.AuthenticatedUser(ctx)`, which the interceptor populates for exactly this. The WebUI handler is safe (it uses `middleware.GetSessionID(c)`); this is reachable only with direct gRPC access. Unchanged from the previous head.
2. **Button visibility for real OAuth logins is code-verified, not exercised live** — this env runs `OAUTH_ENABLED=false` with placeholder GitHub credentials, so no genuine OAuth callback could be driven. Flipping a user's `o_auth_provider`/`o_auth_id` in postgres does *not* hide the button, because `ProfilePage` keys on the session's `auth_method`, which classic login never sets to `oauth:`. That path is unchanged by this PR and the criterion notes it was "already the case"; the part this PR adds (server-side refusal) is verified live.

**Verdict: FAIL** — 1 of 4 criteria failed. Fixer summoned.

SoulKyu:
🔀 Rebase (conflicts resolved): rebased onto `origin/main` (567dccf).

**Conflicts:** `internal/webui/templates/pages/Profile.templ` (+ its generated `Profile_templ.go`), across all 3 branch commits.

**Resolution:**
- `profilePage()` Alpine state: kept `main`'s `idCopied` / `userId: ""` + `init()` reading `this.$el.dataset.userId` (main moved the ID to a `data-user-id` attribute), dropped this branch's now-obsolete `showToast` and `userId: '{ data.User.ID }'` inlining. Added the change-password state fields (`showChangePasswordModal`, `currentPassword`, `newPassword`, `confirmPassword`, `changePasswordError/Success/Loading`, `changePasswordAttempt`) alongside them.
- Dropped `main`'s `showChangePassword()` stub (`alert("coming soon")`) — dead code, the button now calls `openChangePasswordModal()`.
- Kept this branch's `resetChangePassword()` / `openChangePasswordModal()` / `closeChangePasswordModal()` and the single-flight, attempt-scoped `changePassword()`.
- `Profile_templ.go` never hand-edited: reset to the `main` version and regenerated via `make webui-templates` after each conflicting commit.

**Verification:** `go build ./...` ✅, `go vet ./...` ✅, `go test ./internal/backend/services/` ✅ (50 pass), `node internal/webui/static/js/profile_change_password.test.js` ✅ (0 fail). Branch diff vs `main` contains no unrelated `*_templ.go` churn.

SoulKyu:
## 🧪 QA Report

Env: full `docker compose` stack rebuilt from this SHA (backend + webui + fake-alertmanager images built from source, fresh postgres volume). Browser flows driven with real Chromium (Playwright), HTTP probes with `curl`, and the gRPC surface hit directly on `:50051` with a Go client built against the repo's generated stubs. Re-test focused on the two modal fixes (`afbb7e0`, `c103f49`) on top of a full re-run of the acceptance criteria.

|---|---|---|
| A classic-auth user can change their password from `/profile` with the current password; the new one works on next login, the old one doesn't | ✅ | Browser: login → `/profile` → "Change Password" → filled current/new/confirm → green "Password changed successfully." `POST /api/v1/auth/login` with the old password → **401 Invalid credentials**, with the new one → **200**. The submitting session stayed alive (`/api/v1/auth/me` → 200) while the user's three other sessions were killed (all → 401, `sessions` rows 4 → 1 for that user); another user's session was untouched. |
| Wrong current password is rejected with a visible error (no silent success), and the RPC is unreachable without a valid session | ✅ | Modal showed the red **"Current password is incorrect"** banner, no success banner, password unchanged — and the *same* modal then accepted the corrected password (no permanent lock-out from the new single-flight guard). Unauthenticated `POST /api/v1/auth/change-password` → 401 `Authentication required`; garbage cookie → 401; an invalidated session cookie → 401 `Invalid or expired session`. gRPC with no metadata → `Unauthenticated: … requires a session or service token`; garbage metadata → `Unauthenticated: invalid or expired session`; empty `session_id` in the body → `Session ID is required`. |
| OAuth-only accounts never see the button and the RPC refuses them server-side | ✅ | Seeded an OAuth-only user (`o_auth_provider=github`, empty `password_hash`) with a live session: gRPC `ChangePassword` → `success=false "OAuth accounts do not have a password to change"`, both with an empty and a guessed `old_password`. Button is gated on `data.User.OAuthProvider == nil` (`Profile.templ:175`), which `ProfilePage` sets from the session's `auth_method` prefix `oauth:` (`profile_handlers.go:25-36`) — only the OAuth callback writes that value, so an OAuth session renders no button. |
| The `admin:admin` bootstrap credential can be rotated end-to-end through the UI | ✅ | `admin:admin` rotated through the browser modal (submitted with the Enter key), then rotated four more times across the adversarial passes. Old credential → 401, current one → 200, and it survived `docker compose restart backend webui` on the existing volume with no re-seeding of `admin:admin` (still 401 after restart). |

**Adversarial pass**

- **The regression this SHA fixes (double-submit under latency).** Injected a 2 s delay on `/api/v1/auth/change-password` via request interception, then fired **Enter three times** ~50 ms apart on the confirm field: exactly **1** request left the page, the modal showed "Changing…" with Cancel disabled, and the end state was the success banner with the error banner empty and hidden. Repeated with **three rapid clicks on the submit button** (1.5 s delay, force-clicks to bypass the disabled attribute): again **1** request, same clean end state. Server side confirmed only one rotation happened each time (previous password → 401, new one → 200). Cancel/X/backdrop were all verifiably non-dismissible mid-flight (`cancelDisabled: true`, click rejected).
- **Stale state after close/reopen.** The attempt-scoping guard is exercised by the committed `node --test` suite (4/4 pass, including "a response landing after close + reopen cannot write to the modal"). I could not reach that path through the UI at all any more — every close widget is inert while a request is in flight — so it is now defence in depth rather than a live path. Reopening after success/cancel showed all three fields empty and no stale success/error banner.
- **Boundary & empty**: 3-char new password → client-side *and* server-side "must be at least 4 characters long"; empty strings / missing `new_password` → 400; mismatched confirmation → "New password and confirmation do not match" with no request issued; new password == old password → allowed (rotation succeeds, other sessions still invalidated). Exactly-72-byte password → accepted and logs in. **73 bytes and a 1 MiB password → `Internal server error` (400)** — bcrypt's 72-byte cap surfaced opaquely; *not* a fail: the stored password is left untouched (verified: the attempted value 401s, the previous one still 200s) and `Register` behaves identically, so no unreachable state is created.
- **Malformed & hostile**: truncated JSON, wrong types (`"old_password":123`, object as new password), and a form-encoded body all → 400 with a clean error, no 5xx and no panic in the backend log. The form-encoded rejection also means a cross-site form POST cannot drive the endpoint (cookie is `SameSite=Lax`). The error banner renders via `x-text` (escaped) and every string the server returns is a static literal.
- **Error & unavailable path**: `docker compose stop backend` with the modal filled, then submit → red **"Authentication service unavailable"**, no success banner, loading cleared and the controls re-enabled; after restarting the backend, the attempted password 401s and the previous one still 200s — nothing was silently written. Separately, a response that never arrives leaves the modal pinned on "Changing…" with every close widget disabled (a page reload is the only way out) — reachable only from a hop that accepts and never answers, since the webui's own gRPC deadline is 10 s; noted below, not counted against a criterion.
- **Auth & scope**: covered in the table. Additionally, two live sessions of the same user: browser A rotates, browser B (session killed by that rotation) submits its pre-filled modal → visible **"Invalid or expired session"**, no success banner, and B's password never takes effect.
- **Adjacent flows**: after a change — `GET`/`PUT /api/v1/profile/timezone` still 200, dashboard renders, logout → 200 and the session is dead afterwards, re-login with the rotated password works. No new console errors attributable to the modal.
- **Persisted state / upgrade on an existing DB**: built the backend image from `main` (verified genuinely pre-PR: `ChangePassword` → `Unimplemented: unknown method`), ran the stack on it, registered a `legacy` user and logged in — then swapped in this SHA's backend on the **same** volume. That pre-existing row rotated cleanly through the browser modal (old password → 401, new → 200), and the session created before the upgrade was invalidated as expected.
- **Config branch added by this PR**: with `OAUTH_ENABLED=true` + `OAUTH_DISABLE_CLASSIC_AUTH=true`, the RPC refuses with "Username/password authentication is disabled. Please use OAuth authentication." and does not write — matching `Register`'s behaviour.
- **Generated template**: `templ generate` on `Profile.templ` reproduces the committed `Profile_templ.go` byte-for-byte — no stale generated file.

Two observations that did **not** break a criterion (both carried over from the earlier review of this PR, unchanged by these two commits):

1. `ChangePassword` resolves the user from `req.SessionId` in the request **body** and ignores the caller the interceptor already authenticated. Re-verified live: a client authenticated as `bob` (bob's session in the metadata) but carrying **admin's** session id in the body successfully rotated admin's password. Exploiting it requires already holding the victim's session token — itself a full compromise — and every other handler in this service follows the same `req.SessionId` pattern, so it's a hardening gap rather than a new hole. `caller.ID == user.ID` via `backend.AuthenticatedUser(ctx)` would close it.
2. An account holding **both** a password hash and an OAuth link sees the button (the webui derives `OAuthProvider` from the session's `auth_method`, which is `classic` for a password login) while the server always refuses it — the dead-end control #178 is about. I had to forge that row with SQL: `CreateOrUpdateOAuthUser` never links an OAuth identity onto an existing classic account, so it stays a latent trap only if account linking is ever added.

**Verdict: PASS**

SoulKyu:

SoulKyu:
## 🧪 QA Report

Re-test after the rebase onto `main` (`663aab1`). The PR's own files are byte-identical to the
previously-tested tree (`git diff c103f49 fb6bf6e` touches only files coming from `main`:
mention-autocomplete, flap guard, alert-modal), so this run re-validated every criterion against
the **new base** — full `docker compose` rebuild from this SHA (backend + webui images built from
source, fresh Postgres volume), browser flows driven with real Chromium (Playwright), HTTP probes
with `curl`, gRPC hit directly on `:50051` with a Go client built against the repo's stubs.
Verified the running containers actually carry this SHA (webui binary contains the new
`mentionable` route, backend contains `ChangePassword`).

|---|---|---|
| A classic-auth user can change their password from `/profile` with the current password; the new one works on next login, the old one doesn't | ✅ | Browser: login → `/profile` → "Change Password" → current/new/confirm → green "Password changed successfully."; `POST /api/v1/auth/login` old password → **401**, new → **200**. Session scoping re-checked: the submitting session stayed alive (`/auth/me` 200) while the user's 3 other sessions all went **401**, and another user's (`bob`) session was untouched. |
| Wrong current password is rejected with a visible error (no silent success), and the RPC is unreachable without a valid session | ✅ | Red **"Current password is incorrect"** banner, no success banner, password unchanged — and the same modal then accepted the corrected password (no lock-out from the single-flight guard). Unauthenticated POST → 401 `Authentication required`; garbage cookie → 401; a session invalidated by an earlier rotation → 401 `Invalid or expired session`. gRPC: no metadata → `Unauthenticated: … requires a session or service token`; garbage metadata → `Unauthenticated: invalid or expired session`; empty body `session_id` → `Session ID is required`. |
| OAuth-only accounts never see the button and the RPC refuses them server-side | ✅ | This run rendered the page for a real OAuth session instead of only reading the code: pinned `NOTIFICATOR_SESSION_SECRET`, minted a webui cookie for an OAuth-only user (`o_auth_provider=github`, empty hash) with `auth_method=oauth:github` → `/profile` renders **no** "Change Password" button (classic control page renders it; 35 649 B vs 35 934 B). That same session's `POST /api/v1/auth/change-password` → 400 `OAuth accounts do not have a password to change`, and gRPC refuses with empty and guessed `old_password`. |
| The `admin:admin` bootstrap credential can be rotated end-to-end through the UI | ✅ | `admin:admin` rotated through the browser modal (Enter key) on a **fresh** volume and again on the **upgraded** volume that already held other users: `admin` → 401, `NewPass1` → 200. Rotation survived `docker compose restart backend webui` (old password still 401 afterwards). |

**Adversarial pass** — what I tried to break it with, on this base:

- **Double-submit under latency (the regression the last two commits fix).** 2 s delay injected on `/api/v1/auth/change-password`, then **Enter ×3** ~50 ms apart → exactly **1** request left the page, modal pinned on "Changing…", Cancel/X/backdrop verifiably inert (`cancelDisabled: true`, click rejected), end state = success banner with the error banner empty and hidden. Repeated with **3 force-clicks** on the submit button at 1.5 s delay → again **1** request. Server side: only one rotation happened each time (previous password 401, new one 200).
- **Boundary & empty**: empty strings → 400; missing field → 400; 3-char new password → client-side *and* server-side "must be at least 4 characters long"; mismatched confirmation → banner, no request issued; new == old → allowed, other sessions still invalidated; **exactly 72 bytes** → accepted and logs in; **73 bytes** → `Internal server error` (400) — bcrypt's cap surfaced opaquely, but nothing is written (attempted value 401s, previous one still 200s) and `Register` behaves identically, so no unreachable state. New this run: **multi-byte passwords** — `pä🔐sswörd` rotates and logs in; `é🔐` (2 runes / 6 bytes) is accepted server-side (min length is counted in bytes) while the client counts UTF-16 units — both ends still reject anything genuinely short, no wrong outcome.
- **Malformed & hostile**: truncated JSON, `"old_password":123`, object as new password, and a form-encoded body → all 400 with a clean error, no 5xx, `grep -c panic` on both container logs = **0**. Form-encoded rejection also means a cross-site form POST can't drive the endpoint (cookie is `SameSite=Lax`). Set the password to `` — stored and logs in fine; the banner renders through `x-text` and every server string is a static literal.
- **Error & unavailable path**: `docker compose stop backend` with the modal filled, then submit → red **"Authentication service unavailable"**, no success banner, loading cleared, controls re-enabled; after restart the attempted password 401s and the previous one still 200s — nothing silently written. A response that never arrives leaves the modal on "Changing…" with every close widget disabled (page reload is the only exit); only reachable from a hop that accepts and never answers (the webui's own gRPC deadline is 10 s), so not counted against a criterion.
- **Auth & scope**: table above, plus two live sessions of the same user — browser A rotates, browser B (session killed by that rotation) submits its pre-filled modal → visible **"Invalid or expired session"**, no success banner, B's password never takes effect.
- **Adjacent flows on the new base**: `main`'s `@`-mention endpoint `GET /api/v1/users/mentionable` returns 200 with results both **before and after** a rotation on the surviving session; `GET`/`PUT /api/v1/profile/timezone` → 200/200; dashboard renders; logout → 200 and the session is dead; re-login with the rotated password works. No console errors attributable to the modal.
- **Persisted state / upgrade on an EXISTING database**: booted the stack on a backend image built from `main` (verified genuinely pre-PR: `ChangePassword` → `Unimplemented: unknown method`), registered `legacy` and logged in, then swapped in this SHA's backend on the **same volume**. That pre-existing row rotated cleanly through the browser modal (`oldpass1` → 401, `Legacy99` → 200) and the session created before the upgrade was invalidated. Rotation also survived a full `restart`.
- **Config branch**: `OAUTH_ENABLED=true` + `OAUTH_DISABLE_CLASSIC_AUTH=true` → RPC refuses with "Username/password authentication is disabled…" and does **not** write (after re-enabling, the pre-branch password still authenticates).
- **Generated template / committed tests**: `Profile_templ.go` is unchanged by the rebase and still matches the file previously verified byte-for-byte against pinned `templ` v0.3.906 (a regen with the newer local v0.3.1020 differs only in the version header). `node --test internal/webui/static/js/profile_change_password.test.js` → 4/4 pass.

Two observations that did **not** break a criterion (unchanged by the rebase, same as the previous review):

1. `ChangePassword` resolves the user from `req.SessionId` in the request **body** and ignores the caller the interceptor already authenticated. Re-verified live on this base: a client authenticated as `bob` but carrying **admin's** session id in the body rotated admin's password. Exploiting it requires already holding the victim's session token — itself a full compromise — and every other handler in this service follows the same `req.SessionId` pattern, so it's a hardening gap, not a new hole (`caller.ID == user.ID` via `backend.AuthenticatedUser(ctx)` would close it).
2. An account holding **both** a password hash and an OAuth link still sees the button after a classic login (the context user is the login-time snapshot) while the server always refuses it. I had to forge that row with SQL — `CreateOrUpdateOAuthUser` never links an OAuth identity onto an existing classic account — so it stays latent unless account linking is added.

**Verdict: PASS**

SoulKyu (commented):
@SoulKyu Reviewed the full change-password feature at head fb6bf6eb (proto `ChangePassword` RPC, `AuthServiceGorm.ChangePassword` in `internal/backend/services/services.go`, the WebUI route/handler/client wiring, and the Alpine modal in `Profile.templ`), including the two prior rounds of fixes on the async modal-state race.

What I verified independently:
- Backend: old-password bcrypt check, OAuth/passwordless rejection via `IsOAuthUser()`/`HasPassword()`, new-password minimum length, hash rotation via `UpdateUserPassword`, and `InvalidateUserSessions(userID, exceptSessionID)` correctly excluding the current session while hard-deleting the others. `ChangePassword` is absent from `publicMethods` in `auth_interceptor.go`, so it's covered by the deny-by-default gate as claimed.
- Ran `go build ./...` and `go test ./internal/backend/services/... -run TestChangePassword -v` — both pass, including the "current session survives, others invalidated" and OAuth-rejection assertions.
- Frontend: the modal's state machine now uses an attempt counter (`changePasswordAttempt`) that both `openChangePasswordModal()`/`closeChangePasswordModal()` (via `resetChangePassword()`) and `changePassword()` itself bump/check, so a response landing after a cancel-and-reopen can no longer write stale success/error state into a modal it no longer owns, and a second submit while one is in flight is a no-op. This resolves the exact stale-state race flagged in the two earlier review rounds (fa65f11 → 0afebf3 → this head).
- Ran the new `profile_change_password.test.js` under `node --test` — all 4 cases pass, including the "response landing after close + reopen cannot write to the modal" scenario, which reproduces the originally reported bug.

No blocking or non-blocking issues found in this pass. Nice work tightening the async edge cases across the review rounds — the final attempt-counter approach is a clean, general fix rather than a one-off patch.



SoulKyu:

PR #187, opened 2026-08-03
fix(webui): graceful shutdown on SIGTERM/SIGINT

### LINKED ISSUE #184: fix(webui): no graceful shutdown — every restart severs in-flight requests mid-response
## Problem

The webui process has no shutdown handling at all. `runWebUI` ends with a bare `router.Run(listenAddr)` (cmd/webui.go:52). On SIGTERM/SIGINT the Go runtime kills the process immediately:

- In-flight HTTP requests are severed mid-response (connection reset), including dashboard data loads and bulk actions that are already half-executed against the backend.
- SSE clients (`/api/v1/dashboard/stream`) are dropped with a raw TCP reset instead of a clean stream end.
- The alert cache poller (`alertCache.Start()`, internal/webui/router.go:99) can be killed mid-refresh-cycle, in the middle of its backend writes (`CreateResolvedAlert`, statistics capture) — a refresh cycle's worth of resolved-alert bookkeeping is silently lost on every restart.

The backend half of the same deployment does this correctly (`setupGracefulShutdown`, internal/backend/server.go:223-258: signal handler → `GracefulStop` → HTTP `Shutdown` with timeout → DB close). The webui — the half that actually terminates user-facing connections — got nothing.

## Impact

Every rolling deploy or restart (docker-compose `make test` flow, Helm upgrades, k8s pod churn) produces user-visible failed requests: browsers see connection resets, on-call engineers mid-triage see actions fail with no indication whether the silence/ack they just clicked actually landed (the gRPC write may or may not have completed). Under Kubernetes this is the classic "502s during every deploy" pattern; SIGTERM arrives while the pod is still receiving traffic, and there is no drain window at all.

## Proposed approach

Standard ~15-line pattern in `runWebUI` (cmd/webui.go), mirroring what the backend already does:

- Wrap the gin engine in an `http.Server` and start it in a goroutine.
- `signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)`, then on signal call `srv.Shutdown(ctx)` with a 5-10s timeout — this stops accepting new connections and lets in-flight handlers finish. Long-lived SSE handlers already terminate via `c.Request.Context().Done()` when Shutdown closes their connections at the deadline.
- Stop the alert cache refresh loop (it already has a `Stop()`/ticker teardown) before exiting so a refresh cycle isn't killed mid-backend-write.

No new dependencies, no config.

## Acceptance criteria

- [ ] Sending SIGTERM to a webui serving a slow in-flight request lets that request complete (up to the drain timeout) instead of resetting the connection.
- [ ] New connections after SIGTERM are refused while draining; process exits 0 after drain.
- [ ] The alert cache poller is stopped as part of shutdown (no refresh cycle started after the signal).
- [ ] Behaviour verified in the docker-compose flow (`docker-compose restart webui` no longer produces connection-reset errors in an open dashboard's network tab beyond the SSE reconnect).

SoulKyu:
## Summary
- `runWebUI` (`cmd/webui.go`) previously ended with a bare `router.Run(listenAddr)`, so SIGTERM/SIGINT killed the process immediately, severing in-flight HTTP requests (including SSE dashboard streams) and potentially killing the alert cache poller mid backend-write.
- Wraps the gin engine in an `http.Server` started via `ListenAndServe` in a goroutine.
- Listens for `os.Interrupt`/`syscall.SIGTERM` via `signal.NotifyContext`; on signal, stops the alert cache refresh loop (`alertCache.Stop()`) first, then calls `srv.Shutdown(ctx)` with a 10s drain timeout — mirroring the existing `setupGracefulShutdown` pattern already used by the backend (`internal/backend/server.go:223-258`).
- `SetupRouter` now returns the `*services.AlertCache` alongside the `*gin.Engine` so `runWebUI` can stop it during shutdown.

## Test plan
- [x] `go build ./...` and `go build -tags webui ./...`
- [x] `go vet -tags webui ./...`
- [x] `go test -tags webui ./cmd/... ./internal/webui/...` (210 passed)
- [ ] Manual: `docker-compose restart webui` with an open dashboard no longer shows connection-reset errors beyond the expected SSE reconnect

Closes #184

### DIFF
--- a/cmd/webui.go
+++ b/cmd/webui.go
@@ -4,8 +4,14 @@
 package cmd
 
 import (
+	"context"
 	"fmt"
 	"log"
+	"net/http"
+	"os"
+	"os/signal"
+	"syscall"
+	"time"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
@@ -51,11 +57,34 @@ func runWebUI(cmd *cobra.Command, args []string) {
 	fmt.Printf("   Listen: %s\n", listenAddr)
 	fmt.Printf("   Backend: %s\n", backendAddr)
 
-	router := webui.SetupRouter(backendAddr)
+	router, alertCache := webui.SetupRouter(backendAddr)
 
 	fmt.Printf("Visit http://localhost%s to view the WebUI\n", listenAddr)
 
-	if err := router.Run(listenAddr); err != nil {
-		log.Fatal("Failed to start WebUI server:", err)
+	srv := &http.Server{
+		Addr:    listenAddr,
+		Handler: router,
 	}
+
+	go func() {
+		if err := srv.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+			log.Fatal("Failed to start WebUI server:", err)
+		}
+	}()
+
+	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
+	defer stop()
+	<-ctx.Done()
+
+	fmt.Println("🛑 Shutting down WebUI server...")
+
+	alertCache.Stop()
+
+	shutdownCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+	if err := srv.Shutdown(shutdownCtx); err != nil {
+		log.Printf("WebUI server shutdown error: %v", err)
+	}
+
+	fmt.Println("✅ WebUI server shut down gracefully")
 }
--- a/internal/webui/router.go
+++ b/internal/webui/router.go
@@ -19,7 +19,7 @@ import (
 	"github.com/gin-gonic/gin"
 )
 
-func SetupRouter(backendAddress string) *gin.Engine {
+func SetupRouter(backendAddress string) (*gin.Engine, *services.AlertCache) {
 	r := gin.New()
 
 	// Load configuration with Viper to support environment variables
@@ -384,7 +384,7 @@ func SetupRouter(backendAddress string) *gin.Engine {
 		protectedPages.GET("/activity", handlers.ActivityPage)
 	}
 
-	return r
+	return r, alertCache
 }
 
 // generateRandomSecret generates a cryptographically secure random secret


SoulKyu (commented):
@SoulKyu Reviewed the graceful-shutdown fix for the webui server (closes #184).

Verified against the diff in `cmd/webui.go` and `internal/webui/router.go`:
- `runWebUI` now runs `srv.ListenAndServe()` in a goroutine and blocks on `signal.NotifyContext(os.Interrupt, syscall.SIGTERM)` instead of the old blocking `router.Run`, so SIGTERM/SIGINT now trigger an orderly `alertCache.Stop()` + `srv.Shutdown(ctx)` with a 10s drain instead of an immediate kill.
- `SetupRouter`'s new `(*gin.Engine, *services.AlertCache)` signature has a single call site (`cmd/webui.go:60`) and `NewAlertCache` never returns nil on the success path (the only early exit, a failed backend connect, calls `log.Fatalf` before the cache is constructed), so there's no nil-pointer risk on `alertCache.Stop()`.
- Traced several edge cases: SIGTERM arriving before `ListenAndServe` starts listening (safe — `Shutdown` handles a not-yet-serving server), a bind failure inside the goroutine (still `log.Fatal`s the process, same as the old synchronous behavior), and double-signal delivery (standard `NotifyContext` semantics, unchanged from other packages in this repo).
- `go build -tags webui ./...`, `go vet -tags webui ./cmd/... ./internal/webui/...`, and the existing `cmd`/`webui` test suites all pass locally, and there are no test files referencing `SetupRouter` that the signature change could break.

No actionable findings — this is a clean, well-scoped fix that mirrors the existing backend shutdown pattern. Nice work closing out #184.



SoulKyu:
## 🧪 QA Report

Env: docker-compose stack rebuilt from this SHA (webui/backend/alertmanager images), webui on `:10081`. Browser flows driven with Playwright Chromium.

|---|---|---|
| SIGTERM during a slow in-flight request lets it complete (up to drain timeout) instead of resetting | ✅ | Paused `alertmanager`, issued `GET /health/alertmanager` (blocks), sent SIGTERM at T+1s, unpaused at T+3s → request returned **200 with full body at 3.11s** (server log: `"GET /health/alertmanager 200 3.112234183s"`). Repeated with a real backend write: paused `backend`, `POST /api/v1/dashboard/bulk-action {action:acknowledge}`, SIGTERM at T+1s → **200 `processedCount:1`**, and the ack is persisted (`/alert/<fp>/history` → `acknowledged_at 05:19:04Z`, i.e. written *after* the signal). |
| New connections refused while draining; process exits 0 after drain | ✅ | During the drain window `curl /health` fails (exit 56, no service) while the old process still holds the port. Exit code **0** in every run: SIGTERM, SIGINT, `docker compose stop`, and `docker compose restart` — 10 runs total, including one with **60 concurrent SSE clients** (drain 10.0s, `exit=0`, never SIGKILL/137). |
| Alert cache poller stopped as part of shutdown (no refresh cycle started after the signal) | ✅ | Sync interval is 10s, so a tick necessarily falls inside a 10s drain. With 60 SSE clients connected, the container's stderr between `🛑 Shutting down` (05:17:27.969) and `✅ shut down gracefully` (05:17:37.969) contains exactly **one** line — `WebUI server shutdown error: context deadline exceeded` — and **zero** `Alert cache refresh: fetched …` lines. Before the change that window would have contained a full refresh cycle. |
| `docker-compose restart webui` no longer produces connection-reset errors in an open dashboard beyond the SSE reconnect | ✅ | Playwright Chromium logged into the dashboard (11 alert rows rendered), then `docker compose restart webui`. Only 3 failed requests: the SSE stream (`ERR_INCOMPLETE_CHUNKED_ENCODING` — the explicitly allowed SSE case) and 2 × `/health/backend` polls that Chrome **issued 0.0s before they failed**, i.e. fired at the instant of process exit (the unavoidable container-restart gap), not in-flight responses severed mid-body. No aborted dashboard data load, no half-executed bulk action. Post-restart the dashboard keeps rendering and updating (screenshot, "Last updated 7:37:21 AM", 1/1 live source). |

**Adversarial pass** (what I tried to break each criterion with):

- **In-flight completion (C1):** ① in-flight *read* (`/health/alertmanager`) → completed; ② in-flight *gRPC write* (bulk acknowledge, backend paused so it was genuinely mid-write when SIGTERM landed) → completed **and** persisted across the restart, verified from the DB via the history endpoint; ③ **overrun boundary** — request still pending when the 10s drain deadline expires: with `backend` left paused, the handler returned its own `503` at **10.002s** (its gRPC deadline) rather than the client seeing a reset; ④ listener-failure path — `NOTIFICATOR_WEBUI_LISTEN=:10081` on an already-bound port: the new `go ListenAndServe` goroutine still fails fast (`Failed to start WebUI server:listen tcp :10081: bind: address already in use`, **exit 1**, immediate) instead of hanging forever waiting for a signal. That was the main regression risk of moving the listener into a goroutine, and it is handled.
- **Signals / repeat signals (C2):** SIGINT (Ctrl-C) path works identically to SIGTERM. Sending a **second SIGINT 2s into the drain** does **not** abort early — `signal.NotifyContext` keeps the handler installed, so the process still drains the full 10s (operator cannot interrupt a slow drain; cosmetic, not a criterion).
- **Docker grace-period race (C2/C4):** the 10s shutdown timeout is exactly docker's default 10s stop grace period, so I tried to force a SIGKILL: 4 sequential `docker compose stop` runs with an SSE client (10.50–10.79s wall each) and one with **60 SSE clients** (10.75s). All exited **0**; never reproduced a 137. Flagging it as a thin margin rather than a failure — `srv.Shutdown` always burns the *full* 10s whenever any dashboard is open, because the SSE handler only returns on `c.Request.Context().Done()` and `Shutdown` never cancels it (`WebUI server shutdown error: context deadline exceeded` is logged every such run). A 5s timeout, or `RegisterOnShutdown` closing SSE subscribers, would leave real headroom under the default grace period.
- **Poller (C3):** ① verified no cycle *starts* after the signal (above). ② Tried the inverse — kill a cycle that is *already running*: with `backend` paused, a refresh that began 2s before SIGTERM (blocked in `loadBackendData`/`Loading acknowledgments…`) was still cut off when the process exited **0.37s** after the signal. `AlertCache.Stop()` cancels the ctx and the ticker but does not wait for an in-flight `refreshAlerts()` (no WaitGroup), so the issue's stated goal "a refresh cycle isn't killed mid-backend-write" holds only for cycles not yet started. The acceptance criterion as written ("no refresh cycle started after the signal") is met, so this stays ✅ — but it is the remaining half of the original problem and worth a follow-up.
- **Adjacent flows / persisted state (C4):** login → dashboard → restart → dashboard still renders and keeps updating; the acknowledgment written during the previous drain survives in Postgres; the alert list refreshes after restart. Session survival requires `NOTIFICATOR_SESSION_SECRET` (pre-existing, unrelated to this PR — the compose default leaves it empty and the webui warns "Sessions will NOT persist across restarts"); with it set, the open dashboard needs no re-login after a restart. Also noted (pre-existing client behaviour, **not** introduced here): `initSSE().onerror` does `destroySSE(); startAutoRefresh()`, so after a restart the dashboard never re-opens the SSE stream (server-side `subscribers` stayed 0 for 80s) and permanently falls back to polling — data keeps flowing, so no user-visible break.
- **Auth/scope:** the change adds no HTTP surface (no shutdown endpoint), so there is nothing new to reach unauthenticated. `SetupRouter`'s signature change has exactly one caller (`cmd/webui.go:60`); `alertCache` cannot be nil there because `SetupRouter` `log.Fatal`s earlier if the backend is unreachable, so `alertCache.Stop()` can't nil-deref.

**Verdict: PASS**

PR #189, opened 2026-08-03
docs(wiki): sync with recent changes

SoulKyu:
Automated openwiki refresh covering:
fix(backend): enforce auth at the gRPC layer (#172)
fix(auth): public registration cannot be disabled and the OAuth-only gate fails open (#169)
fix(webui): snapshot alert before spawning CaptureAlertFired goroutine (#168)
fix(test): set the required service token in CaptureAlertFired snapshot test
feat(dashboard): Alertmanager source freshness — live/stale/unreachable (#170)
feat(notifications): scope browser/sound notifications to a saved filter preset (#158)
feat(dashboard): snooze an alert for a chosen duration instead of forever (#116)
feat(dashboard): blast-radius panel in the alert modal (#174)
feat(collaboration): @mentions in alert comments (#176)
fix(webui): profile page renders fabricated account/session data and its Copy-ID button copies a literal template string (#181)

### DIFF
--- a/openwiki/.last-update.json
+++ b/openwiki/.last-update.json
@@ -1,6 +1,6 @@
 {
-  "updatedAt": "2026-08-01T05:33:10.000Z",
+  "updatedAt": "2026-08-03T05:34:27.000Z",
   "command": "update",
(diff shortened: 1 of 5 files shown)


SoulKyu:

